PR #25, opened 2026-05-09
Implement POST /certificate_renewal endpoint

### LINKED ISSUE #15: feat: implement POST /certificate_renewal endpoint
## Description

The Puppet CA HTTP API includes a `POST /puppet-ca/v1/certificate_renewal` endpoint that allows agents to renew their certificates using mTLS authentication with their existing cert/key pair. This endpoint is not currently implemented.

## Expected behavior

- **Method:** POST
- **Path:** `/puppet-ca/v1/certificate_renewal?environment=production`
- **Authentication:** mTLS with the agent's existing certificate
- **Request body:** PEM-encoded CSR (`Content-Type: text/plain`)
- **Response:** PEM-encoded renewed certificate (200 OK)
- **Semantics:** The CA verifies the CSR CN matches the authenticated client CN, then issues a new certificate for the same subject. The existing private key is reused by the agent.

## Why this matters

This is one of two missing endpoints that prevent golang-puppet-ca from being a full drop-in replacement for the Puppet Server CA. Any tooling or automation that relies on the certificate renewal API will not work without this endpoint.

bootc:
Agents with an existing CA-signed certificate can now renew it over mTLS without waiting for manual approval: the CA verifies the CSR CN matches the authenticated client CN and issues a replacement certificate immediately. This is required for golang-puppet-ca to be a full drop-in replacement for the Puppet Server CA.

New CA method Renew() acquires the per-subject distributed lock, saves the renewal CSR, and calls signWithDuration() directly — bypassing the pending-CSR queue and autosign check. Auth tier is tierAnyClient so any valid node cert grants access; CN-matching in both the handler and Renew() prevents an agent from renewing another's certificate.

Fixes #15

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,6 +343,14 @@ All endpoints are served under both the bare path and `/puppet-ca/v1/<path>`, so
 |--------|------|-------------|
 | `GET` | `/certificate/{subject}` | Retrieve a signed certificate PEM (`ca` returns the CA cert) |
 
+### Certificate renewal
+
+| Method | Path | Description |
+|--------|------|-------------|
+| `POST` | `/certificate_renewal` | Renew an existing certificate; body: raw CSR PEM; returns new certificate PEM |
+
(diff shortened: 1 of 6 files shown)


trevor-vaughan:
@bootc Blocked on rebase

PR #36, opened 2026-06-02
Audit and improve

trevor-vaughan:
## Security audit & hardening pass

Audits and hardens the CA server across information disclosure, DoS, crash durability, shutdown correctness, and CI coverage. No feature or API changes — existing behavior is preserved except where it was leaking, lossy, or silently skipped. Every fix lands with negative-path test coverage.

### Information disclosure (CWE-209)
- HTTP handlers returned internal error chains — including on-disk paths like `mkdir /var/lib/puppet-ca/requests` — to clients verbatim via `http.Error(w, err.Error(), …)`. All such sites now log server-side and return a generic client message.
- The unauthenticated `PUT /certificate_request` endpoint is the sharpest case: it mixes path-free, client-actionable CSR-validation errors (malformed/mis-signed CSR, CN mismatch) with internal storage/autosign faults that embed paths. A fail-safe `csrValidationError()` allowlist surfaces the validation messages as `400` (agents and tests rely on them) and downgrades everything unrecognized to a logged generic `500`. The same read-body leak in the OCSP `POST` handler and the `POST /certificate_renewal` endpoint is closed too.
- Normalized the lone `invalid subject` handler that still echoed format detail, matching the other seven endpoints.

### Denial-of-service (NIST SC-5)
- JSON `POST`/`PUT` bodies are now capped at 1 MiB via `http.MaxBytesReader`, matching the existing CSR/OCSP limits; oversize bodies translate to `413`. Prevents an authenticated client from streaming an unbounded `certnames` array to exhaust memory.

### Crash durability
- `atomicWriteFile` now `fsync`s the temp file before rename and `fsync`s the parent directory after, so a crash can't leave a zero-length or stale CA key/cert/CRL where the rename was reordered ahead of the data. `AppendLine` `fsync`s appended inventory/CRL lines.
- `AppendInventory` previously downgraded a failed HMAC update to a `Warn`, leaving the inventory and its stored HMAC inconsistent so the *next* verify falsely reported tampering. It now propagates the error so the caller can react.

### Shutdown & cleanup correctness
- Converted ~10 `os.Exit(1)` calls inside the cobra `RunE` closure to returned errors. `os.Exit` skips deferred functions, so storage/signer/log cleanup was silently skipped on error paths — contradicting the documented "deferred cleanup always runs" guarantee. Added `cobra.NoArgs` + `SilenceErrors` and switched to `signal.NotifyContext` so in-flight context is cancelled on `SIGTERM`/`SIGINT`.
- The launcher hard-killed both children 10s after forwarding `SIGTERM`, truncating the frontend's own 30s `http.Server.Shutdown` budget — so the documented 30s drain (and Kubernetes' default pod grace period) never held in the default isolated-process deployment. The graceful budget is now a named 30s constant matching the frontend; the 5s crash-path teardown is unchanged. README signal-handling docs corrected to describe the two-layer mechanism.

### Autosign hardening
- `ValidateSubject` now rejects a leading `-` to prevent argv flag-injection into operator autosign scripts.
- An autosign script exiting `≥126` (not-executable / not-found / killed-by-signal) is logged as a likely misconfiguration so operators can tell a crash apart from a deliberate deny; fail-closed deny semantics are unchanged.

### CI & lint gate
- `mage dev:check` invoked `golangci-lint`, but the CI `check` job never installed it — a `LookPath` guard silently skipped the entire lint gate. Added a pinned install step so the gate actually runs.
- `test:unit`'s explicit package list omitted `./internal/signer/...` — the package holding the CA private key behind the socketpair — so its 12 PSK/handshake/isolation tests never ran in CI. Added.
- New v2 `.golangci.yml` (govet/errcheck/staticcheck/gosec/bodyclose/errname/nilerr/nolintlint and friends) passing clean, a `Dev.Lint` mage target, and `lefthook.yml` wiring the pre-commit/pre-push shims. Audit-confirmed gosec false positives (SHA1 mandated by RFC 5280/6960, self-reexec with operator argv, allowlist-gated autosign) are triaged via documented `//nolint` annotations. Excluded the opinionated `QF1008` staticcheck quick-fix so the gate doesn't reject the maintainer's deliberate embedded-selector style.
- gofmt/ST1005 cleanup and modernization to `min()`/`any` to get the formatting gate green.

### Tests
- Tampered-ciphertext coverage for the at-rest GCM integrity property, end-to-end traversal-certname rejection (`400`), an mTLS tier-classifier default-deny table test, a forced-CSR-save-failure regression asserting no path reaches the client, and `newRootCmd()` extracted so argument validation is testable without `os.Exit`.


internal/ca/ca_test.go
```diff
@@ -803,9 +803,13 @@ var _ = Describe("ValidateSubject", func() {
 	DescribeTable("valid subjects",
 		func(s string) { Expect(ca.ValidateSubject(s)).To(Succeed()) },
 		Entry("simple hostname", "puppet"),
-		Entry("FQDN", "node.example.com"),
+		Entry("FQDN", "foo.example.com"),
+		Entry("with interior hyphens", "a-b"),
 		Entry("with hyphens", "my-node-01"),
 		Entry("with underscores", "my_node"),
+		Entry("starts with digit", "node1"),
```
bootc:
```suggestion
		Entry("ends with digit", "node1"),
```
Either that, or:

```suggestion
		Entry("starts with digit", "1node"),
```

internal/ca/ca_test.go
```diff
@@ -803,9 +803,13 @@ var _ = Describe("ValidateSubject", func() {
...
+		Entry("starts with underscore", "_node"),
+		Entry("starts with dot", ".node"),
```
bootc, line 813: Should subjects with a leading dot be valid? I'm thinking not? When the CSR is written to disk it would be a hidden file, too, which would be surprising.

internal/ca/ca_test.go
```diff
@@ -815,6 +819,11 @@ var _ = Describe("ValidateSubject", func() {
 		Entry("double-dot only", ".."),
 		Entry("uppercase letters", "BadNode"),
 		Entry("empty string", ""),
+		// SECURITY: a leading '-' could be misread as a flag by an operator's
+		// autosign script, enabling argv flag injection. ValidateSubject must
+		// reject it.
+		Entry("leading hyphen (argv flag injection)", "-foo"),
+		Entry("leading hyphen only", "-"),
```
bootc, line 828: I'm thinking there should also be tests for strings containing whitespace, or excessively long names (>253) too. Names with trailing hyphens ("node-") are also invalid DNS labels and should probably be rejected?

I wonder whether there's some library function we can use to check for valid DNS names rather than rolling our own, as we'd always expect the CSR subject to be a valid DNS name.

internal/ca/signing.go
```diff
@@ -115,7 +115,10 @@ func (c *CA) evictRevokedLocked(ctx context.Context, subject string) error {
 	return nil
 }
 
-var subjectRegex = regexp.MustCompile(`^[a-z0-9._-]+$`)
+// subjectRegex forbids a leading '-' so a certname can never be misread as a
+// flag by an operator's autosign script (argv flag injection); the first
+// character must be a letter, digit, underscore, or dot.
+var subjectRegex = regexp.MustCompile(`^[a-z0-9_.][a-z0-9._-]*$`)
```
bootc, line 136: Following on from my earlier comments:
```suggestion
// subjectRegex matches the intersection of a valid lowercase DNS hostname
// (RFC 1123: dot-separated LDH labels, each 1–63 chars, starting and ending
// with a letter or digit) and a path-traversal-safe name. Because every label
// is non-empty and contains no interior dot, the name can never contain "..",
// a leading or trailing dot, or a leading hyphen — so it is always safe as a
// filesystem path component / storage key.
var subjectRegex = regexp.MustCompile(`^[a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?(\.[a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?)*$`)
```

internal/ca/signing.go
```diff
@@ -124,7 +127,7 @@ var subjectRegex = regexp.MustCompile(`^[a-z0-9._-]+$`)
 // NIST 800-53: SI-10 (Information Input Validation)
 func ValidateSubject(subject string) error {
 	if !subjectRegex.MatchString(subject) || strings.Contains(subject, "..") {
-		return fmt.Errorf("invalid subject name %q: must match ^[a-z0-9._-]+$ and must not contain ..", subject)
+		return fmt.Errorf("invalid subject name %q: must match ^[a-z0-9_.][a-z0-9._-]*$ and must not contain path traversal", subject)
 	}
 	return nil
 }
```
bootc, line 149:
```suggestion
// ValidateSubject returns an error if subject is not a valid lowercase DNS
// hostname safe for use as a path component.
//
// NIST 800-53: SI-10 (Information Input Validation)
func ValidateSubject(subject string) error {
	if len(subject) > 253 || !subjectRegex.MatchString(subject) {
		return fmt.Errorf("invalid subject name %q: must be a lowercase DNS hostname (RFC 1123), at most 253 characters", subject)
	}
	return nil
}
```

internal/storage/etcd.go
```diff
@@ -153,7 +153,7 @@ func NewEtcdBackendFromClient(cli *clientv3.Client, keyPrefix string, requestTim
 // for unknown logical keys or obviously unsafe components (e.g. "..").
 func (b *EtcdBackend) physicalKey(logical string) (string, error) {
 	if strings.Contains(logical, "..") {
-		return "", fmt.Errorf("invalid key %q: must not contain ..", logical)
+		return "", fmt.Errorf("invalid key %q: contains '..'", logical)
 	}
```
bootc: All of these "contains '..'" checks are identical between backends which feels like useless duplication. It should also now be covered by the updated `ValidateSubject()` above?

internal/storage/filesystem.go
```diff
@@ -80,7 +81,7 @@ func (b *FilesystemBackend) Path(key string) string {
 
 func (b *FilesystemBackend) pathFor(key string) (string, error) {
 	if strings.Contains(key, "..") {
-		return "", fmt.Errorf("invalid key %q: must not contain ..", key)
+		return "", fmt.Errorf("invalid key %q: contains '..'", key)
 	}
```
bootc: As above.

internal/storage/redis.go
```diff
@@ -255,7 +255,7 @@ func newRedisBackendFromClient(client redis.UniversalClient, owned bool, prefix
 // for unknown logical keys or obviously unsafe components (e.g. "..").
 func (b *RedisBackend) physicalKey(logical string) (string, error) {
 	if strings.Contains(logical, "..") {
-		return "", fmt.Errorf("invalid key %q: must not contain ..", logical)
+		return "", fmt.Errorf("invalid key %q: contains '..'", logical)
 	}
```
bootc: As above.

internal/storage/sql.go
```diff
@@ -260,7 +260,7 @@ func validateKey(key string) error {
 		return fmt.Errorf("empty key")
 	}
 	if strings.Contains(key, "..") {
-		return fmt.Errorf("invalid key %q: must not contain ..", key)
+		return fmt.Errorf("invalid key %q: contains '..'", key)
 	}
```
bootc: And again.

internal/storage/storage.go
```diff
@@ -197,7 +197,11 @@ func (s *StorageService) AppendInventory(ctx context.Context, entry string) erro
 
 	if s.hmacKey != nil {
 		if err := s.updateInventoryHMACLocked(ctx, s.hmacKey); err != nil {
-			slog.Warn("Failed to update inventory HMAC", "error", err)
+			// The line is already durably appended, but the stored HMAC now
+			// lags the inventory: the next verify would falsely report
+			// tampering. Surface the failure so the caller can react (e.g.
+			// roll back the just-written cert) rather than hiding it.
+			return fmt.Errorf("updating inventory HMAC after append: %w", err)
```
bootc, line 215: What can the user actually do about this, though? If we end up in this situation the user is pretty screwed, no? Perhaps we should write both the updated inventory and HMAC to temp files in the CA directory, then use renames to swap them around to avoid this kind of issue? Mind you we still have to bubble up errors somehow and even with a rename we could end up with a broken inventory.

cmd/puppet-ca/launcher.go
```diff
@@ -31,6 +31,20 @@ import (
...
 )
 
+const (
+	// gracefulShutdownTimeout bounds how long the launcher waits for both
+	// children to drain after forwarding SIGTERM before hard-killing them. It
+	// matches the frontend's own http.Server.Shutdown budget so the documented
+	// ~30s drain (and Kubernetes' default pod grace period) is honoured
+	// end-to-end in the default isolated-process deployment, not silently
+	// truncated by the supervisor.
+	gracefulShutdownTimeout = 30 * time.Second
```
bootc: This _matches_ the frontend timeout, and because the signal is delivered after this timeout starts the frontend won't actually necessarily have its own full 30s to shut things down gracefully. Either this needs to be a few seconds longer, or the frontend timeout a bit sorter, IMO.

Honestly because K8s has a 30s default termination grace period, perhaps this should actually be reduced (to 28s?) and the frontend timeout reduced even more (to 25s?). What do you think?

magefile.go
```diff
@@ -260,6 +260,7 @@ func (Test) Unit() error {
 		"./cmd/puppet-ca-ctl/...",
 		"./internal/api/...",
 		"./internal/ca/...",
+		"./internal/signer/...",
```
bootc: Is there any way we can ensure we don't miss more of these in future?

lefthook.yml
```diff
@@ -0,0 +1,36 @@
+# lefthook git-hook configuration for github.com/tvaughan/puppet-ca.
+#
+# The repository ships .git/hooks/pre-commit and .git/hooks/pre-push shims that
+# call `lefthook run <hook>`; this file is what those shims drive. Install the
```
bootc: No it doesn't? And it can't either. I've never used lefthook before; how does it differ from pre-commit, which I was going to suggest at some point? (genuine query, not trying to suggest a change)

bootc (changes requested):
I've had a good look at this, both with Claude and genuinely looking over it line by line myself. Looks like there's some room for various improvements! 😄

bootc (approved):
LGTM!

PR #48, opened 2026-06-20
Pin OpenVox Server to a working version

bootc:
Recent versions of the openvoxserver image have made significant changes to the container layout which unfortunately breaks our tests with errors like the following:

```
Attaching to old-puppet-1, test-runner-1
old-puppet-1   | Entrypoint PID 7
old-puppet-1   | Running /container-custom-entrypoint.d/pre-default/00-fix-cadir-ownership.sh
old-puppet-1   | chown: changing ownership of '/etc/puppetlabs/puppetserver/ca': Operation not permitted (os error 1)
Aborting on container exit...
```

This works around the problem for the time being.

### DIFF
--- a/compose-migration.yml
+++ b/compose-migration.yml
@@ -25,7 +25,7 @@ services:
 
   # -- Old Puppet Server (OpenVox Server, JVM) -------------------------------
   old-puppet:
-    image: ghcr.io/openvoxproject/openvoxserver:latest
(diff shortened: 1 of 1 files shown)


bastelfreak (approved):
Can you please raise a github issue so we don't forget to unpin? (or maybe renovate/dependabot for the image version?)

bootc:
Yeah, Renovate is now enabled in this repo with pinning enabled, and I expect it to update the pinned version etc... and those PRs to fail. I will create an issue in the container-openvoxserver repo to link back to those if there isn't one. Leave that with me!

PR #52, opened 2026-06-20
ci(renovate): pin GitHub Action digests to full semver

bootc:
## What

Extend Renovate with the `helpers:pinGitHubActionDigestsToSemver` preset so that the version comment beside each pinned GitHub Action digest carries the full semver (e.g. `# v7.0.0`) instead of the major-only tag (e.g. `# v7`).

## Why

`config:best-practices` already pins Action digests via `helpers:pinGitHubActionDigests`, but it keeps the trailing comment at whatever precision the tag was written with — so an update to `actions/checkout@v7` lands as `…@<sha> # v7`. The `…ToSemver` variant adds an `extractVersion`/`versioning` pair so the comment is maintained at full semver, making it clear exactly which release a digest corresponds to.

## Notes

- Applies going forward: existing major-only comments are rewritten the next time each action gets an update PR (or when re-pinned manually).
- Actions that only publish a major tag won't gain a fuller comment, since the semver extraction won't match.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,7 +2,8 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
 
   "extends": [
-    "config:best-practices"
+    "config:best-practices",
+    "helpers:pinGitHubActionDigestsToSemver"
   ],
 
   "rangeStrategy": "pin",


renovate.json
```diff
@@ -2,7 +2,8 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
 
   "extends": [
-    "config:best-practices"
+    "config:best-practices",
+    "helpers:pinGitHubActionDigestsToSemver"
```
bastelfreak, line 6: Interested in rolling this out to all renovate configs we have? (maybe discuss on slack/irc first)

bootc: As I mentioned in Slack, we may benefit from having org-wide Renovate configs but this particular option isn't something I've played with before so I'd like to experiment with it. When I asked Claude to research this it did point out https://github.com/renovatebot/renovate/issues/35789 which suggests this may not be good to enable universally?

PR #80, opened 2026-06-26
Add optional Kubernetes export of CA cert and CRL

bootc:
## Summary

Adds an **opt-in** feature to publish the CA certificate and/or CRL into any number of Kubernetes **Secrets** and **ConfigMaps**, so other workloads can mount them directly instead of fetching over the HTTP API or sharing a storage volume.

- Any number of targets, each a Secret **or** a ConfigMap.
- Each target carries the **CA cert**, the **CRL**, or **both** (PEM only for now).
- Configurable per target: `name`, `namespace`, `labels`, `annotations`, Secret `type`, and the data keys for the cert/CRL.
- Every managed object is labelled `app.kubernetes.io/managed-by: openvox-ca`.
- Objects are reconciled with **server-side apply** (idempotent create-or-update, force-owned) using the pod's **in-cluster ServiceAccount**.
- Disabled unless at least one target is configured.

## CRL freshness

CRL-bearing objects must stay current, so `ca.CA` gains a buffered, non-blocking `CRLUpdated()` notification fired from `signCRLLocked` — the single chokepoint for every in-frontend CRL write (revoke, reissue, background refresh, expired-cert cleanup). The exporter reconciles **once at startup** and again **on each CRL update**.

## Behaviour

- Runs only in the frontend / single-process role.
- **Auxiliary**: client/cluster failures are logged and never stop the CA; an invalid config block fails fast at startup.
- Config is **YAML-file only** (the nested target list doesn't map onto flags/env), mirroring how `StorageConfig` is handled.

## Notes / trade-offs

- Adds the **client-go** dependency — a deliberate exception to the otherwise stdlib-only posture, noted in the README. It does **not** break the FIPS build (`GOFIPS140=latest go build ./...` passes); the export path's crypto stays within stdlib `crypto/tls`, which both boringcrypto and the Go native FIPS module route through the validated module.
- In-cluster auth only; PEM only; objects are not deleted when a target is removed from config (they carry the managed-by label for manual cleanup).

## Docs

New [docs/kubernetes-export.md](docs/kubernetes-export.md) covers config reference, key defaults, and the required **RBAC** (`get`/`create`/`patch` on `secrets`/`configmaps`).

## Tests

Ginkgo + Gomega: config validation/defaulting, the apply builder and a fake-clientset round-trip, the `CRLUpdated()` notification (revoke/reissue/coalescing), and YAML parsing of the `kubernetes_export` block. Full `go test ./...`, `go vet`, and `golangci-lint` all green.


bootc:
@trevor-vaughan I'd like your opinion on this please; there's no rush to get this merged or anything so I'm prepared to wait. This PR adds a significant dependency so I wanted your opinion on whether this is appropriate based on your original vision for the CA.

I'd find this useful so that the CA CRL in particular can be kept up to date easily for Kubernetes workloads that need it, most notably OpenVox Server and OpenVox DB. Currently I have resorted to a sidecar container that polls the CRL every few minutes. This works well but it feels a bit ugly.

There are alternatives to this, of course. I could configure the External Secrets Operator to poll the API and create Secrets, but that wouldn't work for ConfigMaps. I fundamentally believe that certificates and CRLs are not secret and thus belong more in ConfigMaps than they do Secrets. I could also build my own simple operator to poll the API and do the same thing. Neither would get immediate updates when certificates are revoked, however.

That last part could potentially be fixed by building a webhook mechanism instead of this (or in addition to it), which would be a much lighter-weight feature than this.

trevor-vaughan:
Hey @bootc! I've got no issue with the addition in general. Before we move it forward, could you trace the crypto in the new library to make sure we're still using the FIPS crypto properly? I want to make sure we're honest about that contract.

bootc:
@trevor-vaughan I asked Claude to trace the crypto through the new dependency tree, and this write-up is what it came up with (I've reviewed it and the reproduction commands are included below). The conclusion: the FIPS contract holds.

**1. The PR adds no new crypto implementations.** Diffing the full set of packages linked into `cmd/openvox-ca` between `main` and this branch (`go list -deps ./cmd/openvox-ca` on each): the PR pulls in ~300 new packages (`k8s.io/*`, `sigs.k8s.io/*`, gnostic, cbor, `golang.org/x/oauth2`, etc.) and none of them implements crypto. Every crypto import made by the new subtree is standard library:

- `k8s.io/client-go/{rest,transport}`, `plugin/pkg/client/auth/exec` → `crypto/tls`, `crypto/x509`
- `k8s.io/client-go/util/{cert,keyutil}` → `crypto/rsa`, `crypto/ecdsa`, `crypto/elliptic`, `crypto/rand`, `crypto/x509`
- `k8s.io/kube-openapi/pkg/handler3` → `crypto/sha512` (ETag hashing)
- The `math/rand` uses in the subtree are non-cryptographic (backoff jitter, fuzz helpers)

Under `GOEXPERIMENT=boringcrypto` all stdlib crypto routes through the validated BoringCrypto module, so the entire client-go tree stays inside the FIPS boundary.

**2. All third-party crypto in the binary pre-dates this PR.** A reverse-import trace of every non-stdlib crypto package linked into the binary:

```
go list -deps -json ./cmd/openvox-ca | \
  jq -r '.ImportPath as $p | (.Imports // [])[] |
         select(test("^(golang.org/x/crypto|filippo.io)")) | "\(.) <= \($p)"' | sort -u
```

```
filippo.io/edwards25519        <= github.com/go-sql-driver/mysql   (MariaDB auth plugin)
golang.org/x/crypto/argon2     <= internal/ca                      (key-at-rest KDF)
golang.org/x/crypto/ocsp       <= internal/ca, internal/api        (OCSP encoding)
golang.org/x/crypto/pbkdf2     <= mellium.im/sasl                  (Postgres SCRAM)
```

Nothing from `k8s.io/*` — all four importers are already on `main`.

**3. The built FIPS binary demonstrably uses BoringCrypto.** Building this branch as `mage build:fips` does (linux/amd64, `GOEXPERIMENT=boringcrypto`, `CGO_ENABLED=1`):

- `go version -m` reports `go1.26.4-X:boringcrypto`
- the symbol table contains 474 `_goboringcrypto_` symbols, including the entry points the CA and the exporter actually exercise (`ECDSA_sign`, `RSA_sign`, `RSA_sign_pss_mgf1`, `SHA256_Init`, `EVP_AEAD_CTX_seal`, `ECDH_compute_key_fips`)
- a runtime probe under the same toolchain reports `crypto/boring.Enabled() == true` and signs with ECDSA P-256 through the boring engine
- `go test ./internal/k8sexport/...` passes under the boringcrypto toolchain

**4. CI exercises it end-to-end.** Both FIPS matrix jobs — *Compose integration tests (fips)* and *Puppet stack tests (fips)* — are green on this PR, running the full stack against the boringcrypto binary.

**Runtime behaviour of the new code:** the exporter's only crypto at runtime is `crypto/tls` + `crypto/x509` for the API server connection, authenticated with the ServiceAccount bearer token — entirely within the boring module.

**One honesty note**, since you asked about the contract: the only crypto in the binary that sits *outside* the validated module is the pre-existing Argon2id key-at-rest KDF and the `x/crypto/ocsp` encoding helpers. That's the existing state on `main` and is untouched by this PR — but if we want the FIPS wording in the README to be precise about it, I'm happy to add a note there too.

trevor-vaughan:
Seems reasonable :+1:

trevor-vaughan-ai:
> **Obsolete.** Superseded by the [current Review Council verdict](https://github.com/voxpupuli/openvox-ca/pull/80#issuecomment-5028796391) for commit `2dda9b1`. 

>
> Models used:
> - claude-opus-4-8

_Reviewed at commit [`c345fab`](https://github.com/voxpupuli/openvox-ca/commit/c345fab93b2b3768d62df8ca5fcc7f852f03c057) (deep effort)._

**TL;DR:** Well-built opt-in feature; no blockers found. Main advisory is missing tests for the cmd-level export loop, plus a least-privilege RBAC trim.

**Findings:** 🔴 0 Critical, 🟠 0 High, 🟡 4 Medium, 🔵 16 Low

- 🧪 **managed-by label "always wins" contract has no regression anchor** ([`internal/k8sexport/build.go:36`](https://github.com/voxpupuli/openvox-ca/blob/c345fab93b2b3768d62df8ca5fcc7f852f03c057/internal/k8sexport/build.go#L36))

  > `labels[managedByLabelKey] = managedByLabelValue`

  <br>

  with doc "The managed-by label always wins so ownership cannot be accidentally masked by configuration." Test grep: only `HaveKeyWithValue("app.kubernetes.io/managed-by", "openvox-ca")` at exporter_test.go:119,155, always with a non-colliding operator label (`{"app": "demo"}`).

  `labelsFor` copies operator labels first, then overwrites `app.kubernetes.io/managed-by`. The documented ownership guarantee is that an operator who sets `app.kubernetes.io/managed-by: something-else` is overridden. No test exercises the collision: existing tests only supply a differently-keyed label, so a regression that reordered the merge (operator value written last) would still pass. This is an ownership/security-flavored invariant left unlocked. Add a target whose `Metadata.Labels` contains `app.kubernetes.io/managed-by: intruder` and assert the applied object still carries `openvox-ca`.
  </details>

- 🧪 **Export loop `runK8sExporter`/`exportK8sOnce` has no test coverage** ([`cmd/openvox-ca/k8sexport.go:38`](https://github.com/voxpupuli/openvox-ca/blob/c345fab93b2b3768d62df8ca5fcc7f852f03c057/cmd/openvox-ca/k8sexport.go#L38))

  > `func runK8sExporter(ctx context.Context, c *ca.CA, exporter *k8sexport.Exporter) {`

  <br>

  grep for the only references:

  ```
  cmd/openvox-ca/k8sexport.go:38:func runK8sExporter(ctx context.Context, c *ca.CA, exporter *k8sexport.Exporter) {
  cmd/openvox-ca/k8sexport.go:57:func exportK8sOnce(ctx context.Context, exporter *k8sexport.Exporter) {
  cmd/openvox-ca/main.go:680:					go runK8sExporter(ctx, myCA, k8sExporter)
  ```
  No k8sexport_test.go exists under cmd/openvox-ca/.

  This is the integration heart of the feature and it is entirely untested. Three behaviors the doc comment promises are unverified by any test: (1) exports once at startup; (2) re-exports on every c.CRLUpdated() receive; (3) terminates on ctx.Done(). Every dependency is tested in isolation - CRLUpdated() in internal/ca/crl_notify_test.go, ExportAll re-export in internal/k8sexport/exporter_test.go:171 - but the wiring that connects the channel to ExportAll is not. A regression that dropped the startup export, selected the wrong channel, or failed to return on cancellation would pass the entire suite. Fix: add a cmd-level test that builds an Exporter via k8sexport.New(fake.NewClientset(), ...) and a real *ca.CA, runs runK8sExporter in a goroutine, asserts an apply happened at startup, triggers a CRL update and asserts a second apply, then cancels ctx and asserts the goroutine returns - or introduce a small interface to allow a spy exporter.
  </details>

- 🧭 **Feature comment claims a setup failure "never stops the CA," but a config-validation failure does** ([`cmd/openvox-ca/main.go:662`](https://github.com/voxpupuli/openvox-ca/blob/c345fab93b2b3768d62df8ca5fcc7f852f03c057/cmd/openvox-ca/main.go#L662))

  > `// but never stops the CA from serving.`

  <br>

  Comment: `// configured Secrets/ConfigMaps. Auxiliary - a setup failure is logged` / `// but never stops the CA from serving.` immediately followed by:

  ```go
  if cfg.KubernetesExport.Enabled() {
      if err := cfg.KubernetesExport.Validate(); err != nil {
          return fmt.Errorf("invalid kubernetes_export config: %w", err)
      }
  ```
  The block comment states that a failure in this optional feature is logged but "never stops the CA from serving." That holds for the NewInCluster path (line 676-678) and for runtime export errors, but not for the Validate() path directly above it, which returns an error and aborts server startup. The two failure modes have opposite behavior, and the comment describes only one. The fail-fast-on-bad-config behavior itself is correct and consistent with how StorageConfig is validated; it is the comment that misstates it.
  Recommendation: Tighten the comment to distinguish the two paths - an invalid config is rejected at startup (fail-fast) while client-initialisation and per-cycle export failures are logged and never stop the CA.
  </details>

- 🛡️ **RBAC example grants unused `get` on Secrets, exceeding stated minimum** ([`docs/kubernetes-export.md:125`](https://github.com/voxpupuli/openvox-ca/blob/c345fab93b2b3768d62df8ca5fcc7f852f03c057/docs/kubernetes-export.md#L125))

  > `verbs: ["get", "create", "patch"]`

  <br>

  ```
  Server-side apply is a `patch`, so
  `patch` is required in addition to `get` and `create`:
  ...
      verbs: ["get", "create", "patch"]
  ```

  The export code path never issues a read. Exporter.applyTarget (internal/k8sexport/exporter.go:168,171) calls only Secrets(ns).Apply(...) / ConfigMaps(ns).Apply(...), which is a single SSA PATCH; there is no Get/List call in ExportAll/applyTarget/fetchMaterials. Server-side apply does not consult the get verb.

  Consequently the documented rule grants get on secrets in the target namespace, i.e. read access to every Secret in that namespace (other apps' TLS keys, DB passwords, etc.) - a capability the exporter demonstrably does not use. The same section tells operators to "Restrict the verbs and resources to the minimum," but get is above that minimum. Since this is copy-paste RBAC on the most sensitive resource type, the over-grant is worth correcting.

  Fix: reduce the example to verbs: ["create", "patch"] (drop get), and adjust the surrounding prose from "in addition to get and create" to "in addition to create". If a get is ever added to the code path, reintroduce it then. This is a non-blocking least-privilege improvement, not a break - the documented config is functional and namespaced.
  </details>

</details>

<details><summary>🔵 LOW (16)</summary>

- 🧪 **Cert error and empty-cert paths untested; only the CRL equivalents are covered** ([`internal/k8sexport/exporter.go:124`](https://github.com/voxpupuli/openvox-ca/blob/c345fab93b2b3768d62df8ca5fcc7f852f03c057/internal/k8sexport/exporter.go#L124))

  > `certPEM, err = e.src.GetCACert(ctx)`

  <br>

  exporter.go:124-127 `certPEM, err = e.src.GetCACert(ctx) ... "reading CA certificate for export: %w"` and exporter.go:158-160 `if t.Cert && len(certPEM) == 0 { return ... "refusing to export an empty CA certificate" }`. Test grep for `reading CA` / `empty CA`: none. `certErr` is set only at exporter_test.go:278, where the cert is deliberately *not* requested (asserts it is skipped).

  The CRL branch has both a read-error test (exporter_test.go:191, "reading CRL") and an empty-material test (exporter_test.go:205, "empty CRL"), but the symmetric cert branches (`fetchMaterials` cert read error, `applyTarget` empty-cert refusal) have no coverage. A cert-only target that fails to read or returns empty PEM is a realistic CA-degraded state; these are the exact clobber-prevention guards the design calls out. Add cert-requesting targets with `src.certErr` set and with `src.cert = nil` and assert the respective error substrings, plus that no object is created.
  </details>

- 🧪 **Annotation propagation into applied objects is unverified** ([`internal/k8sexport/build.go:89`](https://github.com/voxpupuli/openvox-ca/blob/c345fab93b2b3768d62df8ca5fcc7f852f03c057/internal/k8sexport/build.go#L89))

  > `WithAnnotations(t.Metadata.Annotations)`

  <br>

  build.go:89-91 and 101-103 apply `WithAnnotations(t.Metadata.Annotations)`. Test grep (`-i annotation`) over `*_test.go`: no matches. `Metadata.Annotations` is a configurable field (config.go:68).

  Labels flow-through is asserted, but the parallel annotations path has zero assertions, so the entire `WithAnnotations` branch for both Secret and ConfigMap builders is untested. Add annotations to a target's metadata and assert `sec.Annotations` / `cm.Annotations` contain them.
  </details>

- ⚙️ **Kubernetes apply calls have no bounded per-attempt timeout** ([`internal/k8sexport/exporter.go:168`](https://github.com/voxpupuli/openvox-ca/blob/c345fab93b2b3768d62df8ca5fcc7f852f03c057/internal/k8sexport/exporter.go#L168))

  > `_, err := e.client.CoreV1().Secrets(ns).Apply(ctx, t.buildSecretApply(ns, certPEM, crlPEM), opts)`

  <br>

  The only deadline is the caller's ctx, which is the process-lifetime context (`go runK8sExporter(ctx, ...)` in cmd/openvox-ca/main.go:680). grep for `Timeout`/`backoff`/`retry` across internal/k8sexport and cmd/openvox-ca/k8sexport.go returns nothing, and rest.InClusterConfig() sets no default request timeout.

  The reconcile loop calls exportK8sOnce synchronously (cmd/openvox-ca/k8sexport.go:43-52), and the clientset built from InClusterConfig carries no request timeout. If the API server becomes unreachable in a way that black-holes the TCP connection, an .Apply call blocks until the OS transport timeout (potentially minutes), during which the single exporter goroutine cannot service further CRLUpdated() wake-ups and target objects stay stale. Impact is bounded: the CA itself is never blocked (the notify channel is non-blocking), shutdown still unblocks the call via ctx cancellation, and the next CRL update or a restart re-reconciles - hence LOW, not higher. But there is no upper bound on how long a single stuck apply delays re-export.

  Recommendation: Derive a bounded context per cycle or per target, e.g. wrap the apply in context.WithTimeout(ctx, someBound) inside applyTarget (or wrap ExportAll at the call site), so a hung API server surfaces as a counted/logged error and lets the loop proceed to the next wake-up rather than blocking indefinitely. Retry/backoff is not needed - event-driven re-export on the next CRL update already self-heals transient failures.
  </details>

- 📚 **Const-block doc comment names a nonexistent identifier** ([`internal/k8sexport/build.go:25`](https://github.com/voxpupuli/openvox-ca/blob/c345fab93b2b3768d62df8ca5fcc7f852f03c057/internal/k8sexport/build.go#L25))

  > `// managedByLabel marks every object this exporter maintains so operators can`

  <br>

  Lines 25-31:

  ```
  // managedByLabel marks every object this exporter maintains so operators can
  // identify and select resources owned by openvox-ca. It is always present and
  // takes precedence over any operator-supplied value for the same key.
  const (
  	managedByLabelKey   = "app.kubernetes.io/managed-by"
  	managedByLabelValue = "openvox-ca"
  )
  ```

  The comment opens with `managedByLabel`, but no identifier by that exact name exists; the block declares `managedByLabelKey` and `managedByLabelValue`. Non-blocking and purely cosmetic - the constants are unexported, so no exported-doc lint rule fires.

  Recommendation: Reword the opening to reference the actual pair, e.g. "managedByLabelKey/managedByLabelValue mark every object this exporter maintains...", or phrase it generically ("The managed-by label marks...").
  </details>

- 🧪 **Revoke test drains a signal that the setup never produces** ([`internal/ca/crl_notify_test.go:72`](https://github.com/voxpupuli/openvox-ca/blob/c345fab93b2b3768d62df8ca5fcc7f852f03c057/internal/ca/crl_notify_test.go#L72))

  > `// Drain any pending signal so we observe the one from the revoke.`

  <br>

  ```go
  // Drain any pending signal so we observe the one from the revoke.
  select {
  case <-myCA.CRLUpdated():
  default:
  }
  ```
  Neither Init (load-existing fast path) nor Sign (signing.go:156, not a signCRLLocked caller) fires a notification, so the buffer is provably empty here - the drain is dead defensive code. It is harmless but misleads a future reader into thinking Sign/Init emit notifications. Either remove it, or add a comment noting it guards against future setup changes.
  </details>

- 📚 **Chokepoint invariant not documented at signCRLLocked godoc** ([`internal/ca/crl.go:30`](https://github.com/voxpupuli/openvox-ca/blob/c345fab93b2b3768d62df8ca5fcc7f852f03c057/internal/ca/crl.go#L30))

  > `signCRLLocked re-signs the CRL with the given revoked entries`

  <br>

  signCRLLocked godoc reads: "signCRLLocked re-signs the CRL with the given revoked entries, bumping the CRL number past prevNumber and stamping fresh ThisUpdate/NextUpdate. It writes the result to storage and refreshes the in-memory cache." No mention that the CRL-updated notification depends on this being the sole re-sign path. grep shows CRLs are also signed outside it: internal/ca/init.go:160, internal/ca/init.go:415, internal/ca/caImport.go:123 all call x509.CreateRevocationList directly.

  The CRLUpdated()/crlNotify contract holds only because every CRL re-sign for the documented triggers routes through signCRLLocked, where the sole crlNotify send lives. A future maintainer adding another re-sign path outside signCRLLocked would silently drop consumer notifications. This is a maintainability doc gap, not a functional defect; the current code is correct.

  Recommendation: Add one line to the signCRLLocked godoc noting it is the single point through which CRL re-signs are signalled to consumers via crlNotify/CRLUpdated(), so new CRL-signing paths either route through it or deliberately handle notification. No code change required.
  </details>

- 🧪 **No test that an invalid `kubernetes_export` block fails startup** ([`cmd/openvox-ca/main.go:668`](https://github.com/voxpupuli/openvox-ca/blob/c345fab93b2b3768d62df8ca5fcc7f852f03c057/cmd/openvox-ca/main.go#L668))

  > `if err := cfg.KubernetesExport.Validate(); err != nil {`

  <br>

  startup rejects invalid config -

  ```
  if err := cfg.KubernetesExport.Validate(); err != nil {
      return fmt.Errorf("invalid kubernetes_export config: %w", err)
  }
  ```
  The only cmd-level assertion of this path is the happy case in config_test.go:413. No test loads a config with an invalid target and asserts the failure/wrapped error. Validate negatives are covered inside internal/k8sexport/config_test.go:98, so the raw validator is safe - but the cmd contract "a malformed block aborts startup with `invalid kubernetes_export config:`" is unverified. Fix: add a config_test.go case that parses a block with an invalid target and asserts cfg.KubernetesExport.Validate() returns an error.
  </details>

- 🧪 **Config-parse test omits assertions on defaulting/normalization applied by `Validate`** ([`cmd/openvox-ca/config_test.go:413`](https://github.com/voxpupuli/openvox-ca/blob/c345fab93b2b3768d62df8ca5fcc7f852f03c057/cmd/openvox-ca/config_test.go#L413))

  > `Expect(cfg.KubernetesExport.Validate()).To(Succeed())`

  <br>

  after Validate() the test asserts only Targets[1].Kind == "ConfigMap". The YAML sets cert_key/crl_key and type on target 0, and target 1 relies on Validate defaulting CRLKey to ca.crl - none of these post-Validate values are checked. The Validate mutation contract is exercised in internal/k8sexport/config_test.go, so this is a minor completeness gap. Fix (optional): assert Targets[0].Type == "Opaque", Targets[0].CertKey == "ca.crt", and Targets[1].CRLKey == "ca.crl" after Validate.
  </details>

- 🧭 **"File-only … mirroring how StorageConfig is handled" misdescribes StorageConfig** ([`cmd/openvox-ca/config.go:124`](https://github.com/voxpupuli/openvox-ca/blob/c345fab93b2b3768d62df8ca5fcc7f852f03c057/cmd/openvox-ca/config.go#L124))

  > `// impractical to express as flags/env), mirroring how StorageConfig is`

  <br>

  `applyServerEnv` contains 34 PUPPET_CA_STORAGE_BACKEND / PUPPET_CA_ETCD_* / PUPPET_CA_REDIS_* / PUPPET_CA_SQL_* / PUPPET_CA_CA_CERT_FILE / PUPPET_CA_CA_KEY_FILE env overrides (config.go:392-511). The comment characterises KubernetesExport as "File-only … mirroring how StorageConfig is handled." StorageConfig is precisely not file-only: it has an extensive set of environment-variable overrides. The genuine point - that the nested target list, labels, and annotations are impractical as flags/env and so are file-only - stands on its own; the "mirroring how StorageConfig is handled" clause attaches the file-only property to a struct that does not have it.
  Recommendation: Drop or rephrase the "mirroring how StorageConfig is handled" clause.
  </details>

- ⚙️ **Inline comment overstates non-fatality; invalid config is fatal at startup** ([`cmd/openvox-ca/main.go:664`](https://github.com/voxpupuli/openvox-ca/blob/c345fab93b2b3768d62df8ca5fcc7f852f03c057/cmd/openvox-ca/main.go#L664))

  > `// but never stops the CA from serving. Bound to ctx so it stops on`

  <br>

  The block comment (main.go:661-665) frames the export as "Auxiliary - a setup failure is logged but never stops the CA from serving." That holds for the client-init path (main.go:677-678), but the preceding Validate() failure (main.go:667-669) returns an error that aborts server startup entirely. Failing fast on invalid config is consistent with the rest of this startup path, so the behavior itself is defensible - but the comment misleads an operator into believing a malformed kubernetes_export block is non-fatal. Recommend narrowing the comment, e.g. "an invalid config aborts startup; a client-init failure is logged and the CA keeps serving."
  </details>

- ⚙️ **Two pre-existing indirect deps moved from release tags to pre-release pseudo-versions** ([`go.mod:69`](https://github.com/voxpupuli/openvox-ca/blob/c345fab93b2b3768d62df8ca5fcc7f852f03c057/go.mod#L69))

  > `github.com/gorilla/websocket v1.5.4-0.20250319132907-e064f32e3674 // indirect`

  <br>

  The client-go v0.36.2 bump drags gorilla/websocket from released tag v1.5.3 to a pre-release pseudo-version, and google.golang.org/protobuf from v1.36.11 to v1.36.12-0.20260120151049-f2248ac996af (go.mod:143). Both remain // indirect, consistent with what the k8s.io v0.36 module set requires under MVS. Operationally safe (pinned to exact commits with matching go.sum entries), but the project now rides unreleased upstream commits. Recommendation: confirm this is the accepted consequence of adopting client-go v0.36, and note it so a future go get -u is not mistaken for a regression. No action required to merge.
  </details>

- ⚙️ **client-go adds a large transitive tree, growing build surface and binary size** ([`go.mod:29`](https://github.com/voxpupuli/openvox-ca/blob/c345fab93b2b3768d62df8ca5fcc7f852f03c057/go.mod#L29))

  > `k8s.io/client-go v0.36.2`

  <br>

  Pulling in client-go/apimachinery/api adds roughly twenty new indirect modules. Expected for a Kubernetes client and not a defect, but it measurably enlarges the compile surface and output binary for an optional feature. Recommendation: none blocking - if binary size becomes a concern later, gating the k8s client behind a build tag is an option.
  </details>

- 🧪 **"no namespace resolved" error branch in applyTarget is never exercised** ([`internal/k8sexport/exporter.go:152`](https://github.com/voxpupuli/openvox-ca/blob/c345fab93b2b3768d62df8ca5fcc7f852f03c057/internal/k8sexport/exporter.go#L152))

  > `if ns == "" { return fmt.Errorf("no namespace resolved") }`

  <br>

  exporter.go:151-153. Test grep for `no namespace`: none. The default-namespace test (exporter_test.go:158) always supplies `New(..., "default-ns", nil)`.

  A target with no `Metadata.Namespace` constructed via `New` with an empty `defaultNS` reaches this guard and fails at apply time. No test drives it, so a regression removing or inverting the guard (e.g. applying into the empty-string namespace) would go unnoticed. Add a case: target without namespace, `New(..., "", nil)`, assert `ExportAll` errors with "no namespace resolved" and nothing is created.
  </details>

- 🧪 **needsDefaultNamespace / NewInCluster path is entirely untested** ([`internal/k8sexport/exporter.go:79`](https://github.com/voxpupuli/openvox-ca/blob/c345fab93b2b3768d62df8ca5fcc7f852f03c057/internal/k8sexport/exporter.go#L79))

  > `needsDefaultNamespace`

  <br>

  exporter.go:79-86 `needsDefaultNamespace` loops targets for an empty namespace; called only from `NewInCluster` (exporter.go:67). Test grep for `needsDefaultNamespace|NewInCluster|podNamespace|newInClusterClientset`: none.

  `newInClusterClientset` and `podNamespace` (client.go) legitimately need an in-cluster environment and are reasonably left to integration coverage, but `needsDefaultNamespace` is a pure function with no such dependency and gates whether namespace resolution runs. It is small, but an inverted condition would silently change in-cluster startup behavior. A two-case unit test (all targets namespaced -> false; one bare -> true) would lock it down cheaply.
  </details>

- 🧪 **Notification untested on two of four documented trigger paths** ([`internal/ca/crl_notify_test.go:59`](https://github.com/voxpupuli/openvox-ca/blob/c345fab93b2b3768d62df8ca5fcc7f852f03c057/internal/ca/crl_notify_test.go#L59))

  > `// re-signed (revoke, reissue, background refresh, or expired-cert cleanup).`

  <br>

  - Contract, ca.go:141: `// re-signed (revoke, reissue, background refresh, or expired-cert cleanup).`
  - signCRLLocked callers (grep): revoke.go:97 (revoke), crl.go:104 (reissue), crl.go:145 (RefreshCRLIfDue - background refresh), cleanup.go:138 (dropCRLEntriesLocked - expired-cert cleanup).
  - grep -rn CRLUpdated internal/ca/ -> matches only in crl_notify_test.go; the two It blocks exercise ReissueCRL (line 60, 86) and Revoke (line 78). No test drives RefreshCRLIfDue or the cleanup path and asserts a signal.

  The CRLUpdated doc explicitly promises the consumer (K8s exporter) a signal on all four re-sign paths, but only revoke and reissue are verified. Currently all four funnel through signCRLLocked, where the single select/send lives, so the mechanism is exercised - this is why it is MEDIUM, not HIGH. The gap is regression protection: if a later refactor moved the send into individual callers (or a caller stopped routing through signCRLLocked), the background-refresh and cleanup notifications would silently stop and no test would fail. Add two It blocks: one revoking a cert then calling cleanup to drop the now-expired entry (asserting Receive), and one driving RefreshCRLIfDue with a short refreshBefore so it re-signs (asserting Receive).
  </details>

- 📚 **metrics.md "Alerting" summary omits the new export-failing alert** ([`docs/metrics.md:158`](https://github.com/voxpupuli/openvox-ca/blob/c345fab93b2b3768d62df8ca5fcc7f852f03c057/docs/metrics.md#L158))

  > `It alerts on exporter availability, CA/CRL/leaf expiry, and pending requests, with all thresholds configurable.`

  <br>

  This changeset adds the PuppetCAKubernetesExportFailing alert (mixin/alerts.libsonnet:268) and updates mixin/README.md:251-252 to list it, but the parallel one-line enumeration in the metrics.md "## Alerting" section still lists only availability, expiry, and pending requests. Cosmetic staleness - the intro at docs/metrics.md:9-11 has the same omission. Non-blocking; append "and Kubernetes export failures" to the sentence for consistency with mixin/README.md.
  </details>

</details>

---

bootc:
Thanks — worked through the advisories in 2dda9b1. Summary of what changed and what I left.

### Medium (all 4 addressed)

- **Export loop untested** — added a cmd-level test driving `runK8sExporter` against a fake clientset and a real `*ca.CA`: asserts the startup export, a re-export on a real CRL update, and a clean stop on context cancellation.
- **managed-by "always wins" had no regression anchor** — added a test with a target that sets `app.kubernetes.io/managed-by: intruder` and asserts the applied object still carries `openvox-ca`.
- **"never stops the CA" comment vs. fatal config validation** — tightened the `main.go` comment to distinguish the fail-fast `Validate()` path from the logged client-init/per-cycle failures. Also fixed the duplicate low finding on the same comment.
- **RBAC over-grant** — dropped the unused `get` verb from the docs example and reworded the surrounding prose; the exporter only ever creates/patches (SSA), never reads.

### Low (addressed)

- Cert read-error and empty-cert guards now tested (symmetric with the existing CRL cases).
- Annotation propagation tested for both Secret and ConfigMap.
- `no namespace resolved` guard tested.
- `needsDefaultNamespace` covered by a small internal two-case test.
- Notification tests added for the **background-refresh** and **expired-cert-cleanup** paths (the cleanup one signs, revokes, then makes the entry look expired so cleanup drops it from the CRL) — all four documented trigger paths now assert a signal.
- `config_test`: added an invalid-block-fails-validation case and post-`Validate` defaulting/normalisation assertions.
- Per-apply **timeout** (30s) added in `applyTarget` so a black-holed API server can't stall the single exporter goroutine indefinitely.
- Comment/doc fixes: `signCRLLocked` godoc now notes it's the single CRL-notify chokepoint; `managedByLabel` const doc reworded to name the real identifiers; dropped the inaccurate "mirroring StorageConfig" clause; annotated the drain in the revoke notification test; metrics.md now mentions the Kubernetes-export alert in both the intro and the Alerting summary.

### Not addressed (with reasons)

- **go.mod: pre-release pseudo-versions for gorilla/websocket and protobuf**, and **client-go's large transitive tree** — both are inherent consequences of adopting client-go v0.36 under MVS, and the review flagged both as informational / "no action required to merge". Pinning or build-tag gating would be a larger, separate decision; noting here that the pseudo-versions are the expected result of the client-go bump, not a stray `go get -u`.

No crypto path or dependency changed by this commit, so the FIPS build contract is unaffected. `go build`, `go vet`, full `go test ./...` and `golangci-lint` are all green.

trevor-vaughan-ai:

>
> Models used:
> - claude-opus-4-8[1m]

_Reviewed at commit [`2dda9b1`](https://github.com/voxpupuli/openvox-ca/commit/2dda9b10f47337e2034b53f78fb586964831d886) (deep effort)._

**TL;DR:** Approve with advisories: all five reviewers approved across five subsystems; 7 non-blocking findings (2 medium regression-protection gaps, 5 low doc/tooling notes).

**Findings:** 🔴 0 Critical, 🟠 0 High, 🟡 2 Medium, 🔵 5 Low

- 🧪 **Concurrent wiring test is not exercised under the race detector** ([`magefile.go:301`](https://github.com/voxpupuli/openvox-ca/blob/2dda9b10f47337e2034b53f78fb586964831d886/magefile.go#L301))

  > `testArgs := append([]string{"test", "-json", "-cover", "-coverprofile=coverage.out"}, pkgs...)`

  💡 **Recommendation:**
  - Add `-race` to the unit test invocation (either globally in `Test.Unit`, or a dedicated `-race` job covering the `cmd/openvox-ca` and `internal/ca`/`internal/k8sexport` packages) so the new CRL-notification-to-export concurrency is checked.

    **APPROVE** - The added tests are well-architected: the goroutine test verifies startup export, CRL-driven re-export, and clean context-cancel shutdown using atomic counters and `Eventually` (no sleeps, no unsynchronized shared state), and the config tests cover both positive parsing/defaulting and negative validation with specific-value and specific-error assertions. The only finding is a MEDIUM, pre-existing infrastructure gap: CI runs the unit suite without `-race`, leaving this new concurrent wiring's race safety unverified. No HIGH or CRITICAL issues, so this does not block merge.

  <br>

  **Constraint:**
  - Go pack `testing_conventions` - "Race detection: `go test -race` (flag missing race detection in concurrent code tests)."

  **Description:**
  - `Test.Unit` (invoked by CI as `mage test:unit`, ci.yml:73) runs `go test` without `-race`. This PR adds a genuinely concurrent path: `runK8sExporter` (k8sexport.go:38-53) reads `c.CRLUpdated()` from a goroutine while `ReissueCRL`/`signCRLLocked` write to `crlNotify` on another, and `k8sexport_test.go` spins up exactly this goroutine (lines 62-65). Because the suite never enables the race detector, a data race introduced in this notification/export path - or in the exporter it drives - would pass CI undetected. The test itself is correctly synchronized (atomics + coalesced buffered channel), so this is a coverage/observability gap on new concurrent code rather than a live race; the missing `-race` is a pre-existing, repo-wide magefile choice outside this changeset, which is why this is MEDIUM and not HIGH.
  </details>

- 🧪 **Failure-isolation ("others proceed") is only tested with the failing target last, so it cannot catch an early-return regression** ([`internal/k8sexport/exporter_test.go:221`](https://github.com/voxpupuli/openvox-ca/blob/2dda9b10f47337e2034b53f78fb586964831d886/internal/k8sexport/exporter_test.go#L221))

  > `{Secret "good"}, {ConfigMap "bad"}`

  💡 **Recommendation:**
  - Add a spec where the failing target comes first (e.g. fail `secrets` patches, put the failing Secret at `Targets[0]` and a succeeding ConfigMap at `Targets[1]`), then `Get` the second target from the fake clientset and assert it was actually created with its expected data - proving the loop proceeded past the earlier failure.

    **APPROVE** - Coverage and assertion depth are well above the "apply was called" bar this review was chartered to guard against: applied object contents, label/annotation/namespace/type semantics, config negative paths, and error/empty-material handling are all verified against the fake clientset with concrete value assertions, and specs are properly isolated. The single finding is a MEDIUM regression-protection blind spot (failure-isolation is only tested with the failing target last, so it can't catch a regress-to-early-return), not a correctness defect or a missing-assertion problem, so it does not block merge.

  <br>

  ```go
  cfg := &k8sexport.Config{Targets: []k8sexport.Target{
      {Kind: "Secret", Metadata: k8sexport.Metadata{Name: "good", Namespace: "ns1"}, CRL: true},
      {Kind: "ConfigMap", Metadata: k8sexport.Metadata{Name: "bad", Namespace: "ns1"}, CRL: true},
  }}
  ...
  client.PrependReactor("patch", "configmaps",
      func(ktesting.Action) (bool, runtime.Object, error) { return true, nil, errors.New("boom") })
  ```
  The failing target is `Targets[1]` (last). In `exporter.go:109-122` the loop applies `Targets[0]` (success) before `Targets[1]` fails, so the post-failure `continue` (line 117) is never distinguished from a `return errors.Join(errs...)`: with the failure last, "keep going" and "abort" produce identical observable results.

  **Constraint:**
  - Coverage completeness / regression protection - the documented contract "A failure applying one target ... does not prevent the others from being applied" (exporter.go:99-101) is the resilience point of `ExportAll` but is not locked down.

  **Description:**
  - No spec exercises a failure on an *earlier* target followed by a later target that must still be applied. If someone refactored the loop to bail out on the first error (`return` instead of `continue`), the suite would still pass because the sole failing-target test puts the failure at the end. The suite does verify per-target success/error metrics, but that only proves the two-target error joining, not that a mid-list failure lets subsequent targets through.
  </details>

</details>

<details><summary>🔵 LOW (5)</summary>

- 📚 **"Single point" invariant comment omits the existing non-notifying CRL-write paths** ([`internal/ca/crl.go:36`](https://github.com/voxpupuli/openvox-ca/blob/2dda9b10f47337e2034b53f78fb586964831d886/internal/ca/crl.go#L36))

  > `// This is the single point through which CRL re-signs are signalled to`

  💡 **Recommendation:**
  - Tighten the invariant note to say that any CRL write reachable *while the server is serving* must route through `signCRLLocked`, and acknowledge that the bootstrap/import writes (`init.go`, `caImport.go`) intentionally skip notification because they run before any consumer starts and are covered by the exporter's startup reconcile. A one-line back-reference comment at each of those four `UpdateCRL` sites ("bootstrap-only; no consumer yet, so no crlNotify needed") would close the loop for a maintainer working from those sites rather than from `signCRLLocked`.

  <br>

  **Constraint:**
  - Invariant documentation completeness - a safety-relevant invariant comment must let a future maintainer tell which write paths must notify and which legitimately need not.

  **Description:**
  - The comment is accurate that `signCRLLocked` is the sole `crlNotify` *send* site, but it is framed around "CRL-signing path". In practice there are four direct `Storage.UpdateCRL` writes that bypass `signCRLLocked` and fire no notification: `init.go:165`, `init.go:420`, `caImport.go:112`, `caImport.go:128`. These are correct today (bootstrap/import run before any consumer exists, and the exporter reconciles all targets once at startup - see `cmd/openvox-ca/k8sexport.go:48`), but that reasoning is written nowhere. A maintainer who copies the `init.go` direct-write pattern into a path reachable while the server is serving would silently drop consumer wake-ups, and the warning that would stop them lives only inside `signCRLLocked`, which such a write path does not touch. This is the exact "bypassing CRL-write path" hazard the invariant is meant to guard, so the comment's "signing path" framing leaves a real (if narrow) gap. Non-blocking: the existing warning substantially mitigates the risk, and no current path is wrong.
  </details>

- 📚 **Package doc comment omits the in-cluster-only constraint** ([`internal/k8sexport/config.go:18`](https://github.com/voxpupuli/openvox-ca/blob/2dda9b10f47337e2034b53f78fb586964831d886/internal/k8sexport/config.go#L18))

  > `// Package k8sexport publishes the CA certificate and/or CRL into Kubernetes // Secrets and ConfigMaps. It is an optional feature: when no targets are // configured nothing in this package runs. Objects are reconciled with // server-side apply so each export is an idempotent create-or-update.`

  💡 **Recommendation:**
  - Add one sentence to the package comment, e.g. "The exporter authenticates with the pod's in-cluster ServiceAccount, so it is only available when openvox-ca runs inside a Kubernetes cluster." Non-blocking.

    **APPROVE** - The new `k8sexport` package is thoroughly and accurately documented: every exported type and function has a contract-bearing godoc comment, and the config-field comments are consistent with `docs/kubernetes-export.md` on defaults (`ca.crt`/`ca.crl`), the "unset type means unmanaged" semantics, and the mandatory managed-by label. I found no code/doc mismatches and no stale comments. The single LOW is a marginal godoc-placement nit (the in-cluster-only constraint is documented on the constructor and in the user docs but not in the package overview) and does not block merge.

  <br>

  **Constraint:**
  - Godoc convention - the package doc is the top-level overview a reader sees first; a feature's most significant operational constraint belongs there.

  **Description:**
  - The in-cluster-only limitation is the package's headline constraint and is documented elsewhere - on `NewInCluster` (exporter.go:66-68), `newInClusterClientset` (client.go:33-35: "openvox-ca must run inside a pod for kubernetes_export"), and prominently in `docs/kubernetes-export.md` ("It is therefore only available when openvox-ca itself runs inside a Kubernetes cluster."). A reader who lands on the package godoc first will not see it until they reach the constructor. This is a minor completeness gap, not a mismatch: nothing documented is contradictory.
  </details>

- 🛡️ **No `govulncheck` gate over the new client-go transitive surface** ([`.github/workflows/ci.yml:17`](https://github.com/voxpupuli/openvox-ca/blob/2dda9b10f47337e2034b53f78fb586964831d886/.github/workflows/ci.yml#L17))

  > `no govulncheck/osv/trivy/gosec step in ci.yml or codeql.yml; CI runs CodeQL (SAST) only`

  💡 **Recommendation:**
  - Add a `govulncheck ./...` step to `ci.yml` (Go-native, reads the same `go.mod`/`go.sum`) so the enlarged transitive surface is checked against the Go vuln DB on every PR, complementing CodeQL and Renovate. Optional; does not block this merge.

  <br>

  **Constraint:**
  - NIST SSDF PW.4 / RV.1 (verify acquired components; monitor for known vulnerabilities) - OWASP A06:2021.

  **Description:**
  - This PR adds `k8s.io/client-go` plus a large indirect graph (`json-iterator/go`, `mailru/easyjson`, `gnostic-models`, `fxamacker/cbor/v2`, `modern-go/reflect2`, `sigs.k8s.io/*`). CodeQL scans the project's own source, not third-party module versions against the Go vulnerability database, so a known CVE in a pinned client-go dependency would not be caught by a build/test gate. Renovate (`config:best-practices`, `minimumReleaseAge: 2 days`) provides asynchronous advisory alerts, which materially lowers the risk but is not a blocking pipeline check. I found no in-repo evidence of any specific CVE in these versions and do not claim one exists; this is a coverage advisory only, hence LOW and non-blocking.
  </details>

- 📚 **New `k8sExportFailingFor` tunable missing from mixin configurable-thresholds table** ([`mixin/README.md:91`](https://github.com/voxpupuli/openvox-ca/blob/2dda9b10f47337e2034b53f78fb586964831d886/mixin/README.md#L91))

  > `k8sExportFailingFor: '15m' is defined in mixin/config.libsonnet and used in alerts.libsonnet, but absent from the mixin/README thresholds table`

  💡 **Recommendation:**
  - Add `k8sExportFailingFor` to the `mixin/README.md` thresholds table - either as its own row (`| \`k8sExportFailingFor\` | \`15m\` | Debounce for the Kubernetes export-failure alert. |`) or folded into the existing `for:` debounce-durations row alongside the other `*For` keys.

    **APPROVE** - The Kubernetes-export feature is thoroughly and accurately documented. `docs/kubernetes-export.md` covers configuration (with a field table that matches the YAML example exactly), RBAC (correctly scoped to `create`/`patch`, verified against the exporter which only calls `.Apply` and never reads objects), failure handling, and limitations; the metric names and labels in `docs/metrics.md` and `docs/kubernetes-export.md` (`puppetca_k8s_export_applies_total` with `kind,namespace,name,result`; `..._last_success_timestamp_seconds` and `..._last_error_timestamp_seconds` with `kind,namespace,name`) match `internal/k8sexport/metrics.go` verbatim, as do the PromQL examples and the `mixin/alerts.libsonnet` `PuppetCAKubernetesExportFailing` alert; and all cross-links (README→docs, docs↔mixin) resolve to existing targets. The only issue is a single LOW-severity omission of the new `k8sExportFailingFor` tunable from the mixin README's thresholds table, which does not block merge.

  <br>

  **Constraint:**
  - Documentation completeness/consistency - the mixin README table purports to enumerate the overridable thresholds, but the newly added `k8sExportFailingFor` debounce knob is omitted.

  **Description:**
  - This PR adds a new configurable `for:` debounce threshold (`k8sExportFailingFor`, default `15m`) for the `PuppetCAKubernetesExportFailing` alert. It is defined and commented in `config.libsonnet` and used in `alerts.libsonnet`, but the mixin README's "Configurable thresholds" reference table (which already lists the sibling `expiryFor`/`scrapeFor`/`readyFor`/`downFor` debounce durations) does not mention it. An operator using that table as the tunables reference would not learn the export-failure alert's debounce is overridable. All metric names/labels and other prose in the changeset are accurate and consistent with the code; this is the sole reference-table gap. Non-blocking.
  </details>

- 🧪 **New alert ships without a promtool unit test, matching the mixin's existing no-test convention** ([`mixin/alerts.libsonnet:188`](https://github.com/voxpupuli/openvox-ca/blob/2dda9b10f47337e2034b53f78fb586964831d886/mixin/alerts.libsonnet#L188))

  > `puppetca_k8s_export_last_error_timestamp_seconds{...} > puppetca_k8s_export_last_success_timestamp_seconds{...} or ... unless puppetca_k8s_export_last_success_timestamp_seconds{...}`

  💡 **Recommendation:**
  - Optionally add a `mixin/tests.yaml` (promtool unit-test format) covering the four export-alert scenarios and wire `promtool test rules` into `ci.yml`. If the maintainers deliberately keep the mixin test-free, no action is required. Do not block merge on this.

    **APPROVE** - The new `PuppetCAKubernetesExportFailing` alert references real metric series and labels defined in `internal/k8sexport/metrics.go`, and its expr/annotations are self-consistent. The repo has no alert-rule test convention at all (no `tests.yaml`, no `promtool test rules`, no Jsonnet test harness, no CI mixin step), so the new alert's lack of a unit test matches existing convention and is at most a LOW, non-blocking note rather than a regression-protection gap. No HIGH or CRITICAL test-coverage issues found in this docs-and-monitoring subsystem.

  <br>

  **Constraint:**
  - Regression protection for alert-rule logic (Tester coverage completeness).

  **Description:**
  - The `unless` arm ("never-succeeded" target) and the last-error-vs-last-success timestamp comparison are exactly the kind of boundary logic a `promtool test rules` case would lock down (target that only errored → fires; target that recovered so last_success > last_error → does not fire; target that never ran → does not fire). No such test exists, so a future edit to this expr - or a rename of the underlying `puppetca_k8s_export_*` series - could silently break firing behavior. This is not a convention violation: none of the six existing alerts is tested either, so the new alert simply follows the established (untested) pattern. It is called out only as an informational, non-blocking opportunity.
  </details>

</details>

---

bootc:
Second round addressed in 849fa13.

### Medium

- **Failure-isolation only tested with the failing target last** — added a spec that fails the *first* target (a Secret) and asserts the following ConfigMap is still created with its data, so a regress-to-early-return (`return` instead of `continue`) would now fail the suite.
- **Concurrent wiring not exercised under `-race`** — I verified the concurrent notification→export path is race-clean locally (`go test -race ./internal/k8sexport/... ./internal/ca/... ./cmd/openvox-ca/...`, 0 races). I've **not** flipped `-race` on in the magefile here: it's a repo-wide change (all `mage test:unit` runs, local included, would need cgo and run several times slower) and, as the review notes, a pre-existing infra choice outside this changeset. Happy to add it — either globally or as a dedicated race job scoped to these packages — if you'd like it in this PR or as a follow-up.

### Low

- **`signCRLLocked` "single point" comment vs. bootstrap writes** — reworded the invariant to cover writes reachable *while serving*, and added a one-line back-reference at each of the four bootstrap/import `UpdateCRL` sites (`init.go` ×2, `caImport.go` ×2) noting they skip `crlNotify` by design because they run before any consumer exists.
- **Package doc missing in-cluster-only constraint** — added a sentence to the `k8sexport` package godoc.
- **`k8sExportFailingFor` missing from mixin thresholds table** — added it to the `for:` debounce-durations row in `mixin/README.md`.
- **No `govulncheck` gate** — ran `govulncheck ./...` locally: no called vulnerabilities across the client-go surface (one advisory in a required-but-uncalled module). As with `-race`, I've left the CI wiring out of this feature PR; glad to add a `govulncheck ./...` step to `ci.yml` if you want it here.
- **Alert ships without a promtool unit test** — left as-is: the mixin has no alert-test convention (no `tests.yaml`, no `promtool test rules` in CI), and the reviewer flags it as matching that existing convention. Adding a test harness would be a separate mixin-wide change.

`go build`, `go test`, and `golangci-lint` remain green; no crypto path or dependency changed, so the FIPS build contract is unaffected.

trevor-vaughan:
@bootc This one needs a rebase with conflict resolution then we should be good to go.

bootc:
@trevor-vaughan rebased again (twice in fact...) so should be good at last...

trevor-vaughan (approved):

PR #113, opened 2026-07-12
Support wire-compatible auto-renewal (no CSR) on /certificate_renewal

bootc:
Real Puppet/OpenVox agents renew their client certificate by default (`hostcert_renewal_interval`, default 30d) by POSTing an empty body to `/certificate_renewal` and relying solely on the mTLS-presented client certificate to prove identity and key possession — they never send a CSR, and expect the SAME key reissued with a fresh serial and validity. Our existing CSR-only handler rejected every such request with a 400, so a stock agent's automatic renewal silently never worked against openvox-ca (the agent logs a warning and keeps using its current cert).

This adds `CA.AutoRenew`, matching OpenVox Server's own Clojure CA (`renew-certificate!` in `certificate_authority.clj`): it builds the new certificate directly from the already-issued certificate presented over mTLS — no CSR lookup at all, so it also covers certificates whose CSR was deleted after signing (the normal case) or that arrived via migration import with no CSR ever recorded. It carries forward the original certificate's SANs and Puppet OID extensions, and reissues under the same key.

`POST /certificate_renewal` now branches on an empty body to this path; a non-empty body keeps using the existing CSR re-key path unchanged.

The internals of `signWithDuration` were split so the "build template, sign, persist" tail (`issueLeafLocked`) is shared between the CSR path and `AutoRenew`.

**Real-agent regression test:** added Group 3b to `test/puppet/puppet-stack.sh`, which drives the actual `puppet ssl renew_cert` CLI action (the same code path a stock agent's automatic renewal takes) against a live openvox-ca instance in the compose-puppet stack, and asserts: exit 0, a new serial, the same public key, and that the renewed cert still works for a normal agent run. This is deliberately end-to-end with the real `openvox-agent` package rather than a Go-mocked TLS cert, since the bug this PR fixes is specifically a wire-format mismatch that unit tests alone would not catch.

**Scope note:** OpenVox Server's Clojure CA does not revoke the certificate being replaced during this kind of renewal — both the old and new certs (same key) stay valid until the old one naturally expires. This PR preserves that behaviour to match. A follow-up PR (voxpupuli/openvox-ca#114, based on this branch) adds revocation of the superseded certificate as a configurable, secure-by-default option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -393,8 +393,10 @@ All endpoints are served under both the bare path and `/puppet-ca/v1/<path>`, so
 
 Requires a valid CA-signed client certificate. The new certificate is issued immediately without entering the pending-CSR queue or autosign evaluation.
 
-- **CSR body (re-key):** the CSR Common Name must match the authenticated client CN — an agent can only renew its own certificate, not another's. Issues a certificate for the new key in the CSR.
-- **Empty body (wire-compatible auto-renewal):** matches the request real Puppet/OpenVox agents send by default (`hostcert_renewal_interval`, and the `puppet ssl renew_cert` CLI action). Identity and key possession come solely from the mTLS-presented client certificate; the same public key is reissued with a fresh serial and validity, carrying forward the original certificate's SANs and Puppet OID extensions unchanged. This mirrors the behaviour of OpenVox Server's own (Clojure) CA: the certificate being replaced is not revoked, so it remains valid for the same key until it naturally expires.
+- **CSR body (re-key):** the CSR Common Name must match the authenticated client CN — an agent can only renew its own certificate, not another's. Issues a certificate for the new key in the CSR. Puppet OID extensions are copied from the CSR **except** authorization-arc OIDs (`1.3.6.1.4.1.34380.1.3.*`, such as `pp_cli_auth`), which are stripped so a submitted CSR cannot request elevated privileges.
+- **Empty body (wire-compatible auto-renewal):** matches the request real Puppet/OpenVox agents send by default (`hostcert_renewal_interval`, and the `puppet ssl renew_cert` CLI action). Identity and key possession come solely from the mTLS-presented client certificate; the same public key is reissued with a fresh serial and validity, carrying forward the original certificate's SANs and Puppet OID extensions unchanged. Unlike the CSR path, this **preserves authorization-arc OIDs** (e.g. `pp_cli_auth`): they were already vetted when the presented certificate was issued, so a cert that legitimately holds them keeps them across renewal. This mirrors the behaviour of OpenVox Server's own (Clojure) CA: the certificate being replaced is not revoked, so it remains valid for the same key until it naturally expires.
+
+If the presented certificate's (or CSR's) key falls below the CA key-strength policy — for example an RSA-1024 key imported from a legacy CA — the request is rejected with `422 Unprocessable Entity` rather than renewed; the agent must re-key via the CSR path with a compliant key.
 
 ### Bulk signing
 
--- a/README.md
+++ b/README.md
@@ -389,9 +389,12 @@ All endpoints are served under both the bare path and `/puppet-ca/v1/<path>`, so
 
 | Method | Path | Description |
 |--------|------|-------------|
-| `POST` | `/certificate_renewal` | Renew an existing certificate; body: raw CSR PEM; returns new certificate PEM |
+| `POST` | `/certificate_renewal` | Renew an existing certificate; body: raw CSR PEM, or empty; returns new certificate PEM |
 
-Requires a valid CA-signed client certificate. The CSR Common Name must match the authenticated client CN — an agent can only renew its own certificate, not another's. The new certificate is issued immediately without entering the pending-CSR queue or autosign evaluation.
+Requires a valid CA-signed client certificate. The new certificate is issued immediately without entering the pending-CSR queue or autosign evaluation.
+
+- **CSR body (re-key):** the CSR Common Name must match the authenticated client CN — an agent can only renew its own certificate, not another's. Issues a certificate for the new key in the CSR.
+- **Empty body (wire-compatible auto-renewal):** matches the request real Puppet/OpenVox agents send by default (`hostcert_renewal_interval`, and the `puppet ssl renew_cert` CLI action). Identity and key possession come solely from the mTLS-presented client certificate; the same public key is reissued with a fresh serial and validity, carrying forward the original certificate's SANs and Puppet OID extensions unchanged. This mirrors the behaviour of OpenVox Server's own (Clojure) CA: the certificate being replaced is not revoked, so it remains valid for the same key until it naturally expires.
 
 ### Bulk signing
 
--- a/internal/api/api_test.go
+++ b/internal/api/api_test.go
@@ -1309,6 +1309,26 @@ var _ = Describe("API Workflow", func() {
 			Expect(renewed.SerialNumber.Cmp(clientCert.SerialNumber)).NotTo(Equal(0))
 		})
 
+		// The routing decision uses strings.TrimSpace, so a whitespace-only body
+		// (some clients send a trailing newline) must take the same auto-renewal
+		// path as a truly-empty one — not be treated as a malformed CSR. Guards
+		// against the check regressing to a bare len(body)==0.
+		It("should auto-renew when the body is whitespace only", func() {
+			req := httptest.NewRequest("POST", "/certificate_renewal", bytes.NewReader([]byte("  \n\t")))
+			req.TLS = &tls.ConnectionState{PeerCertificates: []*x509.Certificate{clientCert}}
+			rr := httptest.NewRecorder()
+			mux.ServeHTTP(rr, req)
+			Expect(rr.Code).To(Equal(http.StatusOK))
+
+			block, _ := pem.Decode(rr.Body.Bytes())
+			Expect(block).NotTo(BeNil())
+			renewed, err := x509.ParseCertificate(block.Bytes)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(renewed.PublicKey).To(Equal(clientCert.PublicKey),
+				"whitespace-only body must route to auto-renewal and reissue the same key")
+			Expect(renewed.SerialNumber.Cmp(clientCert.SerialNumber)).NotTo(Equal(0))
+		})
+
 		// A legacy-imported node may present a cert whose key predates this CA's
 		// key-strength policy. The empty-body auto-renewal path must surface that
 		// as a client-actionable 422 (re-key via CSR), not mask it behind a 500.
--- a/internal/api/api_test.go
+++ b/internal/api/api_test.go
@@ -28,6 +28,7 @@ import (
 	"encoding/asn1"
 	"encoding/json"
 	"encoding/pem"
+	"math/big"
 	"net/http"
 	"net/http/httptest"
 	"os"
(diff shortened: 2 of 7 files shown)


trevor-vaughan-ai:
## 🟢 Review Council: APPROVE

>
> Models used:
> - claude-opus-4-8[1m]

_Reviewed at commit [`73c52e2`](https://github.com/voxpupuli/openvox-ca/commit/73c52e229a5260321f955bd44f0fb9d2a16624ca) (deep effort)._

**TL;DR:** Five-persona council unanimously approves this no-CSR auto-renewal; nine advisory findings (one medium doc gap, eight low), none blocking.

**Findings:** 🔴 0 Critical, 🟠 0 High, 🟡 1 Medium, 🔵 8 Low

- 📚 **Re-key (CSR) path shares the "old certificate is not revoked" security behavior, but the caveat is documented only for auto-renewal** ([`README.md:396`](https://github.com/voxpupuli/openvox-ca/blob/73c52e229a5260321f955bd44f0fb9d2a16624ca/README.md#L396))

  > `Renew`

  💡 **Recommendation:**
  - State, in the CSR/re-key bullet (or in a shared note covering both paths), that renewal never revokes the certificate it replaces - so when re-keying due to suspected key compromise the operator MUST separately revoke the old certificate. Suggested addition after README.md:396: "The previously issued certificate is not revoked; if you are re-keying because the old key may be compromised, revoke it explicitly via `PUT /certificate_status` with `desired_state=revoked`."

    VERDICT: APPROVE

  <br>

  **Impact**: Re-keying is the exact operation an operator performs when the old key is suspected compromised (the README even motivates the CSR path as the way to replace a sub-policy/legacy key). An operator reading only the CSR bullet could reasonably assume that issuing a new-key certificate invalidates the old one. It does not: the prior certificate (with the old, possibly compromised key) remains valid until natural expiry unless the operator explicitly revokes it via `PUT /certificate_status desired_state=revoked`. The one place the non-revocation caveat is stated frames it as wire-compat parity for auto-renewal, not as an action item for the re-key/compromise scenario where it matters most. This is a security-relevant operator-facing gap, not a cosmetic one.
  </details>

</details>

<details><summary>🔵 LOW (8)</summary>

- 🧪 **Generic 500 fallback on non-policy CA failure is untested** ([`internal/api/handlers.go:938`](https://github.com/voxpupuli/openvox-ca/blob/73c52e229a5260321f955bd44f0fb9d2a16624ca/internal/api/handlers.go#L938))

  > `slog.Warn("Auto-renewal failed", "subject", cn, "error", err)`

  💡 **Recommendation:**
  - Optional - inject a CA that returns a non-policy error and assert 500. Non-blocking.

    VERDICT: APPROVE

  <br>

  **Evidence**:

  ```go
  slog.Warn("Auto-renewal failed", "subject", cn, "error", err)
  http.Error(w, "internal server error", http.StatusInternalServerError)
  ```
  grep of api_test.go renewal Context shows no `StatusInternalServerError` assertion in the renewal path - the non-`ErrKeyPolicy` failure branches of both `AutoRenew` and `Renew` have no corresponding HTTP-layer test. These are defensive catch-alls; the client-actionable paths (422/400/403) are covered, so this is LOW.

  **Impact**: A future change that mis-maps a generic CA error (e.g. accidentally returning 200 with empty body) would not be caught at the routing layer.
  </details>

- 🧪 **Documented "not revoked" Clojure semantics has no assertion** ([`internal/ca/signing.go:745`](https://github.com/voxpupuli/openvox-ca/blob/73c52e229a5260321f955bd44f0fb9d2a16624ca/internal/ca/signing.go#L745))

  > `AutoRenew`

  💡 **Recommendation:**
  - Add one assertion after a successful `AutoRenew` that the original serial does not appear in the CRL (or `IsRevoked(original)` is false), locking in the non-revocation contract.

  <br>

  **Impact**: A future change that added a revoke-and-reissue step to `AutoRenew` (breaking wire-compat with OpenVox Server's `renew-certificate!`) would pass every existing test. The load-bearing design decision is unguarded.
  </details>

- 🧪 **AutoRenew subject-validation rejection branch untested** ([`internal/ca/signing.go:753`](https://github.com/voxpupuli/openvox-ca/blob/73c52e229a5260321f955bd44f0fb9d2a16624ca/internal/ca/signing.go#L753))

  > `if err := ValidateSubject(subject); err != nil { return nil, err }`

  💡 **Recommendation:**
  - Optional - a spec presenting a cert with an invalid CN (e.g. containing `..`) asserting a non-nil error would close the branch.

    VERDICT: APPROVE

  <br>

  **Impact**: Minimal - `subject` derives from an mTLS-authenticated cert this CA itself issued, so the branch is effectively defensive/unreachable in normal flow. Low regression value.
  </details>

- 📚 **AutoRenew godoc overstates what changes on renewal ("key identifiers are refreshed")** ([`internal/ca/signing.go:743`](https://github.com/voxpupuli/openvox-ca/blob/73c52e229a5260321f955bd44f0fb9d2a16624ca/internal/ca/signing.go#L743))

  > `SubjectKeyId`

  💡 **Recommendation:**
  - Narrow the clause, e.g. "...only the serial number and validity window are freshly generated; the SubjectKeyIdentifier and AuthorityKeyIdentifier are recomputed but, for the same key and CA, remain identical. Standard key-usage and distribution-point extensions are regenerated from current CA configuration rather than copied." No code change required.

    VERDICT: APPROVE

  <br>

  **Impact**: For auto-renewal the public key and issuing CA are by definition unchanged, so both key identifiers are recomputed to *identical* values - they are not "refreshed" in any observable sense. A maintainer relying on this godoc could wrongly assume the SubjectKeyIdentifier differs between the old and renewed certificate (e.g. when reasoning about de-duplication or chain building). Separately, standard fields `KeyUsage`, `ExtKeyUsage`, `CRLDistributionPoints`, and the AIA/OCSP extension are regenerated from current CA config (369-400), not carried "verbatim" from `presentedCert`, which the "carried forward verbatim ... only ... refreshed" phrasing glosses over. This is a wording imprecision, not a behavioral defect; the security-critical claims (SANs and Puppet/auth OIDs preserved, cert not revoked) are accurate.
  </details>

- ⚙️ **slog level inconsistent across 500-returning failure modes** ([`internal/api/handlers.go:938`](https://github.com/voxpupuli/openvox-ca/blob/73c52e229a5260321f955bd44f0fb9d2a16624ca/internal/api/handlers.go#L938))

  > `body, err := io.ReadAll(io.LimitReader(r.Body, 1<<20))`

  💡 **Recommendation:**
  - Log the `CA.AutoRenew`/`CA.Renew` failures that produce a 500 at `slog.Error` to match the body-read 500 and standard 5xx-alerting conventions; keep the `ErrKeyPolicy` 422 branches at `slog.Warn` (correctly client-caused). Optionally unify the CN field key (`"client"` vs `"subject"`) for consistent log queries.

    VERDICT: APPROVE

  <br>

  **Evidence**:

  ```go
  body, err := io.ReadAll(io.LimitReader(r.Body, 1<<20))
  if err != nil {
      slog.Error("read renewal body failed", "client", cn, "error", err)   // 500 at Error
      ...
  }
  ...
      slog.Warn("Auto-renewal failed", "subject", cn, "error", err)        // 500 at Warn
      http.Error(w, "internal server error", http.StatusInternalServerError)
  ...
      slog.Warn("Renewal failed", "subject", cn, "error", err)             // 500 at Warn
      http.Error(w, "internal server error", http.StatusInternalServerError)
  ```

  **Impact**: Three distinct failures all return HTTP 500, but the body-read failure logs at Error while the `CA.AutoRenew` / `CA.Renew` internal failures log at Warn. Operators whose alerting keys on Error level (a common convention for 5xx) will page on a body-read failure but silently miss a genuine internal CA failure (signing/storage infra), which is the more serious of the two. The key-field naming also differs (`"client"` vs `"subject"`) for the same CN value, adding minor friction to log correlation. The Warn-for-500 pattern is pre-existing; this PR extends it into the new AutoRenew branch rather than introducing it.
  </details>

- 🧪 **Both 422 tests assert status code only, not the distinct branch messages** ([`internal/api/api_test.go:1356`](https://github.com/voxpupuli/openvox-ca/blob/73c52e229a5260321f955bd44f0fb9d2a16624ca/internal/api/api_test.go#L1356))

  > `// empty-body key-policy test (1352-1356)`

  💡 **Recommendation:**
  - Add `Expect(rr.Body.String()).To(ContainSubstring("new CSR"))` (empty-body) and `To(ContainSubstring("CSR key"))` (CSR) to pin each branch to its own message.

  <br>

  **Evidence**:

  ```go
  // empty-body key-policy test (1352-1356)
  req := httptest.NewRequest("POST", "/certificate_renewal", bytes.NewReader(nil))
  ...
  Expect(rr.Code).To(Equal(http.StatusUnprocessableEntity))
  ```

  ```go
  // CSR key-policy test (1370-1374)
  req := httptest.NewRequest("POST", "/certificate_renewal", bytes.NewReader(weakCSR))
  ...
  Expect(rr.Code).To(Equal(http.StatusUnprocessableEntity))
  ```
  The handler emits two different client-actionable messages for these branches - `"certificate key does not meet policy; renew with a new CSR"` (handlers.go:935) vs `"CSR key does not meet policy"` (handlers.go:965). Both tests check only the 422 code, so swapping the two error strings would go undetected. The status code is the hard contract and it is verified; the message text is informational, hence LOW.

  **Impact**: A regression that crosses the two 422 messages (misdirecting an agent on how to recover) would not be caught.
  </details>

- 🧭 **parseCertPEM and issue helper closures duplicated across sibling Describe blocks** ([`internal/ca/renew_test.go:184`](https://github.com/voxpupuli/openvox-ca/blob/73c52e229a5260321f955bd44f0fb9d2a16624ca/internal/ca/renew_test.go#L184))

  > `CA AutoRenew`

  💡 **Recommendation:**
  - Optional - hoist `parseCertPEM` (and `issue` if desired) to package-level `ca_test` helpers so both Describe blocks share one definition, or leave as-is to preserve the current closure-scoped convention. No change required for correctness.

    VERDICT: APPROVE

  <br>

  ```go
  parseCertPEM := func(certPEM []byte) *x509.Certificate { ... }   // == lines 51-57
  issue := func(subject string) *x509.Certificate { ... }          // == lines 61-68
  ```
  `issue` is invoked only once in the new block (line 276); the sibling SANs spec (322-327) open-codes the same SaveRequest+Sign+parse rather than calling it.

  **Impact**: Minor zero-waste: two identical helper bodies now live twice in one file and can drift during maintenance. Non-blocking. This mirrors the file's pre-existing choice of per-Describe closure-scoped helpers, so it is consistent with the established local pattern rather than a departure from it.
  </details>

- 🧪 **Group 3b "new serial" check can false-pass when the pre-renewal serial read returns empty** ([`test/puppet/puppet-stack.sh:541`](https://github.com/voxpupuli/openvox-ca/blob/73c52e229a5260321f955bd44f0fb9d2a16624ca/test/puppet/puppet-stack.sh#L541))

  > `_serial_before=$(exec_client openssl x509 -in "$_client_cert_path" -noout -serial 2>/dev/null) || true`

  💡 **Recommendation:**
  - Assert `-n "$_serial_before"` (and `-n "$_pubkey_before"`) right after capture, or fail explicitly if the pre-renewal read returns empty, so a broken baseline read is a test failure rather than a silent pass on the serial-changed check.

    The CA-layer and API-layer specs are thorough, assertion-deep, and well-isolated, with every new `AutoRenew`/handler branch covered including the auth-OID asymmetry and both key-policy 422 paths. The single finding is LOW and non-blocking.

    VERDICT: APPROVE

  <br>

  **Evidence**:

  ```
  _serial_before=$(exec_client openssl x509 -in "$_client_cert_path" -noout -serial 2>/dev/null) || true
  ...
  [[ -n "$_serial_after" && "$_serial_after" != "$_serial_before" ]] \
      && pass "renewed cert carries a new serial" \
      || fail "renewed cert carries a new serial" "before=$_serial_before after=$_serial_after"
  ```

  **Impact**: If the pre-renewal `openssl` read fails, `_serial_before` is empty (the `|| true`, plus absence of global `set -e`, suppresses any signal). The assertion then reduces to "`_serial_after` is non-empty and differs from empty", which is trivially true - so it reports "renewed cert carries a new serial" without ever having compared against the original serial. The adjacent pubkey check at :545 uses `==` and fails safe on an empty `before`, so only the serial-changed assertion carries this masking. Low likelihood (the cert was just used for a successful agent run) and corroborated by the pubkey/Downloaded/catalog assertions, so this is a robustness gap in a smoke test, not a coverage hole.
  </details>

</details>

---

trevor-vaughan (approved):

PR #114, opened 2026-07-12
Revoke the certificate replaced by renewal, configurable for auto-renew

bootc:
**Depends on / stacked on #113** (base branch of this PR is `feature/auto-renew-wire-compat`, not `main`). Merge that one first; this one's diff will shrink to just the commit below once it lands.

Ensures only one certificate per subject is ever valid: once a renewal's replacement certificate is safely signed and stored, the certificate it replaces is revoked.

- **CSR-based `Renew`** (a genuine re-key) always revokes the replaced certificate — it's a different key, so the old one must not remain a usable credential.
- **`AutoRenew`** (the empty-body, same-key path from #113) does this too, but gated behind a new `RevokeOnAutoRenew` option, **default `true`**. OpenVox Server's own Clojure CA (`renew-certificate!`) does not revoke on this path — both the old and new certs (same key) stay valid until the old one naturally expires — so operators who need to match that exactly can set `revoke_on_auto_renew: false` / `PUPPET_CA_REVOKE_ON_AUTO_RENEW=false`.

`revokeLocked` is split into itself (subject → latest-inventory-serial lookup) plus `revokeSerialLocked` (revoke an exact serial), since both `Renew` and `AutoRenew` need to capture the old serial *before* signing — once the new cert's inventory row is appended, a subject-based lookup would resolve to the new serial rather than the one being retired.

**Best-effort revocation + observability:** revoking the superseded certificate is best-effort — a failure to amend the CRL never fails the renewal the agent is waiting on (the renewed certificate is still returned). So that this can't silently hide a broken CRL, a new counter `puppetca_crl_update_failures_total` increments on *any* failure to amend the CRL: a revocation that can't be recorded, or a CRL that can't be re-signed or written. It's incremented at the two shared choke points (`revokeSerialLocked` and `signCRLLocked`), so it covers every path — renewal, admin and clean revocation, cleanup pruning, reissue and refresh — not just this feature. The monitoring mixin gains a `PuppetCACRLUpdateFailing` warning alert; see `docs/metrics.md`.

**CRL growth:** because the default now revokes on every auto-renewal, the CRL gains one entry per renewal until those certificates expire. The `revoke_on_auto_renew` docs now point operators at `enable_expired_cert_cleanup` (which prunes expired entries and is off by default) and `puppetca_crl_revoked_certificates` to keep CRL size bounded on busy CAs.

**Why the default stays `true`:** I validated this end-to-end against a real agent in `test/puppet/puppet-stack.sh` (Group 3b, extended): with `revoke_on_auto_renew: true` (the default), a real agent's `puppet ssl renew_cert` still succeeds, the pre-renewal serial lands in the CRL, and a subsequent normal agent run using the renewed cert still passes. Since the real-agent test passes cleanly with revocation enabled, the more secure default (revoke immediately) is kept, per the plan of only flipping the default if that test had failed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -393,8 +393,10 @@ All endpoints are served under both the bare path and `/puppet-ca/v1/<path>`, so
 
 Requires a valid CA-signed client certificate. The new certificate is issued immediately without entering the pending-CSR queue or autosign evaluation.
 
-- **CSR body (re-key):** the CSR Common Name must match the authenticated client CN — an agent can only renew its own certificate, not another's. Issues a certificate for the new key in the CSR.
-- **Empty body (wire-compatible auto-renewal):** matches the request real Puppet/OpenVox agents send by default (`hostcert_renewal_interval`, and the `puppet ssl renew_cert` CLI action). Identity and key possession come solely from the mTLS-presented client certificate; the same public key is reissued with a fresh serial and validity, carrying forward the original certificate's SANs and Puppet OID extensions unchanged. This mirrors the behaviour of OpenVox Server's own (Clojure) CA: the certificate being replaced is not revoked, so it remains valid for the same key until it naturally expires.
+- **CSR body (re-key):** the CSR Common Name must match the authenticated client CN — an agent can only renew its own certificate, not another's. Issues a certificate for the new key in the CSR. Puppet OID extensions are copied from the CSR **except** authorization-arc OIDs (`1.3.6.1.4.1.34380.1.3.*`, such as `pp_cli_auth`), which are stripped so a submitted CSR cannot request elevated privileges.
+- **Empty body (wire-compatible auto-renewal):** matches the request real Puppet/OpenVox agents send by default (`hostcert_renewal_interval`, and the `puppet ssl renew_cert` CLI action). Identity and key possession come solely from the mTLS-presented client certificate; the same public key is reissued with a fresh serial and validity, carrying forward the original certificate's SANs and Puppet OID extensions unchanged. Unlike the CSR path, this **preserves authorization-arc OIDs** (e.g. `pp_cli_auth`): they were already vetted when the presented certificate was issued, so a cert that legitimately holds them keeps them across renewal. This mirrors the behaviour of OpenVox Server's own (Clojure) CA: the certificate being replaced is not revoked, so it remains valid for the same key until it naturally expires.
+
+If the presented certificate's (or CSR's) key falls below the CA key-strength policy — for example an RSA-1024 key imported from a legacy CA — the request is rejected with `422 Unprocessable Entity` rather than renewed; the agent must re-key via the CSR path with a compliant key.
 
 ### Bulk signing
 
--- a/README.md
+++ b/README.md
@@ -163,6 +163,8 @@ encrypt_ca_key: false           # encrypt the CA private key (AES-256-GCM + Argo
 ca_key_passphrase_file: ""      # path to passphrase file; auto-generated if omitted
 # Date/time format in JSON responses.
 puppet_datetime_format: false   # use Puppet CA style "2006-01-02T15:04:05MST" instead of RFC 3339
+# Certificate auto-renewal (empty-body POST /certificate_renewal).
+revoke_on_auto_renew: true      # false matches OpenVox Server's Clojure CA (no revocation on auto-renewal)
 ```
 
 **Environment variables (mirrors CLI flags):**
@@ -237,6 +239,7 @@ The CA key passphrase can also be provided via `PUPPET_CA_KEY_PASSPHRASE` (env v
 | `etcd_tls_cert_file` | `PUPPET_CA_ETCD_TLS_CERT_FILE` |
 | `etcd_tls_key_file` | `PUPPET_CA_ETCD_TLS_KEY_FILE` |
 | `puppet_datetime_format` | `PUPPET_CA_PUPPET_DATETIME_FORMAT` |
+| `revoke_on_auto_renew` | `PUPPET_CA_REVOKE_ON_AUTO_RENEW` |
 
 > **Note:** `--daemon` is intentionally excluded from config file and environment
 > variable support because `PUPPET_CA_DAEMON` is used internally as the daemon fork
@@ -391,12 +394,14 @@ All endpoints are served under both the bare path and `/puppet-ca/v1/<path>`, so
 |--------|------|-------------|
 | `POST` | `/certificate_renewal` | Renew an existing certificate; body: raw CSR PEM, or empty; returns new certificate PEM |
 
-Requires a valid CA-signed client certificate. The new certificate is issued immediately without entering the pending-CSR queue or autosign evaluation.
+Requires a valid CA-signed client certificate. The new certificate is issued immediately without entering the pending-CSR queue or autosign evaluation, and the certificate it replaces is revoked once the new one is safely stored (see `revoke_on_auto_renew` below for the auto-renewal case).
 
 - **CSR body (re-key):** the CSR Common Name must match the authenticated client CN — an agent can only renew its own certificate, not another's. Issues a certificate for the new key in the CSR. Puppet OID extensions are copied from the CSR **except** authorization-arc OIDs (`1.3.6.1.4.1.34380.1.3.*`, such as `pp_cli_auth`), which are stripped so a submitted CSR cannot request elevated privileges.
-- **Empty body (wire-compatible auto-renewal):** matches the request real Puppet/OpenVox agents send by default (`hostcert_renewal_interval`, and the `puppet ssl renew_cert` CLI action). Identity and key possession come solely from the mTLS-presented client certificate; the same public key is reissued with a fresh serial and validity, carrying forward the original certificate's SANs and Puppet OID extensions unchanged. Unlike the CSR path, this **preserves authorization-arc OIDs** (e.g. `pp_cli_auth`): they were already vetted when the presented certificate was issued, so a cert that legitimately holds them keeps them across renewal. This mirrors the behaviour of OpenVox Server's own (Clojure) CA: the certificate being replaced is not revoked, so it remains valid for the same key until it naturally expires.
+- **Empty body (wire-compatible auto-renewal):** matches the request real Puppet/OpenVox agents send by default (`hostcert_renewal_interval`, and the `puppet ssl renew_cert` CLI action). Identity and key possession come solely from the mTLS-presented client certificate; the same public key is reissued with a fresh serial and validity, carrying forward the original certificate's SANs and Puppet OID extensions unchanged. Unlike the CSR path, this **preserves authorization-arc OIDs** (e.g. `pp_cli_auth`): they were already vetted when the presented certificate was issued, so a cert that legitimately holds them keeps them across renewal.
 
 If the presented certificate's (or CSR's) key falls below the CA key-strength policy — for example an RSA-1024 key imported from a legacy CA — the request is rejected with `422 Unprocessable Entity` rather than renewed; the agent must re-key via the CSR path with a compliant key.
+
+`revoke_on_auto_renew` (env `PUPPET_CA_REVOKE_ON_AUTO_RENEW`, default `true`) controls whether the certificate replaced by an auto-renewal (empty body) is revoked. The default keeps only the newest serial per subject valid. Set to `false` to match OpenVox Server's own (Clojure) CA exactly, which leaves the replaced certificate valid — for the same key — until it naturally expires. This setting has no effect on the CSR-body (re-key) path, which always revokes the certificate it replaces.
 
 ### Bulk signing
 
--- a/README.md
+++ b/README.md
@@ -403,6 +403,8 @@ If the presented certificate's (or CSR's) key falls below the CA key-strength po
 
 `revoke_on_auto_renew` (env `PUPPET_CA_REVOKE_ON_AUTO_RENEW`, default `true`) controls whether the certificate replaced by an auto-renewal (empty body) is revoked. The default keeps only the newest serial per subject valid. Set to `false` to match OpenVox Server's own (Clojure) CA exactly, which leaves the replaced certificate valid — for the same key — until it naturally expires. This setting has no effect on the CSR-body (re-key) path, which always revokes the certificate it replaces.
 
+> **CRL growth:** with the default `true`, every auto-renewal appends the retired serial to the CRL, and the entry stays there until the certificate expires. Entries are only pruned by the expired-certificate cleanup job, which is off by default — enable `enable_expired_cert_cleanup` to bound CRL size on busy CAs, and watch `puppetca_crl_revoked_certificates` to keep an eye on it. Revocation is best-effort (a failure never fails the renewal); the `puppetca_crl_update_failures_total` metric counts any failure to amend the CRL, including a superseded certificate that could not be revoked.
+
 ### Bulk signing
 
 | Method | Path | Description |
--- a/README.md
+++ b/README.md
@@ -389,9 +389,12 @@ All endpoints are served under both the bare path and `/puppet-ca/v1/<path>`, so
 
 | Method | Path | Description |
 |--------|------|-------------|
-| `POST` | `/certificate_renewal` | Renew an existing certificate; body: raw CSR PEM; returns new certificate PEM |
+| `POST` | `/certificate_renewal` | Renew an existing certificate; body: raw CSR PEM, or empty; returns new certificate PEM |
 
-Requires a valid CA-signed client certificate. The CSR Common Name must match the authenticated client CN — an agent can only renew its own certificate, not another's. The new certificate is issued immediately without entering the pending-CSR queue or autosign evaluation.
+Requires a valid CA-signed client certificate. The new certificate is issued immediately without entering the pending-CSR queue or autosign evaluation.
+
+- **CSR body (re-key):** the CSR Common Name must match the authenticated client CN — an agent can only renew its own certificate, not another's. Issues a certificate for the new key in the CSR.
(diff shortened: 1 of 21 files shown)


trevor-vaughan-ai:

>
> Models used:
> - claude-opus-4-8 (Capable tier)
> - claude-opus-4-8 (Standard tier)

_Reviewed at commit [`a697b0a`](https://github.com/voxpupuli/openvox-ca/commit/a697b0ae78383d78c7f5fe3ea0830dcc08545dcb) (deep effort)._

**TL;DR:** Approve with advisories: the empty-body auto-renew feature is correct and secure; all 11 findings are non-blocking test-coverage or doc nits.

**Findings:** 🔴 0 Critical, 🟠 0 High, 🟡 5 Medium, 🔵 6 Low

<details><summary>🟡 MEDIUM (5)</summary>

- 🧪 **Best-effort revoke-failure branch in Renew/AutoRenew is untested** ([`internal/ca/signing.go:740`](https://github.com/voxpupuli/openvox-ca/blob/a697b0ae78383d78c7f5fe3ea0830dcc08545dcb/internal/ca/signing.go#L740))

  > `\t\t\tslog.Warn("Renew: failed to revoke replaced certificate", "subject", subject, "serial", oldSerial, "error", err)`

  💡 **Recommendation:**
  - Add one regression spec per path to `renew_test.go` that induces a revoke failure using the existing real-storage fixtures (no mock needed) - e.g. overwrite the seeded CRL with non-PEM/garbage bytes after `BeforeEach`, then call `Renew`/`AutoRenew` and assert the new cert is returned with no error (optionally that the old serial is *not* revoked). This anchors the best-effort contract that renewal succeeds despite a revoke failure. Non-blocking: the primary happy, negative, and security paths of both renewal flows are already thoroughly covered.

  <br>

  **Description:**
  - This `slog.Warn` is the failure arm of the best-effort revoke that `Renew` performs after signing the replacement cert (the enclosing `if err := c.Storage.WithLock(ctx, lockNameCRL, ...); err != nil {` block spans signing.go:735-741; the identical pattern for `AutoRenew` is at signing.go:827). Both methods document a deliberate contract - a revoke failure must **not** undo the renewal ("Best-effort... a failure here shouldn't undo the renewal the caller is waiting on"). Every revocation test (renew_test.go:151 for `Renew`, renew_test.go:382 for `AutoRenew`) asserts only the success side: that the old serial ends up revoked. No test forces `revokeSerialLocked` to fail and then asserts that `Renew`/`AutoRenew` still returns a valid new cert with `err == nil`. The branch is entirely uncovered: a refactor that let the revoke error propagate would break agent renewal whenever the CRL is transiently unreadable/corrupt, and the suite would stay green.
  </details>

- 🧪 **New `ErrKeyPolicy` -> 422 handler branches have no HTTP-boundary test** ([`internal/api/handlers.go:869`](https://github.com/voxpupuli/openvox-ca/blob/a697b0ae78383d78c7f5fe3ea0830dcc08545dcb/internal/api/handlers.go#L869))

  > `if errors.Is(err, ca.ErrKeyPolicy) {`

  <br>

  **File**: internal/api/handlers.go:869 (and the twin at handlers.go:899)

  **Evidence**:

  ```go
  if errors.Is(err, ca.ErrKeyPolicy) {
      slog.Warn("Auto-renewal rejected: key policy", "subject", cn, "error", err)
      http.Error(w, "certificate key does not meet policy; renew with a new CSR", http.StatusUnprocessableEntity)
      return
  }
  ```
  These two branches are the primary net-new logic of the handler diff: they deliberately translate a key-policy rejection into a client-actionable `422` instead of masking it as a `500`. No test exercises either branch through the mux. `grep -rn "ErrKeyPolicy\|StatusUnprocessableEntity\|422" internal/ --include=*_test.go` returns only CA-layer unit tests (internal/ca/renew_test.go:357,379) - those confirm `AutoRenew`/`Renew` *return* `ErrKeyPolicy`, but nothing verifies the handler maps it to `422` rather than `500`. The renewal context in api_test.go (lines 1216-1317) covers 200 (CSR + empty-body), 403 (no cert), 403 (CN mismatch), and 400 (garbage CSR), but has no `422` case.

  Failure scenario: a refactor that drops or reorders the `errors.Is(err, ca.ErrKeyPolicy)` guard makes both paths fall through to `http.StatusInternalServerError`. Agents presenting a sub-policy key would then receive `500` (transient/retryable) instead of `422` (re-key via CSR), and the whole suite stays green.

  Fix: add two renewal specs - (1) empty body with a client cert whose key is below policy, asserting `rr.Code == http.StatusUnprocessableEntity`; (2) a CSR body carrying a sub-policy key, same assertion. Reuse the sub-policy key material already built in internal/ca/renew_test.go:357.
  </details>

- 🧪 **`RevokeOnAutoRenew` config default and `PUPPET_CA_REVOKE_ON_AUTO_RENEW` env parsing are untested** ([`cmd/openvox-ca/config.go:143`](https://github.com/voxpupuli/openvox-ca/blob/a697b0ae78383d78c7f5fe3ea0830dcc08545dcb/cmd/openvox-ca/config.go#L143))

  > `RevokeOnAutoRenew: true, // only the newest serial per subject should be valid`

  <br>

  **Evidence**:

  ```go
  RevokeOnAutoRenew: true, // only the newest serial per subject should be valid
  ...
  if v := os.Getenv("PUPPET_CA_REVOKE_ON_AUTO_RENEW"); v != "" {
      if b, err := strconv.ParseBool(v); err == nil {
          cfg.RevokeOnAutoRenew = b
      }
  }
  ```
  `grep -n "RevokeOnAutoRenew\|REVOKE_ON_AUTO_RENEW" cmd/openvox-ca/config_test.go` returns nothing. The `loadServerConfig built-in defaults` spec (config_test.go:156-168) asserts `Host/Port/CADir/NoTLSRequired/Verbosity/CAPathLength` but not `RevokeOnAutoRenew`, so the `true` default literal is unguarded here. The `applyServerEnv each variable` table (config_test.go:405-480) has entries for every other env var but none for `PUPPET_CA_REVOKE_ON_AUTO_RENEW`, so the true/false/invalid parse paths are unexercised - note the tested `ENABLE_EXPIRED_CERT_CLEANUP` (config_test.go:256) proves the `strconv.ParseBool` pattern is normally covered. The var is also absent from `serverEnvVars` (config_test.go:30-64), so `clearServerEnv()` does not isolate it.

  Partial mitigation (why MEDIUM, not HIGH): the *behavioral effect* is covered outside this subsystem - the default-true revocation is asserted end-to-end at api_test.go:1312-1315, and the false path (cert stays valid) at internal/ca/renew_test.go:397-408. The gap is the config-layer wiring for a new security-relevant toggle: a regression flipping the default literal to `false`, or a broken env-parse assignment, would silently leave superseded certificates valid with no config test to catch it.

  Fix: add a defaults assertion `Expect(cfg.RevokeOnAutoRenew).To(BeTrue())` in the defaults spec, add a `DescribeTable` entry for `PUPPET_CA_REVOKE_ON_AUTO_RENEW=false` (and confirm invalid values are ignored), and add `"PUPPET_CA_REVOKE_ON_AUTO_RENEW"` to `serverEnvVars`.
  </details>

- 🧪 **AutoRenew specs claim "fresh validity" but never assert the renewed expiry is extended** ([`internal/ca/renew_test.go:285`](https://github.com/voxpupuli/openvox-ca/blob/a697b0ae78383d78c7f5fe3ea0830dcc08545dcb/internal/ca/renew_test.go#L285))

  > `\t\t\t"auto-renewed cert must carry a different serial than the original")`

  💡 **Recommendation:**
  - In the happy-path spec (renew_test.go:274-291), add an assertion such as `Expect(renewed.NotAfter).To(BeTemporally(">", original.NotAfter), "auto-renewal must extend validity, not just mint a new serial")`. Non-blocking: the existing assertions are otherwise deep and correct; this closes the one claimed-but-unchecked dimension.

  <br>

  **Description:**
  - The Describe block comment (renew_test.go:180-185) names a "fresh serial and validity" as core properties of AutoRenew, and extending the expiry window is the entire purpose of a renewal. The happy-path spec asserts the key is preserved (line 282) and the serial differs (line 284), but no spec in the block asserts the renewed certificate's validity is extended. A grep for `NotAfter`/`NotBefore` in renew_test.go finds them only as template *inputs* when minting seed certs (lines 227-228, 322-323, 369-370), never inside an assertion. A regression where `issueLeafLocked` reissued the cert with the same or a shorter validity window - leaving the agent's cert to expire on its original schedule despite a "successful" renewal - would pass every one of these specs.
  </details>

- ⚙️ **Empty pre-renewal serial makes CRL revocation assertion unfalsifiable** ([`test/puppet/puppet-stack.sh:551`](https://github.com/voxpupuli/openvox-ca/blob/a697b0ae78383d78c7f5fe3ea0830dcc08545dcb/test/puppet/puppet-stack.sh#L551))

  > `_old_serial_hex=${_serial_before#serial=}`

  <br>

  **File**: test/puppet/puppet-stack.sh:551-557

  **Evidence**:

  ```bash
  _old_serial_hex=${_serial_before#serial=}
  _crl_after_renew=$(curl -sfk "${_CA[@]}" \
      "${CA_HOST_URL}/puppet-ca/v1/certificate_revocation_list/ca" 2>/dev/null \
      | openssl crl -text -noout 2>/dev/null) || true
  grep -qF "$_old_serial_hex" <<< "$_crl_after_renew" \
      && pass "pre-renewal serial is revoked by default after auto-renewal" \
      || fail "pre-renewal serial is revoked by default after auto-renewal" "serial $_old_serial_hex not found in CRL"
  ```
  `_serial_before` is captured with `... || true` (line 518), so if the pre-renewal `openssl x509 -serial` call yields nothing, `_old_serial_hex` becomes the empty string, and `grep -qF ""` matches any non-empty input - the assertion passes unconditionally rather than testing revocation. The two sibling assertions do not share this hole: the new-serial check (541) and pubkey check (545) both gate on `-n "$_serial_after"` / `-n "$_pubkey_after"` first. This is narrow (the client cert provably exists after Group 3, so the empty case is unlikely), which is why it is LOW rather than higher, but the CRL assertion should likewise refuse to pass on an empty serial.

  Fix: guard the pattern before matching, e.g.

  ```bash
  [[ -n "$_old_serial_hex" ]] && grep -qF "$_old_serial_hex" <<< "$_crl_after_renew" \
      && pass ... \
      || fail ... "serial '$_old_serial_hex' not found in CRL"
  ```
  so an unextractable pre-renewal serial fails the suite instead of silently passing.
  </details>

</details>

<details><summary>🔵 LOW (6)</summary>

- 📚 **AutoRenew's auth-OID carry-forward has no marker at the point of divergence** ([`internal/ca/signing.go:788`](https://github.com/voxpupuli/openvox-ca/blob/a697b0ae78383d78c7f5fe3ea0830dcc08545dcb/internal/ca/signing.go#L788))

  > `\t\tif IsPuppetOID(ext.Id) {`

  💡 **Recommendation:**
  - Add a one-line comment at the loop pointing back to the deliberate divergence, e.g. `// Unlike signWithDuration's CSR path (!IsAuthOID), AutoRenew keeps every Puppet OID - these were vetted when presentedCert was issued. Do NOT add an IsAuthOID filter here; see this method's godoc. NIST 800-53: AC-6, CM-7.` Code behavior already matches the documented intent, so this is a non-blocking inline-comment gap.

  <br>

  **Description:**
  - This loop (signing.go:786-791) copies every Puppet OID from `presentedCert` onto the renewed cert. Its near-twin in `signWithDuration` (signing.go:281, `if IsPuppetOID(ext.Id) && !IsAuthOID(ext.Id)`) additionally strips authorization-arc OIDs and carries a four-line `SECURITY:` comment explaining why. Grep confirms the asymmetry: line 281 strips auth OIDs, line 788 keeps them, line 612 warns on them. The security-relevant divergence (the CSR path strips auth OIDs to prevent privilege escalation; AutoRenew intentionally preserves them because they were vetted at original issuance) is documented only in the `AutoRenew` godoc ~30 lines above (signing.go:753-757). The loop itself, the exact place a maintainer edits, has no local marker, so a maintainer "harmonizing" the two loops could silently add an `!IsAuthOID` filter and break intended wire-compat carry-forward of `pp_cli_auth`.
  </details>

- 🧪 **Empty-body auto-renew routing only tested with a truly-empty body, not whitespace-only** ([`internal/api/handlers.go:854`](https://github.com/voxpupuli/openvox-ca/blob/a697b0ae78383d78c7f5fe3ea0830dcc08545dcb/internal/api/handlers.go#L854))

  > `if strings.TrimSpace(string(body)) == "" {`

  <br>

  **Evidence**:

  ```go
  if strings.TrimSpace(string(body)) == "" {
  ```
  The routing decision uses `strings.TrimSpace`, but the only empty-body spec sends `bytes.NewReader(nil)` (api_test.go:1296), which would also pass a plain `len(body)==0` check. Nothing exercises a whitespace-only body (e.g. `"\n"` / `"  "`), so the `TrimSpace` - as opposed to a bare emptiness check - is not locked down. A regression replacing `TrimSpace(...)==""` with `len(body)==0` would route a whitespace body into `parseCSR` and return `400` instead of auto-renewing, undetected.

  Fix: add a spec posting a whitespace-only body with a client cert and assert `200` + same-key reissue, mirroring api_test.go:1295.
  </details>

- 🧭 **Direct leaf-minting scaffold duplicated across three specs (zero-waste)** ([`internal/ca/renew_test.go:218`](https://github.com/voxpupuli/openvox-ca/blob/a697b0ae78383d78c7f5fe3ea0830dcc08545dcb/internal/ca/renew_test.go#L218))

  > `\tseedCertWithoutCSR := func(subject string) *x509.Certificate {`

  💡 **Recommendation:**
  - Extract a small block-scoped helper (e.g. `mintLeaf(keyBits int, template *x509.Certificate) *x509.Certificate`) to collapse the three copies while keeping each spec's distinguishing template inline. The file already establishes a precedent for shared package-level test helpers (`buildCSR`), so a helper here would be consistent, not a new pattern.

  <br>

  **Description:**
  - The same generate-key / random-128-bit-serial / build-template / `CreateCertificate` sequence is repeated three times: `seedCertWithoutCSR` (first line at 219), the non-DNS SANs spec (`rsa.GenerateKey` at line 312), and the sub-policy-key spec (`rsa.GenerateKey` at line 361). The per-case variation is only the key size and the template fields. This is test-only fixture scaffolding with meaningful per-case differences and no functional impact, so it does not gate merge - noted for maintenance-waste reduction only.
  </details>

- 📚 **Spec description says "SANs" but assertion only checks DNSNames** ([`internal/ca/renew_test.go:293`](https://github.com/voxpupuli/openvox-ca/blob/a697b0ae78383d78c7f5fe3ea0830dcc08545dcb/internal/ca/renew_test.go#L293))

  > `It("carries the original certificate's SANs forward unchanged", func() {`

  <br>

  **Evidence**:

  ```go
  It("carries the original certificate's SANs forward unchanged", func() {
      ...
      Expect(renewed.DNSNames).To(Equal(original.DNSNames))
  })
  ```
  The description promises SANs (all types) are carried forward, but the body issues via `buildCSR` (which sets only a CommonName, no explicit SANs) and asserts only `renewed.DNSNames`. The claim is substantively defensible - CSR-issued openvox-ca certs carry only DNS SANs, and the immediately following spec (line 308) covers IP/email/URI - so this is not misleading in effect. Optional tightening for reader accuracy: either narrow the string to "...the original certificate's DNS SANs forward unchanged", or add a one-line comment noting that a CSR-issued cert carries only DNS SANs and the non-DNS case is covered by the next spec. No functional impact.
  </details>

- 🧪 **Auto-renewed agent-run assertion is shallower than the Group 3 equivalent** ([`test/puppet/puppet-stack.sh:561`](https://github.com/voxpupuli/openvox-ca/blob/a697b0ae78383d78c7f5fe3ea0830dcc08545dcb/test/puppet/puppet-stack.sh#L561))

  > `AGENT_RENEW_OUT=$(run_agent) || AGENT_RENEW_EXIT=$?`

  <br>

  **Evidence**:

  ```bash
  AGENT_RENEW_OUT=$(run_agent) || AGENT_RENEW_EXIT=$?
  [[ "$AGENT_RENEW_EXIT" =~ ^(0|2)$ ]] \
      && pass "Agent run with auto-renewed cert exits 0 or 2" \
      || fail "Agent run with auto-renewed cert exits 0 or 2" ...
  ```
  The claim is "the renewed cert must still work for a normal agent run," but unlike Group 3 (lines 478-491, which also assert `Applied catalog`, absence of `certificate revoked`/`SSL error`, and the smoke-test notify) this run only checks the exit code. A revoked-cert TLS rejection would surface as exit 1 (so 0/2 does discriminate), but `AGENT_RENEW_OUT` is captured and never inspected - the "no `certificate revoked` in output" and "catalog applied" properties that would most directly prove the renewed cert is trusted go unasserted. Optional: add a `grep -qi "Applied catalog"` / no-SSL-error check against `AGENT_RENEW_OUT` to match Group 3's depth.

  No other issues: the integration-compose.sh cert switches preserve both the invalid-CSR-400 and bare-path-200 cases, and the serial/pubkey guards are not tautological.
  </details>

- 📚 **Pubkey fail-detail string can misreport an extraction failure as a key change** ([`test/puppet/puppet-stack.sh:545`](https://github.com/voxpupuli/openvox-ca/blob/a697b0ae78383d78c7f5fe3ea0830dcc08545dcb/test/puppet/puppet-stack.sh#L545))

  > `[[ -n "$_pubkey_after" && "$_pubkey_after" == "$_pubkey_before" ]] \\`

  <br>

  **Evidence**:

  ```
  [[ -n "$_pubkey_after" && "$_pubkey_after" == "$_pubkey_before" ]] \
      && pass "renewed cert keeps the same public key (no CSR, no re-key)" \
      || fail "renewed cert keeps the same public key (no CSR, no re-key)" "public key changed across renewal"
  ```
  The condition also fails when `_pubkey_after` is empty (e.g., the `exec_client openssl ... -pubkey` on line 539 errored, which is swallowed by `|| true`). In that case the hardcoded detail "public key changed across renewal" is inaccurate - the key did not change, extraction failed. The sibling serial check on line 543 avoids this by emitting the actual `before=/after=` values. Consider mirroring that here (e.g., a detail noting whether `_pubkey_after` was empty) so a failing run points at the real cause. Cosmetic; does not affect the passing path.
  </details>

</details>

---

trevor-vaughan (approved):

bootc:
Thanks @trevor-vaughan! If you're happy with this one, though, would you mind approving #113 that this one includes too, please?

trevor-vaughan:
That one hiccuped on my reviewer pass last night. Running a fresh pass and will check when it's done!

PR #115, opened 2026-07-12
Add import-cert: register externally-issued certificates in the inventory

bootc:
Certificates issued outside this CA's normal `Sign`/`Generate` flow (e.g. migrated from a legacy CA sharing this CA's key) previously had no way to enter the inventory, so they were invisible to listing, expiry cleanup, and by-subject revocation, even though they were genuinely signed by this CA.

This adds `PUT /certificate/{subject}` (admin-only, sharing its path with the existing public `GET /certificate/{subject}`) and an `openvox-ca-ctl import-cert --certname --cert-file` CLI command to import such a certificate:

- The certificate's signature must verify against this CA's certificate (a pure cryptographic check via `CheckSignatureFrom`, so an already-expired legacy certificate is still accepted for record-keeping).
- The path `{subject}` must match the certificate's CN or one of its DNS SANs.
- CA certificates (`IsCA: true`) are rejected — use `openvox-ca-ctl import` for CA bundle import instead.
- Conflict handling, in priority order: (1) resubmitting the exact same certificate is an idempotent no-op; (2) a serial already tracked anywhere in the inventory is rejected (`409`); (3) an existing active certificate for the subject is rejected (`409`), or evicted if it's revoked.

Once imported, the certificate is tracked exactly like a normally-issued one: it shows up in listings, gets cleaned up by the expiry sweep, and can be revoked via the normal `PUT /certificate_status/{subject}` (`desired_state: "revoked"`) mechanism.

Also hardens `AppendInventory` to reject a duplicate serial atomically on **every** storage backend — previously only the SQL backend enforced this, via its DB-level unique index; filesystem/etcd/redis backends had no equivalent check at all. This benefits normal certificate issuance too, not just the new import path, and is what the import feature's serial-conflict check ultimately relies on for race-safety (a separate best-effort pre-check exists only to get error precedence right in the common, non-racing case).

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -445,6 +445,8 @@ Conflict handling, in priority order:
 2. Otherwise, if the certificate's serial number is already tracked anywhere in the inventory (under this subject or another), the request is rejected with `409 Conflict`.
 3. Otherwise, if the subject already has an active (non-revoked) certificate, the request is rejected with `409 Conflict`. If the subject's existing certificate is revoked, it is evicted and the import proceeds.
 
+Invalid certificates — malformed or multi-block PEM, a signature that does not chain to this CA, a CA certificate (`IsCA: true`), a subject that matches neither the CN nor any DNS SAN, a non-positive serial, or a bad validity window — are rejected with `400 Bad Request`. If the CA has not finished initialising, the request returns `503 Service Unavailable` (retry once it is ready).
+
 Response:
 
 ```json
--- a/README.md
+++ b/README.md
@@ -427,6 +427,32 @@ Response:
 { "private_key": "-----BEGIN RSA PRIVATE KEY-----\n...", "certificate": "-----BEGIN CERTIFICATE-----\n..." }
 ```
 
+### Certificate import
+
+| Method | Path | Description |
+|--------|------|-------------|
+| `PUT` | `/certificate/{subject}` | Import a certificate issued outside this CA's normal signing flow into the inventory; admin-only |
+
+Shares its path with `GET /certificate/{subject}` (certificate retrieval, above) but is a distinct, admin-only operation. Request body: raw certificate PEM. `{subject}` must match the certificate's CN or one of its DNS Subject Alternative Names — this lets an operator import under a specific identity even when the certificate's SANs list several names.
+
+This is for certificates that were signed by this CA's key but never went through `Sign`/`Generate` — most commonly certificates migrated from a legacy CA installation that shared this CA's key material. The certificate's signature must verify against this CA's certificate (a pure cryptographic check, so an already-expired certificate is still accepted for record-keeping); CA certificates (`IsCA: true`) are rejected — use `openvox-ca-ctl import` for CA bundle import instead.
+
+Once imported, the certificate is tracked exactly like a normally-issued one: it appears in listings and status lookups, is cleaned up by the normal expiry sweep, and can be revoked via the usual `PUT /certificate_status/{subject}` (`desired_state: "revoked"`) mechanism.
+
+Conflict handling, in priority order:
+
+1. If the exact same certificate (same serial, byte-identical) is already the tracked certificate for the subject, the request succeeds as a no-op (`"imported": false` in the response).
+2. Otherwise, if the certificate's serial number is already tracked anywhere in the inventory (under this subject or another), the request is rejected with `409 Conflict`.
+3. Otherwise, if the subject already has an active (non-revoked) certificate, the request is rejected with `409 Conflict`. If the subject's existing certificate is revoked, it is evicted and the import proceeds.
+
+Response:
+
+```json
+{ "subject": "legacy-node.example.com", "serial": "1A2B3C4D5E6F", "not_before": "2020-01-01T00:00:00Z", "not_after": "2025-01-01T00:00:00Z", "imported": true }
+```
+
+`serial` is uppercase hex (matching the inventory/CRL/OCSP convention), unlike the decimal `serial_number` field in certificate status responses (which is decimal only to preserve the full 128-bit value without int64 truncation — a constraint that doesn't apply to this string field).
+
 ### OCSP
 
 | Method | Path | Description |
@@ -457,7 +483,7 @@ When mTLS is enabled (both `--tls-cert` and `--tls-key` set), each endpoint requ
 | **Public** | None | `GET /healthz/*`, `GET /certificate/{subject}`, `GET /certificate_revocation_list/ca`, `PUT /certificate_request/{subject}`, `GET /expirations`, `POST /ocsp`, `GET /ocsp/{request}` |
 | **Any client** | Any CA-signed cert | `GET /certificate_status/{subject}` (public with `--allow-public-status`), `POST /certificate_renewal` |
 | **Self or admin** | Cert CN matches path subject, OR cert is admin | `GET /certificate_request/{subject}` |
-| **Admin** | Cert is admin (see below) | `PUT /certificate_status/{subject}`, `DELETE /certificate_status/{subject}`, `DELETE /certificate_request/{subject}`, `GET /certificate_statuses/*`, `POST /sign`, `POST /sign/all`, `POST /generate/{subject}`, `PUT /clean`, `PUT /certificate_revocation_list/ca` |
+| **Admin** | Cert is admin (see below) | `PUT /certificate_status/{subject}`, `DELETE /certificate_status/{subject}`, `DELETE /certificate_request/{subject}`, `GET /certificate_statuses/*`, `POST /sign`, `POST /sign/all`, `POST /generate/{subject}`, `PUT /clean`, `PUT /certificate_revocation_list/ca`, `PUT /certificate/{subject}` |
 
 In plain HTTP mode (no TLS), all endpoints are accessible without authentication.
 
@@ -649,6 +675,10 @@ openvox-ca-ctl reissue-crl
 # Generate a server-side key+cert pair (key saved to ./agent.example.com_key.pem)
 openvox-ca-ctl generate --certname agent.example.com
 openvox-ca-ctl generate --certname agent.example.com --dns alt.example.com --out-dir /etc/ssl
+
+# Import a certificate issued outside this CA's normal flow (e.g. migrated
+# from a legacy CA sharing this CA's key)
+openvox-ca-ctl import-cert --certname legacy-node.example.com --cert-file legacy-node_cert.pem
 
 # Bootstrap a new CA offline (no server required)
 openvox-ca-ctl setup --cadir /etc/puppetlabs/puppet/ssl --hostname puppet.example.com
--- a/cmd/openvox-ca-ctl/main.go
+++ b/cmd/openvox-ca-ctl/main.go
@@ -421,6 +421,65 @@ func newGenerateCmd() *cobra.Command {
 	return cmd
 }
 
+// newImportCertCmd registers openvox-ca-ctl's "import-cert" subcommand,
+// which imports a certificate issued OUTSIDE this CA's normal signing flow
+// (e.g. migrated from a legacy CA sharing this CA's key) into the running
+// server's inventory via PUT /certificate/{subject}. This is distinct from
+// the "import" subcommand below, which imports a whole CA cert/key/CRL
+// bundle offline, directly into storage.
+func newImportCertCmd() *cobra.Command {
+	var certname, certFile string
+	cmd := &cobra.Command{
+		Use:          "import-cert",
+		Short:        "Import an externally-issued certificate into the inventory",
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			certPEM, err := os.ReadFile(certFile)
+			if err != nil {
+				return fmt.Errorf("reading --cert-file: %w", err)
+			}
+
+			c, err := newClient()
+			if err != nil {
+				return err
+			}
+
+			path := "/puppet-ca/v1/certificate/" + certname
+			code, body, err := c.put(path, certPEM)
+			if err != nil {
+				return err
+			}
+			if err := checkHTTP(code, body, "PUT", path); err != nil {
+				return err
+			}
+
+			var result struct {
+				Subject   string `json:"subject"`
+				Serial    string `json:"serial"`
+				NotBefore string `json:"not_before"`
+				NotAfter  string `json:"not_after"`
+				Imported  bool   `json:"imported"`
+			}
+			if err := json.Unmarshal(body, &result); err != nil {
+				return fmt.Errorf("could not parse response: %w", err)
+			}
+			if result.Imported {
+				fmt.Printf("Imported %s (serial %s, valid %s to %s)\n",
+					result.Subject, result.Serial, result.NotBefore, result.NotAfter)
+			} else {
+				fmt.Printf("%s already tracked (serial %s), no changes made\n",
+					result.Subject, result.Serial)
+			}
+			return nil
+		},
+	}
+	cmd.Flags().StringVar(&certname, "certname", "", "Subject name for the imported certificate")
+	cmd.Flags().StringVar(&certFile, "cert-file", "", "Path to the certificate PEM to import")
+	_ = cmd.MarkFlagRequired("certname")
+	_ = cmd.MarkFlagRequired("cert-file")
+	return cmd
+}
+
 func newSetupCmd() *cobra.Command {
 	var caDir, hostname string
 	var encryptKey bool
@@ -571,6 +630,7 @@ Global flags must be specified before the subcommand.`,
 		newReissueCRLCmd(),
 		newCleanCmd(),
 		newGenerateCmd(),
+		newImportCertCmd(),
 		newSetupCmd(),
 		newImportCmd(),
 		newMigrateCmd(),
--- a/go.mod
+++ b/go.mod
@@ -24,6 +24,7 @@ require (
 	go.etcd.io/etcd/server/v3 v3.7.0
 	go.yaml.in/yaml/v3 v3.0.4
 	golang.org/x/crypto v0.54.0
+	modernc.org/sqlite v1.52.0
 )
 
 require (
@@ -130,7 +131,6 @@ require (
 	modernc.org/libc v1.73.4 // indirect
 	modernc.org/mathutil v1.7.1 // indirect
 	modernc.org/memory v1.11.0 // indirect
-	modernc.org/sqlite v1.52.0 // indirect
 	sigs.k8s.io/yaml v1.6.0 // indirect
 )
 
--- a/internal/api/auth_test.go
+++ b/internal/api/auth_test.go
@@ -462,6 +462,15 @@ var _ = Describe("Auth Middleware", func() {
 			mux.ServeHTTP(rr, req)
 			Expect(rr.Code).To(Equal(http.StatusForbidden))
 		})
+
+		It("returns 403 for PUT /certificate/{subject} (admin-only tier, import)", func() {
+			clientCert := issueClientCert("regular-node", caCert, caKey)
+			req := httptest.NewRequest("PUT", "/certificate/some-node", bytes.NewReader([]byte("not a real cert")))
+			req = withClientCert(req, clientCert)
+			rr := httptest.NewRecorder()
+			mux.ServeHTTP(rr, req)
+			Expect(rr.Code).To(Equal(http.StatusForbidden))
+		})
 	})
 
 	// --- Non-self client on self-or-admin endpoints ---
@@ -1056,6 +1065,10 @@ var _ = Describe("lookupTier classification", func() {
 		Entry("status delete is admin-only", "DELETE", "/certificate_status/node1", "node1", "adminOnly"),
 		Entry("unknown path falls through to admin-only", "GET", "/totally/unknown/path", "", "adminOnly"),
 		Entry("unknown method on public path is admin-only", "POST", "/certificate/node1", "node1", "adminOnly"),
+		// PUT certificate/{subject} (certificate import) shares its path with the
+		// public GET above, but is admin-only — proving the two methods on this
+		// path really do have different tiers.
+		Entry("PUT certificate is admin-only (import)", "PUT", "/certificate/node1", "node1", "adminOnly"),
 	)
 
 	DescribeTable("with AllowPublicStatus=true",
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -284,9 +284,9 @@ func (s *Server) handleGetCert(w http.ResponseWriter, r *http.Request) {
 type ImportResponse struct {
 	Subject   string `json:"subject"`
 	Serial    string `json:"serial"`
-	NotBefore string `json:"not_before"`
-	NotAfter  string `json:"not_after"`
-	Imported  bool   `json:"imported"` // false if this was a no-op (already tracked)
+	NotBefore string `json:"not_before"` // UTC timestamp, rendered with the server's configured time format (timeFormat())
+	NotAfter  string `json:"not_after"`  // UTC timestamp, rendered with the server's configured time format (timeFormat())
+	Imported  bool   `json:"imported"`   // false if this was a no-op (already tracked)
 }
 
 // handlePutCert imports a certificate that was issued OUTSIDE this CA's
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -104,6 +104,7 @@ func (s *Server) Routes() http.Handler {
 		{"PUT", "/certificate_request/{subject}", s.handlePutRequest},
 		{"DELETE", "/certificate_request/{subject}", s.handleDeleteRequest},
 		{"GET", "/certificate/{subject}", s.handleGetCert},
+		{"PUT", "/certificate/{subject}", s.handlePutCert},
 		{"GET", "/certificate_revocation_list/ca", s.handleGetCRL},
 		{"PUT", "/certificate_revocation_list/ca", s.handleReissueCRL},
 		{"POST", "/ocsp", s.handleOCSP},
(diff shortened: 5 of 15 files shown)


trevor-vaughan-ai:

>
> Models used:
> - claude-opus-4-8
> - Capable tier
> - Standard tier

_Reviewed at commit [`9293567`](https://github.com/voxpupuli/openvox-ca/commit/9293567b1861f9b304742d3c0c724a3ccf805f81) (deep effort)._

**TL;DR:** Council approves PR #115: the certificate-import feature is admin-gated, cryptographically sound, and well-tested, with 15 non-blocking advisories (3 medium, 12 low).

**Findings:** 🔴 0 Critical, 🟠 0 High, 🟡 3 Medium, 🔵 12 Low

- 🧪 **Documented race-safe authoritative enforcement path (ErrDuplicateSerial mapping + rollback) has no ca-layer test** ([`internal/ca/importcert.go:209`](https://github.com/voxpupuli/openvox-ca/blob/9293567b1861f9b304742d3c0c724a3ccf805f81/internal/ca/importcert.go#L209))

  > `if errors.Is(err, storage.ErrDuplicateSerial) {`

  💡 **Recommendation:**
  - Add a test that forces `AppendInventory` to return `ErrDuplicateSerial` after the pre-check passes - e.g. append an inventory entry for the serial out-of-band between the `SerialExists` check and the append, or inject a storage stub - and assert (a) the returned error matches `ErrSerialExists` and (b) `store.HasCert(subject)` is false afterward (rollback occurred).

  <br>

  **Constraint:**
  - Coverage Completeness / regression protection for the mechanism the code documents as authoritative.

  **Description:**
  - The code (lines 172-177) explicitly states the `SerialExists` pre-check is an *ordering pre-check only* and that "AppendInventory below ... [is] the authoritative, race-safe enforcement." That authoritative branch - mapping `storage.ErrDuplicateSerial` back to `ErrSerialExists` (line 209) and rolling the just-saved cert back with `DeleteCert` (line 205) to keep storage and inventory in sync - is exercised by no test. Only the *pre-check* path (`SerialExists`) is tested (lines 188, 200). `internal/storage/duplicate_serial_test.go` proves `AppendInventory` returns `ErrDuplicateSerial` at the storage layer, but the ImportCertificate translation and rollback-on-conflict are untested, so a regression that dropped the `errors.Is` mapping (leaking a raw inventory error instead of `ErrSerialExists`) or skipped the rollback (leaving an orphaned cert file after a losing race) would not be caught here.
  </details>

- 🧭 **Inventory-line format hand-rolled a third time, bypassing the documented single source of truth** ([`internal/ca/importcert.go:196`](https://github.com/voxpupuli/openvox-ca/blob/9293567b1861f9b304742d3c0c724a3ccf805f81/internal/ca/importcert.go#L196))

  > `inventoryEntry := fmt.Sprintf("%s %s %s /%s",`

  💡 **Recommendation:**
  - Build the inventory line through `canonicalInventoryLine` (constructing an `InventoryEntry`) and format the times via the existing `inventoryTimeFormat` constant, so the import path cannot drift from the writer/reader/HMAC path. Consider aligning `signing.go:362` the same way to fully retire the duplication.

    APPROVE

  <br>

  **Constraint:**
  - Structural Coherence / Zero-Waste (DRY as waste prevention - duplicated business logic that will drift apart during maintenance)

  **Description:**
  - The on-disk inventory line format is documented as having a single source of truth: `canonicalInventoryLine` in `internal/storage/storage.go:773-777` states it "is the single source of truth for the on-disk blob format and the input to the integrity hash chain, so the two cannot drift." The new code hand-rolls that same `"%s %s %s /%s"` layout inline instead, adding a third independent copy (the second is `signing.go:362`). It also hardcodes the time layout literal `"2006-01-02T15:04:05UTC"` even though a named constant `inventoryTimeFormat` exists for exactly this (`cleanup.go:35`, consumed by the reader at `cleanup.go:69`). Because this line feeds the HMAC integrity hash chain (`chainInventoryMAC`, `storage.go:766`), a future edit to the canonical format or time layout that misses this inline copy would silently break inventory integrity verification for imported certs. Note this matches the existing (imperfect) `signing.go` pattern, so the new code is at least internally consistent with its sibling - the drift is incremental, not novel, which keeps it below HIGH.
  </details>

- ⚙️ **Duplicate-serial scan doubles the whole-blob read on the blob-backend append path** ([`internal/storage/storage.go:219`](https://github.com/voxpupuli/openvox-ca/blob/9293567b1861f9b304742d3c0c724a3ccf805f81/internal/storage/storage.go#L219))

  > `// Blob backends have no native uniqueness constraint on serial, so check`

  💡 **Recommendation:**
  - Read `KeyInventory` once under the lock and reuse the bytes for both the duplicate scan (parse) and the HMAC recompute - e.g. have the blob path read the blob, scan the parsed entries for the duplicate, then pass the already-read bytes into an HMAC helper instead of calling `computeInventoryHMAC` which re-reads. Non-blocking; the current form is correct, just does one avoidable full-blob read per issuance.

  <br>

  **Evidence**:

  ```go
  // Blob backends have no native uniqueness constraint on serial, so check
  // explicitly under inventoryMu (already held) before appending.
  existing, err := s.inventoryEntriesLocked(ctx)
  if err != nil {
      return err
  }
  for _, e := range existing {
      if e.Serial == parsed.Serial {
          return fmt.Errorf("%w: %s", ErrDuplicateSerial, parsed.Serial)
      }
  }
  ```

  **Constraint:**
  - Operator criterion #2 - redundant reads that could be combined (`severity.md`: "Redundant read", MEDIUM).

  **Description:**
  - On the blob path, `inventoryEntriesLocked` (line 219) reads and parses the entire `KeyInventory` blob via `readInventoryForHMAC` → `backend.Get`, and then a few lines later `updateInventoryHMACLocked` (line 234) → `computeInventoryHMAC` → `readInventoryForHMAC` reads the *same* blob again to recompute the whole-blob HMAC. Both reads happen under the same continuously-held `inventoryMu.Lock`, so the blob cannot change between them - a single read could feed both the duplicate scan and the HMAC computation.

  Honest scoping of the impact:
  - This is a constant-factor increase, not a new asymptotic regression. The whole-blob read+hash on every append already existed (HMAC recompute), so blob-backend inventory build was already O(n) per append / O(n²) overall. The dedup adds exactly one more full read+parse per append.
  - It is nonetheless a real, avoidable cost that scales with inventory size on the certificate-issuance hot path. For remote blob backends each `Get(KeyInventory)` transfers the entire inventory: Redis (`redis.go:414`, server-side Lua append) now does 2 whole-blob Gets per append instead of 1; etcd (`etcd.go:300`, whose `AppendLine` already does a full Get) now does 3 instead of 2. The added read also lengthens the `inventoryMu` write-lock critical section, marginally increasing append serialization latency.
  - Holding the *write* lock across the scan is correct and necessary (check-then-append must be atomic); that is not the issue - the redundant second full read is.
  </details>

</details>

<details><summary>🔵 LOW (12)</summary>

- 📚 **ImportResult.Serial doc comment refers to the unexported serialHexStr helper** ([`internal/ca/importcert.go:49`](https://github.com/voxpupuli/openvox-ca/blob/9293567b1861f9b304742d3c0c724a3ccf805f81/internal/ca/importcert.go#L49))

  > `Serial    string // serialHexStr(cert.SerialNumber) - matches inventory/CRL/OCSP form`

  💡 **Recommendation:**
  - Describe the format directly, e.g. "uppercase hex serial, no leading zeros - the form used in the inventory, CRL, and OCSP responses", dropping the unexported function name.

  <br>

  **Constraint:**
  - godoc convention - doc comments on an exported field should describe the value in terms visible to a consumer, not by naming an unexported implementation detail (`serialHexStr` is unexported, signing.go:57).

  **Description:**
  - A reader of the exported `ImportResult` type cannot resolve `serialHexStr` from the rendered godoc. The load-bearing part of the comment ("matches inventory/CRL/OCSP form") is fine; the leading symbol reference adds nothing a consumer can act on. Cosmetic, non-blocking.
  </details>

- 🧪 **x509.ParseCertificate failure branch untested** ([`internal/ca/importcert.go:99`](https://github.com/voxpupuli/openvox-ca/blob/9293567b1861f9b304742d3c0c724a3ccf805f81/internal/ca/importcert.go#L99))

  > `cert, err := x509.ParseCertificate(block.Bytes)`

  💡 **Recommendation:**
  - Add a case passing `pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: []byte("not-der")})` and assert `MatchError(ca.ErrImportInvalid)`.

  <br>

  **Constraint:**
  - Coverage Completeness - negative/error paths.

  **Description:**
  - The malformed-PEM test (line 307) exercises `pem.Decode` returning nil, but the distinct branch where PEM decoding succeeds yet `x509.ParseCertificate` fails (a `-----BEGIN CERTIFICATE-----` block containing garbage bytes) is never hit. Low severity: the branch is a trivial error wrap and the surrounding negative coverage is thorough.
  </details>

- 🧪 **503 not-ready branch of handlePutCert is untested** ([`internal/api/handlers.go:319`](https://github.com/voxpupuli/openvox-ca/blob/9293567b1861f9b304742d3c0c724a3ccf805f81/internal/api/handlers.go#L319))

  > `case errors.Is(err, ca.ErrNotInitialized):`

  💡 **Recommendation:**
  - Add a spec that constructs a CA without `Init()` (or otherwise not ready), PUTs a certificate, and asserts `rr.Code == http.StatusServiceUnavailable`.

  <br>

  **Constraint:**
  - Coverage completeness - error-path branches with a distinct, contractual status code must be exercised (severity pack: "test isolation concern / convention deviation" tier; Go hint: missing error-path tests).

  **Description:**
  - `ImportCertificate` returns `ca.ErrNotInitialized` when the CA is not yet initialized (importcert.go:84-85), and the handler deliberately maps it to 503 rather than the generic 500 default case - the same "poll again while the CA warms up" contract used by `handleGetExpirations` (handlers.go:845). No test in `import_test.go` submits to this handler against a not-ready CA; every spec calls `myCA.Init(...)` in `BeforeEach` (import_test.go:102). A grep for `ServiceUnavailable`/`ErrNotInitialized` across `internal/api/*_test.go` matches only the health handlers and an unrelated handler at api_test.go:640 - never the import path. A regression collapsing this branch into the 500 default (returning "internal server error") would go unnoticed.
  </details>

- 🧪 **Import 200 response never asserts not_before / not_after** ([`internal/api/import_test.go:127`](https://github.com/voxpupuli/openvox-ca/blob/9293567b1861f9b304742d3c0c724a3ccf805f81/internal/api/import_test.go#L127))

  > `var result api.ImportResponse`

  💡 **Recommendation:**
  - In the 200 spec, assert `result.NotBefore` and `result.NotAfter` are non-empty and parse under `time.RFC3339` (mirroring api_test.go:1314-1317).

    APPROVE

  <br>

  **Constraint:**
  - Assertion quality - verify the response struct fields that form the success contract.

  **Description:**
  - `ImportResponse` carries `NotBefore`/`NotAfter` formatted through `s.timeFormat()` (handlers.go:338-339), but no spec asserts either field. The success test checks `Subject`, `Imported`, and `Serial` (non-empty, and equal across the idempotent resubmission), so this is not an `err == nil`-only shallow test - it is a partial gap on the response contract. Date formatting is covered indirectly for other response types (api_test.go:1306-1335) and `timeFormat()` is shared, which keeps the severity low; still, a bug that emitted an empty or malformed `not_before`/`not_after` on the import path would not be caught here.
  </details>

- 🛡️ **SQLite unique classifier also matches the broad SQLITE_CONSTRAINT primary code** ([`internal/storage/sql.go:522`](https://github.com/voxpupuli/openvox-ca/blob/9293567b1861f9b304742d3c0c724a3ccf805f81/internal/storage/sql.go#L522))

  > `return liteErr.Code() == 2067 || liteErr.Code() == 19 // SQLITE_CONSTRAINT_UNIQUE / SQLITE_CONSTRAINT`

  💡 **Recommendation:**
  - Optionally narrow the SQLite check to the extended code `2067` only (matching the MySQL/PostgreSQL precision the comment describes), or amend the comment to acknowledge that the SQLite branch intentionally accepts the broad primary code as a driver-compatibility fallback. Non-blocking.

    APPROVE

    Duplicate-serial enforcement is correct and injection-safe: the SQL backend is backed by a real unique index inside an atomic transaction with a non-retryable, correctly-classified violation, and the blob path is atomic within a process under a held write lock. The cross-replica gap on etcd/redis is honestly and prominently documented, and all SQL access is parameterized. The single LOW note about the SQLite classifier's breadth is informational and has no path to admitting a duplicate or corrupting data.

  <br>

  **Constraint:**
  - The function's own doc comment (`sql.go:504-510`) states it "deliberately checks the specific unique-violation SQLSTATE/error code for each dialect rather than a broader 'integrity violation' classifier, so a future schema change adding an unrelated constraint … is not silently misreported as a duplicate serial."

  **Description:**
  - For MySQL (1062) and PostgreSQL (23505) the classifier matches the exact unique-violation code, matching the stated intent. For SQLite it also matches `19` (SQLITE_CONSTRAINT), the generic primary result code covering NOT NULL, CHECK, FOREIGN KEY, and PRIMARY KEY violations - the exact breadth the comment says it is avoiding. This is defensive against `modernc.org/sqlite` returning the primary rather than the extended code, but it contradicts the documented "specific code only" design. Impact is bounded: on `puppet_ca_inventory` the `id` PK is autoincrement (cannot conflict on insert) and all NOT NULL columns are populated from `parseInventoryEntry` (which guarantees four non-empty fields), so no non-unique constraint can realistically fire on the `AppendEntry` insert. Even if one did, the transaction has already rolled the row back, so the only effect would be an error mislabelled as `ErrDuplicateSerial` - no duplicate is admitted and no data is corrupted. Reported for completeness because the delegation asked specifically about the 2067-vs-19 breadth.
  </details>

- 📚 **ImportCertificate godoc omits its returned sentinel errors** ([`internal/ca/importcert.go:83`](https://github.com/voxpupuli/openvox-ca/blob/9293567b1861f9b304742d3c0c724a3ccf805f81/internal/ca/importcert.go#L83))

  > `ErrImportInvalid`

  💡 **Recommendation:**
  - Add an error-contract line to the `ImportCertificate` godoc listing the sentinel errors it returns, following the bulleted-list style at signing.go:73-76.

  <br>

  **Constraint:**
  - Documentation convention - sibling functions in this package list their returned sentinel errors in godoc (signing.go:61-76, generate.go:71); `ImportCertificate` omits three of its four.

  **Description:**
  - `ImportCertificate` returns four distinct sentinel errors but its godoc enumerates only one, so a caller using `errors.Is` to branch on outcome has no documented contract for the other three. The validation gate confirmed the conflict priority order itself IS documented (README.md:442, numbered list), so only the godoc sentinel-error omission stands - severity reduced to LOW.
  </details>

- ⚙️ **SerialExists pre-check duplicates the authoritative atomic check with a full-inventory scan** ([`internal/ca/importcert.go:184`](https://github.com/voxpupuli/openvox-ca/blob/9293567b1861f9b304742d3c0c724a3ccf805f81/internal/ca/importcert.go#L184))

  > `exists, err := c.Storage.SerialExists(ctx, serialStr)`

  💡 **Recommendation:**
  - Acceptable as-is given import's rarity. If the priority-ordering guarantee can be derived from `AppendInventory`'s duplicate error alone (it already distinguishes the duplicate-serial case), the pre-scan could be dropped to avoid the redundant full-inventory read. No change required for merge.

  <br>

  **Constraint:**
  - Operator §2 - redundant computation that could be avoided (efficiency, scope-bounded to demonstrable impact).

  **Description:**
  - Every import performs an O(n) scan/parse of the entire inventory while holding both the per-subject lock and the global `c.mu`, purely to select which error to return. The duplicate is already detected authoritatively and race-safely by `AppendInventory` (`ErrDuplicateSerial`, remapped at `importcert.go:215-217`), so the scan adds no correctness guarantee. Impact is bounded and low: import is an infrequent migration/admin operation run one cert at a time, not a hot path, and inventory parse cost is small relative to import frequency. Flagging as informational only - the redundancy is a deliberate, documented tradeoff to honor the ErrSerialExists-over-ErrCertExists priority order.
  </details>

- 🧪 **No concurrent/`-race` test exercising ImportCertificate despite documented cluster-wide serialization** ([`internal/ca/importcert_test.go:40`](https://github.com/voxpupuli/openvox-ca/blob/9293567b1861f9b304742d3c0c724a3ccf805f81/internal/ca/importcert_test.go#L40))

  > `Describe("ImportCertificate", ...)`

  💡 **Recommendation:**
  - Mirror the `ca_test.go:1577` autosign race test: launch N goroutines importing the same serial/subject concurrently and assert exactly one succeeds (`Imported==true`) while the rest return `ErrSerialExists`/`ErrCertExists`, run under `go test -race`.

  <br>

  **Constraint:**
  - lang-go testing_conventions - "flag missing race detection in concurrent code tests."

  **Description:**
  - ImportCertificate's design centers on race safety (per-subject `WithLock` at line 135, `c.mu` at 136, and the race-safe serial enforcement at 202-213). No test drives two concurrent imports for the same subject or the same serial, unlike the existing pattern at `ca_test.go:1577` (`allows exactly one autosign when concurrent requests race for the same subject`). The underlying locking is tested indirectly through that autosign test and the storage layer, so this is non-blocking, but a direct concurrent-import test under `-race` would lock down the documented property.
  </details>

- 🧪 **Several negative tests assert only the ErrImportInvalid sentinel, not the failing branch** ([`internal/ca/importcert_test.go:333`](https://github.com/voxpupuli/openvox-ca/blob/9293567b1861f9b304742d3c0c724a3ccf805f81/internal/ca/importcert_test.go#L333))

  > `Expect(err).To(MatchError(ca.ErrImportInvalid))`

  💡 **Recommendation:**
  - Add a `ContainSubstring` assertion pinning the expected message fragment to each of these tests (e.g. "non-positive serial number", "NotAfter", "matches neither").

    APPROVE

    The ImportCertificate test suite is well-architected and covers the vast majority of documented branches with meaningful, isolated, sandboxed assertions - including the security-critical chain-of-trust, CA-reject, and subject-match paths with distinguishing message checks. The gaps are confined to error-injection and race paths that do not block merge: the strongest of these (MEDIUM) is the absence of any ca-layer test for the AppendInventory `ErrDuplicateSerial`→`ErrSerialExists` mapping and rollback, which the production code itself flags as the *authoritative* race-safe enforcement mechanism; the remaining items (concurrent-import test, ParseCertificate-fail branch, and sentinel-only assertions) are LOW. No HIGH or CRITICAL findings.

  <br>

  **Constraint:**
  - Assertion Quality - depth on which contract branch is exercised.

  **Description:**
  - Because every input-validation branch returns the same `ErrImportInvalid` sentinel, a regression that made one condition short-circuit another (e.g. the non-positive-serial check masking the backwards-validity check) could leave these tests green while the wrong branch executed. The security-relevant checks already guard against this with distinguishing substrings; the remaining input-validation tests would benefit from the same precision. Non-blocking.
  </details>

- 📚 **ImportResponse timestamp fields do not document their string format** ([`internal/api/handlers.go:287`](https://github.com/voxpupuli/openvox-ca/blob/9293567b1861f9b304742d3c0c724a3ccf805f81/internal/api/handlers.go#L287))

  > `NotBefore string \\`

  💡 **Recommendation:**
  - Add a short inline comment on `NotBefore`/`NotAfter` noting they are UTC timestamps rendered with the server's configured time format (`timeFormat()`), mirroring the existing comment on `Imported`.

    ### Notes

    - No Docs repo / forge tooling was provided in the delegation prompt, so no
      documentation-issue filing command is included. The new PUT endpoint is a
      user-facing behavior change; whether the top-level `README.md` HTTP-API
      contract needs a corresponding entry is owned by the README subsystem
      review, not this one, and is flagged there rather than duplicated here.

    APPROVE

  <br>

  **Constraint:**
  - godoc convention - an exported response type is part of the API surface; a `string` timestamp field should state its format so consumers and maintainers do not have to trace the encoder.

  **Description:**
  - `NotBefore` and `NotAfter` are typed `string`, and their value is produced by `result.NotBefore.UTC().Format(s.timeFormat())` (handlers.go:338-339). A maintainer or API consumer reading `ImportResponse` alone cannot tell the wire format of these fields without following the call into `timeFormat()` (handlers.go:670). The struct doc comment (handlers.go:283) documents the type as a whole but not the per-field format. This is a minor, non-blocking discoverability gap; the `Serial` and `Subject` fields are self-explanatory and need no comment.
  </details>

- 🧪 **etcd and redis backends lack an end-to-end duplicate-serial rejection assertion** ([`internal/storage/etcd_integration_test.go:261`](https://github.com/voxpupuli/openvox-ca/blob/9293567b1861f9b304742d3c0c724a3ccf805f81/internal/storage/etcd_integration_test.go#L261))

  > `The changed etcd block only reformats the appended lines and re-asserts the`

  💡 **Recommendation:**
  - Optionally add to the existing etcd and redis end-to-end `It` blocks a
    third `AppendInventory` reusing `line1`'s serial (`"0001 ... /node3"`) asserting
    `errors.Is(err, ErrDuplicateSerial)` and that `ReadInventory` is unchanged, mirroring the
    filesystem case. Not required for merge given the shared, already-covered code path.

    APPROVE

  <br>

  happy path - `const line1 = "0001 ..."`, `Expect(svc.AppendInventory(..., line1)).To(Succeed())`,
  `Expect(svc.AppendInventory(..., line2)).To(Succeed())`, then a `ReadInventory` equality
  check. `grep -rn "ErrDuplicateSerial" internal/storage/` returns hits only in
  `duplicate_serial_test.go` and `storage.go` - neither `etcd_integration_test.go` nor
  `redis_test.go` (identical reformat at `redis_test.go:283`) asserts a duplicate is rejected.

  **Constraint:**
  - Coverage completeness vs. the PR's "every storage backend" hardening claim
  (severity.md: optional test expansion is LOW, not HIGH).

  **Description:**
  - Duplicate-serial rejection for etcd and redis is not directly asserted; only
  SQL (unique-index path) and filesystem (blob-scan path) have explicit `ErrDuplicateSerial`
  tests. The risk is low because etcd and redis are blob backends (no `InventoryStore`
  implementation), so their rejection runs the exact shared branch in `storage.go:217-227`
  that the filesystem test already exercises, and the pre-existing round-trip assertions
  (`etcd_integration_test.go:265-269`, `redis_test.go:287-290`) already prove `Get` after
  `AppendLine` returns the appended entries the scan reads. The distributed backends are also
  where the code deliberately provides no cross-replica guarantee (storage.go:179-184), so a
  single-process rejection test there would document intended in-process behavior.
  </details>

- 📚 **Rejection status code (400) not stated for invalid-certificate cases** ([`README.md:115`](https://github.com/voxpupuli/openvox-ca/blob/9293567b1861f9b304742d3c0c724a3ccf805f81/README.md#L115))

  > `IsCA: true`

  💡 **Recommendation:**
  - Add a brief note that invalid certificates are rejected with `400 Bad Request` (and, optionally, that an uninitialized CA returns `503`), mirroring how the `409` conflict cases are called out.

    APPROVE

  <br>

  **Constraint:**
  - API reference completeness - conflict cases document their HTTP status (`409 Conflict`) explicitly, but the invalid-input rejection cases document only the behavior, not the status code.

  **Description:**
  - The section's "Conflict handling" list names `409 Conflict` for the serial and active-cert cases, but the invalid-input rejections (malformed/multi-block PEM, signature not chaining to the CA, `IsCA: true`, subject/CN-SAN mismatch, non-positive serial, bad validity window) return `400 Bad Request` (`handlers.go:321-323`, `errors.Is(err, ca.ErrImportInvalid)`) and this status is never stated. A reader consuming this as an API reference cannot tell what code a rejected import returns. This is a minor completeness gap in otherwise thorough documentation; the qualitative behavior is described correctly.
  </details>

</details>

---

bootc:
Thanks — worked through the Review Council advisories. All 15 were non-blocking; I addressed 12 and deliberately skipped 3. Pushed as four commits (`2a0cdb7`, `cea77bd`, `ac0dedd`, `2c119d7`) on top of the rebased branch. No `go.mod`/`go.sum` changes, so the FIPS/boringcrypto contract is untouched (only `crypto/hmac` + `crypto/sha256`, both already in use).

### Addressed

**Medium**
- **Inventory-line format hand-rolled a third time** — added `storage.FormatInventoryLine` (plus an exported `storage.InventoryTimeFormat`) as the single constructor, and routed both `signing.go` and the import path through it; `cleanup.go` now shares the same time-format constant. One source of truth alongside `canonicalInventoryLine`.
- **Duplicate-serial scan doubles the whole-blob read** — the blob-backend `AppendInventory` path now reads the inventory once under the held write lock and reuses those bytes for both the duplicate scan and the HMAC recompute (reconstructing the post-append blob in memory rather than re-fetching it). Factored the whole-blob HMAC into `wholeBlobInventoryMAC` so the recompute-from-storage and append-in-place paths can't diverge; the HMAC/tamper tests confirm the value is byte-identical.

**Low** — SQLite classifier comment expanded to own the `19` breadth as an intentional driver-compat fallback; `ImportResult.Serial` doc rewritten in consumer terms; `ImportCertificate` godoc now lists all four sentinel errors; `ImportResponse` timestamp format documented; README now states `400`/`503` alongside the existing `409`; new tests for the `x509.ParseCertificate`-fail branch, the `503` not-ready handler branch, `not_before`/`not_after` in the 200 response, `ContainSubstring` on the three named negative tests, and duplicate-serial rejection assertions in the etcd + redis round-trip tests.

### Not addressed (with reasons)

- **Medium: no ca-layer test for the `ErrDuplicateSerial`→`ErrSerialExists` mapping + rollback**, and **Low: no concurrent `-race` import test** — both target the *authoritative* `AppendInventory` conflict branch, which is **unreachable from a single process**. `ImportCertificate` holds the global `c.mu` around the whole `importCertificateLocked` body, so every import is serialised and the `SerialExists` pre-check always catches a duplicate first; even two concurrent imports of the same serial under different subjects serialise on `c.mu`, so the loser still trips the pre-check, not the authoritative branch. That branch only fires on a genuine cross-process race against a SQL backend (its DB unique index) — it's defensive belt-and-braces. Reaching it deterministically needs an injected storage stub, which would mean turning `CA.Storage` from a concrete `*storage.StorageService` into an interface — disproportionate for a non-blocking advisory. The storage-layer `ErrDuplicateSerial` behaviour itself is covered by `duplicate_serial_test.go` and now by the new etcd/redis assertions.
- **Low: `SerialExists` pre-check redundancy** — left as-is, per the advisory's own "no change required for merge". It's a deliberate, documented trade-off to honour the `ErrSerialExists`-over-`ErrCertExists` priority order.

trevor-vaughan (approved):

PR #123, opened 2026-07-18
fix(ci): stop the Redis backend test flaking on the bootstrap race

bootc:
The `backends-redis` job flaked on PR #119 (a Renovate MySQL-digest bump that touches nothing Redis-related), so the failure is pipeline flakiness rather than the PR.

## What was happening

`backends-redis` is the only storage-backend suite that runs **two** `openvox-ca` replicas against a shared Redis backend, so it is the only one exercising the concurrent CA bootstrap path. The bootstrap/lock code itself is correct, but neither replica had a restart policy. When replica 1 lost the bootstrap race — or its Phase 1 loopback CA was slow to answer under CI load — its entrypoint aborted and the container stayed dead for the entire run. Every request to it then returned HTTP `000` and every `compose exec` came back empty, cascading into ~29 downstream "failures" that were all really one dead container.

The readiness wait loops made this hard to see: they printed ` OK` whether or not the endpoint ever answered, so the first honest signal was a `FATAL: CA not reachable` much later, with no logs.

## Changes

- **`compose-backends-redis.yml`**: add `restart: on-failure` to both CA replicas. A replica that loses the bootstrap race now restarts and recovers — on the next start the CA cert/key already exist in shared Redis, so bootstrap short-circuits, the replica regenerates its own TLS service cert, and it comes up. This also mirrors how a real clustered CA replica would be supervised. If a failure were ever *deterministic* the healthcheck would still never pass and the wait loop would still fail, so this heals transient races without masking real bugs.
- **`test/backends/redis-stack.sh`** and **`test/puppet/puppet-stack.sh`**: make the readiness wait loops fail loudly. On timeout they now print a clear message and dump the offending container's last 80 log lines, instead of printing ` OK` and marching on. Fixing `puppet-stack.sh` also improves diagnosis for the Puppet stack and Compose integration jobs, which share it.

No production code changes; the CA bootstrap and Redis lock implementation are untouched.

### DIFF
--- a/compose-backends-redis.yml
+++ b/compose-backends-redis.yml
@@ -90,6 +90,15 @@ services:
     # Distinct host port from compose-puppet.yml so both stacks can coexist.
     ports:
       - "8241:8140"
+    # Both replicas race to bootstrap the shared CA on first start. If this one
+    # loses that race and its Phase 1 entrypoint aborts (or the loopback CA is
+    # slow to answer under CI load), restart it: on the next attempt the CA
+    # cert/key already exist in the shared Redis backend, so bootstrap short-
+    # circuits, the replica generates its own TLS service cert, and it comes up.
+    # Without this a transiently-failed replica stays dead for the whole run and
+    # cascades into dozens of downstream failures. Mirrors how a real CA replica
+    # in a cluster would be supervised.
+    restart: on-failure
     depends_on:
       redis:
         condition: service_healthy
@@ -130,6 +139,10 @@ services:
       - "8140"
     ports:
       - "8242:8140"
+    # See the restart rationale on the openvox-ca service above: a replica that
+    # loses the bootstrap race must be able to restart and recover from shared
+    # Redis state rather than staying dead for the whole run.
+    restart: on-failure
     depends_on:
       redis:
         condition: service_healthy
--- a/test/backends/redis-stack.sh
+++ b/test/backends/redis-stack.sh
@@ -59,6 +59,20 @@ for arg in "$@"; do
     esac
 done
 
+# -- Helper: abort a readiness wait loudly, dumping the culprit's logs ------
+# The wait loops below used to print " OK" whether or not the endpoint ever
+# answered, so a replica that never came up surfaced only as a downstream
+# cascade of HTTP-000 test failures. Calling this on timeout instead turns
+# that into a single actionable failure plus the container logs that explain
+# why (e.g. a Phase 1 bootstrap abort on the losing replica).
+abort_not_ready() {  # human-description  service-name
+    printf ' TIMEOUT\n'
+    printf 'FATAL: %s did not become ready in time\n' "$1" >&2
+    printf '# ---- last 80 log lines from %s ----\n' "$2" >&2
+    "${_COMPOSE[@]}" logs --tail 80 "$2" >&2 2>/dev/null || true
+    exit 1
+}
+
 # -- Stack lifecycle + temp dir cleanup ------------------------------------
 WORK_DIR=$(mktemp -d /tmp/redis-stack-integ.XXXXXX)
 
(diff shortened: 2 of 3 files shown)


trevor-vaughan-ai:

>
> Models used:
> - claude-opus-4-8[1m]

_Reviewed at commit [`02ec866`](https://github.com/voxpupuli/openvox-ca/commit/02ec866e95af03e990709a3d22607932678daf5c) (deep effort)._

**TL;DR:** Council approves PR #123: the CI-flake fix is sound; two non-blocking MEDIUM advisories on restart-count masking and a swallowed stderr log dump.

**Findings:** 🔴 0 Critical, 🟠 0 High, 🟡 2 Medium, 🔵 0 Low

|---|---|---|
| 🧪 Tester (code) | ✅ Approve | 1 MEDIUM |

- 🧪 **`restart: on-failure` with no bound can mask a crash-on-lost-bootstrap-race defect the multi-replica suite exists to catch** ([`compose-backends-redis.yml:101`](https://github.com/voxpupuli/openvox-ca/blob/02ec866e95af03e990709a3d22607932678daf5c/compose-backends-redis.yml#L101))

  > `restart: on-failure`

  💡 **Recommendation:**
  - Keep the flake tolerance but restore detection: after the readiness gates, assert the CA replicas' restart count is 0 (e.g. `"${_COMPOSE[@]}" ps` / engine `inspect` of `.State.RestartCount`) and `fail` if a replica had to restart to become ready. That converts a silent auto-recovery back into a visible, actionable signal while still preventing the downstream cascade the PR set out to eliminate.

  <br>

  **Constraint:**
  - Test-harness integrity - a CI change must surface, not suppress, the class of failure the suite is designed to detect (severity pack, Tester column: "test masks real failure").

  **Description:**
  - The PR body claims the healthcheck plus wait-loop timeout still catches deterministic failures. That holds only for a replica that crashes on *every* attempt: it never answers `:8241`/`:8242`, the readiness loop exhausts its 180s budget (redis-stack.sh:112-120, 570-578), and `abort_not_ready` fails the run. It does **not** hold for the crash-once-then-recover case, which is exactly the bootstrap-race shape this suite targets. If a real regression makes the losing replica *exit non-zero* on a lost race (instead of gracefully short-circuiting on the already-written `ca:cert`), `restart: on-failure` silently restarts it; on the next attempt the cert exists in shared Redis, it comes up, the readiness loop prints `OK`, and Phase 4's "bootstrap lock worked" assertion (redis-stack.sh:594-599) passes. Before this change the dead replica would have cascaded into failures and turned the run red. No assertion anywhere checks container restart count, so the suite can no longer distinguish an acceptable transient from a genuine crash-recover defect in its headline scenario. This weakens (does not merely fail to add) the regression protection the second replica was introduced to provide. I hold this at MEDIUM rather than HIGH because the masking is contingent on a latent code defect actually existing, and the crash-every-time case is still caught.
  </details>

- ⚙️ **Log-dump redirection sends stderr to /dev/null, discarding the diagnostics it exists to surface** ([`test/backends/redis-stack.sh:72`](https://github.com/voxpupuli/openvox-ca/blob/02ec866e95af03e990709a3d22607932678daf5c/test/backends/redis-stack.sh#L72))

  > `"${_COMPOSE[@]}" logs --tail 80 "$2" >&2 2>/dev/null || true`

  💡 **Recommendation:**
  - Drop the `2>/dev/null` and use plain `>&2`, or merge both streams to stderr with `2>&1 1>&2`, so container stderr (and any error from the `logs` invocation itself) reaches the operator.

  <br>

  **Constraint:**
  - runtime-observability - diagnostics emitted at a failure boundary must actually reach the operator; a dump helper must not silently drop the logs it exists to show.

  **Description:**
  - The identical line also appears at `test/puppet/puppet-stack.sh:193`. Redirections apply left-to-right: `>&2` first points fd1 (stdout) at the current stderr target (the terminal), then `2>/dev/null` points fd2 (stderr) at `/dev/null`. Net result: the compose command's stdout is shown but its stderr is discarded. The stated purpose of `abort_not_ready` (comments at lines 62-67 / 184-188) is to dump the container logs that explain why a service failed (e.g. a Phase 1 bootstrap abort on the losing replica). But under `podman-compose` - which `redis-stack.sh:28-40` selects by default whenever `podman` is on PATH (it defaults `_ENGINE=podman` before checking docker, and prefers `podman-compose`) - `logs` delegates to `podman logs`, which replays the container's stdout to stdout and its stderr to stderr; a Go service that writes startup/abort diagnostics to stderr (the common default) will have exactly those lines routed to fd2 -> `/dev/null`. The operator then sees the `# ---- last 80 log lines from <svc> ----` header followed by nothing, on the one code path whose job is to explain a failed bootstrap. If the `logs` invocation itself errors, that error is also swallowed by `2>/dev/null || true`, so an empty dump is indistinguishable from "service produced no logs." Under `docker compose logs` (which prints everything to its own stdout) the dump is intact, so the defect is engine-conditional - hence PLAUSIBLE rather than CONFIRMED.
  </details>

</details>

---

bootc:
Thanks — worked through both advisories in 9ce8691.

**🧪 Log-dump swallows stderr (`redis-stack.sh:72`) — fixed.**

Confirmed and fixed. `>&2 2>/dev/null` redirects the compose command's stdout to stderr but then sends its *stderr* to `/dev/null`, and under `podman logs` a container's stderr (where a Go service writes its startup/abort diagnostics) is replayed to the command's stderr — so the one path whose job is to explain a failed bootstrap was discarding exactly the lines that explain it. Dropped the `2>/dev/null`; plain `>&2` already routes fd1 to stderr and fd2 defaults there, so both streams now reach the operator. Applied to the identical line in `puppet-stack.sh:193` too, and added a comment on why the `2>/dev/null` must not come back.

**⚙️ `restart: on-failure` masking (`compose-backends-redis.yml:101`) — addressed, but not as a hard-fail.**

The concern is valid, but I declined the specific recommendation to assert `RestartCount == 0` and `fail`, because that re-arms the exact flake this PR exists to remove. The compose comment (lines 94–100) is explicit that a replica losing the bootstrap race under CI load *legitimately* aborts and restarts to recover from shared Redis state. A benign slow-loopback restart and a genuine crash-on-lost-race regression are indistinguishable from the outside — both show `RestartCount ≥ 1` — so a hard-fail assertion would fail on the benign case and reintroduce the flake.

What the finding really wants is for the auto-recovery to stop being *silent*, and that part I've done: after the Phase 4 readiness gates, the suite now inspects each CA replica's `RestartCount` (via `<compose> ps -q` + engine `inspect`) and emits it as a visible, non-fatal diagnostic. A clean run prints `restart count: 0`; any restart prints a loud `NOTE` on stderr pointing a human at the replica logs. That restores the actionable signal without converting a tolerated transient back into a red run. It degrades gracefully to `unknown` if the engine can't be inspected.

trevor-vaughan (approved):

PR #142, opened 2026-07-25
Complete the release machinery: version stamping, native-runner artefact builds, release gates, and release:prepare

bootc:
Completes the release machinery ahead of v0.9.0. What was originally planned as three PRs is consolidated here so the machinery, its CI coverage, and its documentation land as one self-contained change.

**Version stamping.** New `internal/version` package holding a bare-semver `Version` constant (`0.9.0-dev` between releases) — the single source of truth for the version baked into binaries and artefact names. `Full()` augments it with the commit hash, commit time, and dirty marker that the Go toolchain embeds automatically (no `-ldflags` needed). Both binaries gain cobra's `--version` flag:

```
$ openvox-ca --version
openvox-ca version 0.9.0-dev (commit efb27281a134, 2026-07-25T07:06:51Z, dirty)
```

`--version` is root-only on `openvox-ca-ctl` (it cannot follow a subcommand), and `--verbose` gains the `-v` shorthand so `-v` means verbose/verbosity on both binaries — without it, cobra would have claimed `-v` for the synthesised version flag on ctl only, giving the siblings opposite `-v` semantics. Both behaviours are pinned by specs.

**Versioned artefacts, built natively.** `mage build:dist` produces `openvox-ca_VERSION_linux_{amd64,arm64}[_fips].tar.gz` (with `-trimpath` for reproducibility), the FIPS arm64 variant being new — matching the amd64+arm64 coverage of the container images. In CI and the Release workflow, each variant builds via the new `mage build:distVariant` on a runner native to its architecture (arm64 on the free `ubuntu-24.04-arm` runners), so the cgo/boringcrypto builds need no cross toolchain and every binary — including both FIPS ones — is executed and its `--version` output checked; FIPS legs additionally assert boringcrypto via `go version -m`. Per-variant Go caches keep the legs to ~2 minutes and stay warm for tag builds.

**Release gates.** A shared `verify-release-tag` composite action gates both tag-triggered workflows (Release and Container images): the tag must equal `"v" + Version` — so cutting `vX.Y.Z` requires first landing the version bump — and the tagged commit must have a passing *CI success* check run, so a commit that never went green on `main` cannot be released. The Release workflow is now verify → build matrix → publish, generating `checksums.txt` across all variants in the aggregation job.

**Release presentation.** Tags with a semver pre-release suffix (`v0.9.0-rc1`) publish with `--prerelease` and are withheld from the `latest` container tags. `.github/release.yml` shapes the generated notes: renovate/dependabot PRs are excluded, the rest grouped by PR label.

**`mage release:prepare X.Y.Z`.** Automates the version-bump PR the gate requires: branches `release/vX.Y.Z` off the remote's main, rewrites the constant, pushes, and opens the PR with a preview of the auto-generated release notes. A `-dev` version produces the post-release bump PR instead. `OPENVOX_CA_RELEASE_REMOTE` targets a fork for rehearsals.

**Documentation.** `docs/development/releasing.md` documents the whole process — prepare PR, gates, tagging, notes, pre-release behaviour, the post-release `-dev` bump, and a fork-rehearsal procedure exercising the real flow end to end. The `--version` flag is added to the server and operator-CLI references. Remaining known gap, documented as such: artefact signing/attestation.

**FIPS trace for the new arm64 variant.** The crypto contract is unchanged — same `GOEXPERIMENT=boringcrypto` build path as the existing amd64 FIPS variant, no dependency changes. Verified on the built arm64 binary: `go version -m` reports `GOEXPERIMENT=boringcrypto`, the symbol table contains 475 `_goboringcrypto_*` symbols, and the binary runs on linux/arm64 (glibc); CI now re-verifies the boringcrypto build info on every PR. Like the amd64 FIPS variant, it is dynamically linked and will not run on musl.


trevor-vaughan-ai:

>
> Models used:
> - claude-sonnet-5

_Reviewed at commit [`89e49fe`](https://github.com/voxpupuli/openvox-ca/commit/89e49fe7926eaa64cba265e5a49ff045f50951f5) (standard effort)._

**TL;DR:** Review found five confirmed documentation and release-tooling gaps (four medium, one low) around a release/publishing PR; the testing persona requests changes over untested magefile logic.

**Findings:** 🔴 0 Critical, 🟠 0 High, 🟡 4 Medium, 🔵 1 Low

|---|---|---|
| 📚 Curator (code) | ✅ Approve | 1 MEDIUM |
| 🧪 Tester (code) | ❌ Changes | 1 MEDIUM |

- 📚 **One-time setup step describes a permissions model this PR removed** ([`docs/development/publishing-images.md:41`](https://github.com/voxpupuli/openvox-ca/blob/89e49fe7926eaa64cba265e5a49ff045f50951f5/docs/development/publishing-images.md#L41))

  ```
  The workflow already
     requests `packages: write`, but if an organization policy overrides this,
  ```

  💡 **Recommendation:** Update the sentence to reflect the per-job grant, e.g. "The build and merge jobs already request `packages: write`" (matching the new comment in container-images.yml), so the one-time setup instructions stay accurate for the org-policy-override case they describe.
  **Constraint:** Cross-reference consistency - documentation must describe the actual current behavior of the workflow it documents

  This PR changes `.github/workflows/container-images.yml`'s top-level `permissions:` block from `packages: write` to `checks: read` (with a new comment: "GITHUB_TOKEN needs packages:write to push to GHCR; it is granted per-job to the build and merge jobs that actually push"), and grants `packages: write` only inside the `build` and `merge` job blocks. `docs/development/publishing-images.md` step 1 ("Allow Actions to publish packages") still tells maintainers "The workflow already requests `packages: write`" at the workflow level, which is no longer true post-PR - the grant is now job-scoped. The PR updated other parts of this same file (the callout box and the release-tag row of the trigger table) but missed this section, which sits right next to the ones that were touched.
  </details>

- 🧭 **Release-machinery mage targets missing from AGENTS.md's canonical command table** ([`AGENTS.md:16`](https://github.com/voxpupuli/openvox-ca/blob/89e49fe7926eaa64cba265e5a49ff045f50951f5/AGENTS.md#L16))

  ```
  | `mage build:all` | Build `openvox-ca` and `openvox-ca-ctl` binaries |
  | `mage test:unit` | Run the unit suite (all packages, coverage to `coverage.out`) |
  | `mage dev:lint` | Run `golangci-lint` (gate; see `.golangci.yml`) |
  ```

  💡 **Recommendation:** Add rows for `mage build:dist`, `mage build:distVariant &lt;variant&gt;`, and `mage release:prepare &lt;version&gt;` to the Build/test/lint table in AGENTS.md (or add a pointer row to docs/development/releasing.md), so the authoritative conventions file stays in sync with the shipped mage surface.
  **Constraint:** Cross-Component

  AGENTS.md opens with "This file is authoritative for repository conventions" and its Build/test/lint table is the canonical listing of mage targets. This PR adds three new exported mage targets - `Build.Dist`, `Build.DistVariant`, and the new `Release` namespace's `Release.Prepare` (magefile.go, new `type Release mg.Namespace // release:prepare` and the `func (Build) Dist()`, `func (Build) DistVariant(name string)`, `func (Release) Prepare(ver string)` functions) - none of which appear in this table. `docs/development/releasing.md` (new) documents `mage build:dist`, `mage build:distVariant`, and `mage release:prepare` fully, but the authoritative AGENTS.md reference an agent or contributor would consult first for "what mage targets exist" is silent on all three, leaving it stale relative to the actual build system surface this PR ships.
  </details>

- ⚙️ **Release variant list hardcoded in four places with no automated cross-check** ([`.github/workflows/release.yml:99`](https://github.com/voxpupuli/openvox-ca/blob/89e49fe7926eaa64cba265e5a49ff045f50951f5/.github/workflows/release.yml#L99))

  ```
  for variant in linux_amd64 linux_arm64 linux_amd64_fips linux_arm64_fips; do
  ```

  💡 **Recommendation:** Derive the workflow matrices from `distVariants()` at run time (e.g., add a `mage build:distVariantsJSON` target that emits the variant names as a JSON array via `fmt.Println` and consume it via a `matrix: ${{ fromJSON(...) }}` setup job in ci.yml/release.yml, and reuse the same list in the release job's tarball-count check), so magefile.go remains the single source of truth instead of four independently hand-maintained lists.
  **Constraint:** Release artifacts complete for all declared target platforms (matrix completeness)

  The set of dist variants (linux_amd64, linux_arm64, linux_amd64_fips, linux_arm64_fips) is hardcoded independently in four places: this bash loop in release.yml's checksum-verification step, the `build` job's matrix a few lines above it (release.yml:36-43), the `dist` job's matrix in ci.yml (ci.yml:113-122), and `distVariants()` in magefile.go (magefile.go:239-253), which is the actual source of truth used by `mage build:dist`/`build:distVariant`. Confirmed via `grep -n "linux_amd64\|linux_arm64"` across all three files: each list is maintained by hand and only a code comment ("Keep this list in sync with the matrix above and distVariants() in magefile.go") enforces agreement. If a future variant is added to `distVariants()` but the workflow matrices aren't updated to match, CI's dist job and the Release build job simply never build or verify the new variant - no job fails, so the gap is silent. The release job's exact-4-tarball count check (release.yml:103) only catches the case where the count itself was also forgotten to be updated; it would happily pass 5 variants against a stale count-4 check only if raised in lockstep, and would not catch a 5th variant being silently dropped from the matrix while the bash loop's variant list still lists only 4 (both would agree with each other and both be wrong).
  </details>

- 🧪 **New release/build logic in magefile.go has zero Go test coverage** ([`magefile.go:201`](https://github.com/voxpupuli/openvox-ca/blob/89e49fe7926eaa64cba265e5a49ff045f50951f5/magefile.go#L201))

  ```
  func fipsCrossCC(goarch string) string {
  	if os.Getenv("CC") != "" {
  		return ""
  	}
  	if runtime.GOOS == "linux" && runtime.GOARCH == goarch {
  		return ""
  	}
  	switch goarch {
  	case "amd64":
  		return "x86_64-linux-gnu-gcc"
  	case "arm64":
  		return "aarch64-linux-gnu-gcc"
  	}
  	return ""
  }
  ```

  💡 **Recommendation:** Add a `magefile_test.go` (with the same `//go:build mage` tag so it only builds under `-tags mage`, matching how magefile.go itself is gated) with table-driven tests for `fipsCrossCC` (native vs cross-arch, `CC` env override), `bareSemverRe`/the validation branch of `Release.Prepare` (accept `0.9.0`, `0.9.0-rc1`, `0.10.0-dev`; reject `v0.9.0`, `0.9`, empty string), `repoSlug` (SSH and HTTPS remote forms, malformed URL), and `releaseVersion`/`distVariants`/`Build.DistVariant`'s unknown-variant error path. Wire `go test -tags mage ./...` (or an equivalent mage-tagged target) into CI so these run on every PR.
  **Constraint:** Coverage Completeness - untested code paths in core functionality (severity.md Tester/HIGH: "missing regression test"; lang-go.md testing_conventions expects `go test`-based coverage for deterministic logic)

  The PR's core deliverable - release-version stamping and the artefact/version machinery in magefile.go - ships with no Go unit tests at all: `releaseVersion()` (line 183), `fipsCrossCC()` (line 201), `distVariants()`/`buildDistVariant()` (lines 229/260), `repoSlug()` (line 390), `bareSemverRe` (line 386), `Release.Prepare()` (line 412), and `Build.DistVariant`'s "unknown dist variant" error branch (line 378) are all new, non-trivial, and deterministic (pure string/regex parsing plus a few `sh.*` calls), yet not one has a corresponding `_test.go`. I searched for any test referencing this file (`grep -rln "mage\b|magefile" --include="*_test.go" .` from the repo root) and found none targeting magefile.go. `fipsCrossCC` is the cleanest example: it is 100% pure and both branches are reachable from a single test process (verified locally: `fipsCrossCC("amd64")` on an amd64 host returns "", `fipsCrossCC("arm64")` falls through to the cross-compiler switch), yet CI's new `dist` job in `.github/workflows/ci.yml` only ever builds each variant on a runner *native* to that variant's architecture, so the cross-compile branch (`x86_64-linux-gnu-gcc` / `aarch64-linux-gnu-gcc`) is never exercised by CI either - a wrong compiler name here would go undetected by any automated process in the repository, contradicting the new `docs/development/releasing.md`'s guidance to "leave the FIPS builds to CI" for exactly this case. Similarly, `bareSemverRe`/`Release.Prepare` gates what version strings are allowed to open a release PR and rewrite `internal/version/version.go`, and `repoSlug` parses SSH vs HTTPS git remote URLs by regex - both are readily unit-testable pure-ish functions with meaningful edge cases (malformed versions, remotes with ports, non-GitHub hosts, trailing `.git`) and have none.
  </details>

</details>

- ⚙️ **verify-release-tag's documented prerequisites omit the contents:read it needs** ([`.github/actions/verify-release-tag/action.yml:9`](https://github.com/voxpupuli/openvox-ca/blob/89e49fe7926eaa64cba265e5a49ff045f50951f5/.github/actions/verify-release-tag/action.yml#L9))

  ```
    repository to be checked out and the calling workflow to grant
    checks: read.
  ```

  💡 **Recommendation:** Update the description to state both prerequisites: "Requires the repository to be checked out and the calling workflow to grant checks: read and contents: read."
  **Constraint:** Environment prerequisites explicit for operational documentation

  The composite action's own description states its only permission prerequisite is `checks: read` (plus checkout), but its second step also calls `gh api repos/$GITHUB_REPOSITORY/compare/$base...$sha`, which requires `contents: read` on the GITHUB_TOKEN (the GitHub compare-commits API reads repository contents). Both current callers (release.yml:11-13, container-images.yml:38-40) happen to grant `contents: read` already, so this doesn't fail today, but the action's stated contract is incomplete: a future caller that follows the documented prerequisites literally and grants only `checks: read` (recalling that setting any `permissions:` block sets all unlisted scopes to none) would get the compare API call failing with a 403 instead of a clear "missing permission" message pointing at the actual cause.
  </details>

</details>

---

bootc:
All five findings from the council review are addressed in 22645f3, and two further internal review-council rounds were run over the result (both returned APPROVE); their findings are fixed in 6bae7c2 and 1a41e6b.

**Finding 1 — publishing-images.md described the removed workflow-wide grant:** the one-time setup step now says the build and merge jobs request `packages: write` per-job.

**Finding 2 — release targets missing from AGENTS.md:** the mage table now lists `build:dist`, `build:distVariant <name>`, and `release:prepare <version>` (and, from the later rounds, `test:magefile`).

**Finding 3 — variant list hand-maintained in four places:** rather than dynamic `fromJSON` matrices (an extra setup job on every run, less readable workflows), the lists stay static and `mage dev:check` now enforces agreement: `verifyDistVariants` parses ci.yml's dist matrix, release.yml's build matrix, the checksum-step shell loop, and the tarball-count literal against `distVariants()`, failing loudly on any drift. Its failure branches are covered by negative specs (see finding 4).

**Finding 4 — magefile logic untested:** a build-tagged Ginkgo suite (`magefile_test.go`, run via the new `mage test:magefile` in CI's check job and the pre-push hook) covers `releaseVersion` (round-trips the real constant), `bareSemverRe` and `repoSlugFromURL` accept/reject tables, `fipsCrossCC` (exact cross-compiler names pinned — as the review noted, CI's native-runner builds never exercise them), `distVariants`' shape, `DistVariant`'s unknown-variant error, `Release.Prepare`'s validation guard, and the workflow parsers including every `verifyDistVariantsIn` failure branch.

**Finding 5 — action prerequisites incomplete:** verify-release-tag's description now names both `checks: read` and `contents: read`.

---

**Internal re-review rounds** (same five-persona council, run locally over the full branch):

*Round 1 — APPROVE, 11 verified findings, fixed in 6bae7c2. The substantive catches:*

- The CI-green gate counted check runs with `gh api --paginate --jq`, which applies the filter per page: past 100 check runs the output goes multi-line, `test`'s integer comparison errors, and the `if` treats that as false — the gate failed **open**. Now `--slurp`ed with an explicit non-numeric guard so it fails closed.
- The new workflow parsing had imported the archived `gopkg.in/yaml.v3` although the maintained `go.yaml.in/yaml/v3` fork is already a direct dependency used everywhere else; switched, and the archived module is back to indirect.
- The tagged magefile suite was CI-only; it now has the canonical `mage test:magefile` target, runs in the pre-push hook, and is documented in AGENTS.md, testing.md and CONTRIBUTING.md.
- Artefact verification was factored into a shared `verify-dist-artifact` action that the Release workflow's build job now also runs, so the exact binaries being published are executed (and FIPS variants boringcrypto-checked) before upload, not only their PR rehearsals.
- A validator pass proved the apparent dependency downgrades in the two-dot diff vs main are merge-base artefacts — the branch's only go.mod change is the yaml promotion above. The branch will be rebased onto main before merge so the forge diff sheds the stale pins.

*Round 2 — APPROVE, 6 findings, fixed in 1a41e6b. Fixes to the fixes:*

- The `fips:` matrix keys introduced in round 1 were themselves an unchecked hand-maintained duplicate: a missing or `false` key on a `_fips` variant would have silently skipped the only boringcrypto check. FIPS-ness is now derived from the `_fips` variant-name suffix inside the action (an invariant the `distVariants` spec pins) and the keys are deleted from both matrices.
- `Test.Magefile` had been inserted between `Test.Unit`'s doc comment and its declaration, misattributing both targets' `mage -l` descriptions; reordered so each keeps its own comment.
- Plus doc catch-ups (CONTRIBUTING checklist, lefthook header), a DescribeTable conversion per the repo's spec conventions, and a spec pinning that `Release.Prepare`'s bare-semver guard rejects before any side effect.

Not addressed here by design: the `--ca-cert` description fix lives in #143 (stacked on this PR) to avoid a conflicting duplicate edit.

bootc:
Thanks — all five findings are addressed, in `62b476b`, `e4dc60a` and `074dd42` (pushed; CI green). Two of them turned out to be worth more than their severity suggested.

**The four MEDIUMs and the LOW, as filed:**

- **`publishing-images.md` described a permissions model this branch removed.** The one-time GHCR setup step now describes the per-job `packages: write` grant rather than the workflow-wide one.
- **`AGENTS.md`'s command table was missing the release targets.** `build:dist`, `build:distVariant` and `release:prepare` added, plus `test:magefile`.
- **The variant list was hardcoded in four places with no cross-check.** `verifyDistVariants` is now split so its comparison logic takes the workflow contents as parameters, with specs driving each failure branch — drifted `ci.yml` matrix, drifted `release.yml` matrix, drifted shell loop, stale count literal — each asserting the error names the disagreeing location.
- **`magefile.go` had no Go test coverage.** It has a build-tagged suite now (~320 lines), plus a `test:magefile` target wired into CI's check job and the pre-push hook. Previously only CI ran it, so a magefile contributor passed every local gate and first saw failures remotely.
- **`verify-release-tag`'s documented prerequisites omitted `contents: read`.** Stated now, alongside `checks: read`, with the reason (the compare API call).

**Two things the follow-up rounds found that the original pass did not:**

**The CI-green release gate could fail open.** The check-run count used `gh api --paginate --jq`, which applies the filter *per page*. Past 100 check runs the count becomes multi-line, the integer test errors with status 2, and the `if` condition swallows that as false — so a commit that had never gone green could be released. The pages are slurped before filtering now, with an explicit integer guard so it fails closed even if the output shape regresses again.

**A FIPS variant could skip the only boringcrypto gate.** Artefact verification is extracted into a shared `verify-dist-artifact` action, used by CI's dist job and — newly — the Release workflow's build job, so the exact binaries being published are executed and FIPS-checked before upload rather than only their PR rehearsals. It derives FIPS-ness from the `_fips` variant-name suffix instead of a hand-maintained `fips` input duplicated across both matrices: a false or missing flag would have silently skipped the check, and the drift check never read those keys.

Also from the follow-ups: the magefile moved to `go.yaml.in/yaml/v3`, the maintained fork the rest of the codebase already uses, returning the archived `gopkg.in/yaml.v3` to an indirect dependency; `Test.Magefile` was moved below `Test.Unit` after the insertion landed between `Unit`'s doc comment and its declaration, which gave `test:magefile` `Unit`'s description in `mage -l` and left `test:unit` with none; and a spec pins that `release:prepare`'s bare-semver guard rejects before any side effect.

PR #145, opened 2026-07-25
Move test compose topologies under test/ and add a user-facing compose.yml

bootc:
All nine `compose*.yml` files at the repository root were integration-test topologies rather than deployment examples, and `Dockerfile.run` is the throwaway runtime image they build from locally-compiled binaries. Sitting at the top level, they looked like end-user entry points. This moves them all under `test/` and puts a genuinely user-facing `compose.yml` at the root in their place.

Three commits:

- **Move test compose topologies and Dockerfile.run under `test/`** — relocates the files and updates everything that referenced them: build contexts and bind mounts inside the compose files (these resolve relative to the compose file's directory), the mage targets, the test scripts (including the sed-rewriting `run-puppet-stack-on-redis.sh` wrapper and its substitution sanity checks), the hadolint CI matrix, and the docs.
- **Fix stale TLS examples in configuration and container docs** — found while testing: the documented `docker run` example refused to start as written, because the certificate-injection guard rejects plain HTTP on a non-loopback address. The example now passes `--tls-cert`/`--tls-key` using the CA's own certificate, which works from first boot. The config-file example also pointed at `ca_key.pem` in the cadir root; the key actually lives under `private/`.
- **Add a user-facing `compose.yml` at the repository root** — deployment-oriented compose file for the published GHCR images: named volume for the CA state, port 8140, a readiness healthcheck, and TLS served with the CA's own certificate. Referenced from the README and a new Compose section in `docs/container-images.md`.


trevor-vaughan-ai:
## 🟢 Review Council: APPROVE

>
> Models used:
> - claude-sonnet-5

_Reviewed at commit [`e7ad956`](https://github.com/voxpupuli/openvox-ca/commit/e7ad956023d48ece11a4807b0ab14e997bc4cccd) (standard effort)._

**TL;DR:** Clean review: all five personas approved; only two low/medium findings survived, both about the new root compose.yml's image pinning and healthcheck.

**Findings:** 🔴 0 Critical, 🟠 0 High, 🟡 1 Medium, 🔵 1 Low

|---|---|---|
| 🛡️ Adversary (code) | ✅ Approve | 1 LOW |

- ⚙️ **Healthcheck breaks for the alpine image variant the same block recommends** ([`compose.yml:31`](https://github.com/voxpupuli/openvox-ca/blob/e7ad956023d48ece11a4807b0ab14e997bc4cccd/compose.yml#L31))

  ```
      # NOTE: the healthcheck below uses curl, which the Alpine variant does
      # not ship.
      image: ghcr.io/voxpupuli/openvox-ca:latest
  ```

  💡 **Recommendation:** Either drop the curl-based healthcheck test in favor of one that works on both image variants (e.g. `wget -qO- --no-check-certificate` if busybox wget is present in the alpine image, or a Go-native check), or make the alpine incompatibility more prominent than a mid-file comment (e.g. a warning next to the `-alpine` tag mention in docs/container-images.md and README.md) so operators who pin to `-alpine` don't silently lose health status.
  **Constraint:** Healthcheck correctness for the deployment example (Operator criterion 5, runtime observability)

  The comment on lines 29-32 tells operators a `-alpine` variant of every tag is published as a legitimate pinning target, immediately above a healthcheck (`curl -skf https://localhost:8140/healthz/ready`, line 49) that requires curl. `Dockerfile.alpine` confirms the alpine runtime stage only installs `ca-certificates` (no curl), so swapping the image tag to any `-alpine` variant leaves `docker inspect`/`docker compose ps` reporting the service permanently unhealthy (curl: command not found, exit 127) even though the CA is serving correctly. The comment documents the incompatibility but the compose file takes no action to prevent or work around it (e.g. a wget-based check, since alpine images commonly ship busybox wget, or an explicit warning to not combine `-alpine` with this healthcheck as shipped).
  </details>

</details>

- 🛡️ **Deployment-example compose file pulls a floating, non-digest-pinned image tag** ([`compose.yml:33`](https://github.com/voxpupuli/openvox-ca/blob/e7ad956023d48ece11a4807b0ab14e997bc4cccd/compose.yml#L33))

  ```
  image: ghcr.io/voxpupuli/openvox-ca:latest
  ```

  💡 **Recommendation:** Either default compose.yml to a pinned digest/semver tag with a comment showing how to move to `:latest`, or keep the renovate carve-out but strengthen the in-file comment to state explicitly that `:latest` is unpinned by design and unsuitable for production use without operator action.
  **Constraint:** OWASP A06:2021 / NIST SSDF PW.4 (Verify Acquired Software) - supply-chain integrity of externally-sourced container images

  This PR repurposes the root compose.yml from a locally-built test topology into a real deployment example that pulls a prebuilt image from GHCR by mutable tag (`:latest`). The companion renovate.json change (new packageRule for `ghcr.io/voxpupuli/openvox-ca`, `"pinDigests": false`) deliberately opts this new externally-hosted dependency out of Renovate's `config:best-practices` default digest-pinning behavior, so tooling will never propose pinning it. A `:latest` reference gives no integrity guarantee that a `docker compose up`/`podman-compose up` on this file at two different times pulls the same bytes - if the registry is compromised or a bad image is force-pushed to the tag, operators following this compose file get it silently. The compose file's own comment (line 29) tells operators to "pin to a specific release tag... for reproducible deployments", which is good guidance, but the shipped example itself, and the tooling that would otherwise nudge toward pinning, both default to the floating tag.
  </details>

</details>

---

bootc:
Both findings from the Review Council run at e7ad956 are addressed in cdbba8c:

- **Healthcheck breaks on the `-alpine` variant (MEDIUM):** the healthcheck is now a `CMD-SHELL` with a busybox-wget fallback (`curl -skf ... || wget -q --no-check-certificate -O /dev/null ...`), so the same check works on both published variants: the default image ships curl, and the Alpine base ships busybox wget plus `ssl_client` for HTTPS. The now-obsolete "Alpine does not ship curl" caveat comment is dropped.
- **Floating `:latest` image reference (LOW):** the comment above `image:` now states explicitly that `:latest` is unpinned by design so the example tracks the newest release, that a floating tag carries no supply-chain integrity guarantee, and that production deployments should pin a release tag or digest. The Renovate carve-out is kept, per that same intent.

PR #150, opened 2026-07-25
Pass the signer PSK over an inherited pipe instead of the environment

### LINKED ISSUE #146: Signer PSK passed via inheritable environment variable risks IPC authentication disclosure
> Extracted from a report from my personal LLM/AI-based Threat Modeler project. Does not indicate allowance of a license to use said product.

### 🟡 OPENVOX-CA-20260725-044117-003: Signer PSK passed via inheritable environment variable risks IPC authentication disclosure

**Severity:** Medium &nbsp;|&nbsp; **Risk Score:** 8 (L2 × I4) &nbsp;|&nbsp; **Status:** Open &nbsp;|&nbsp; **Evidence:** Verified

#### System Context

- **Service:** launcher / signer IPC
- **Affected components:** cmd/openvox-ca/launcher.go:109, cmd/openvox-ca/launcher.go:125, cmd/openvox-ca/main.go:349
- **Attack surface:** Infrastructure

#### Evidence

- **Source:** semgrep; co-located scanner hit: scanner:semgrep
- **Rule / Check ID:** go.lang.security.audit.dangerous-exec-command.dangerous-exec-command
- **CVE ID:** N/A
- **Locations:**
  - `cmd/openvox-ca/launcher.go:109` — `exec.Command(...) spawn of signer/frontend child with PSK passed via PUPPET_CA_SIGNER_PSK env`
  - `cmd/openvox-ca/launcher.go:125` — `exec.Command(...) second child spawn inheriting socketpair fd:3 and PSK env`
  - `cmd/openvox-ca/main.go:349` — `exec.Command(...) role dispatch spawn flagged as non-static command`

#### Asset & Security Criteria

- **Business asset:** Isolation guarantee protecting the CA private key from the network-facing frontend
- **IS asset:** PSK authenticating the signer<->frontend socketpair channel
- **Criteria violated:** Confidentiality, Authentication

#### Misuser Profile

- **Actor:** Local attacker or compromised sibling process able to read /proc/<pid>/environ
- **Motivation:** Impersonate the frontend to the signer and obtain signing operations
- **Capability required:** Same-UID local code execution or a container escape into the process namespace

#### Preconditions (Vulnerabilities)

- Attacker has same-UID local execution
- PSK remains present in the child's environment after spawn
- Spawns re-exec the binary with role dispatch

#### Attack Path

**Primary:**
1. launcher generates a 32-byte PSK and passes it via PUPPET_CA_SIGNER_PSK to spawned children
2. Attacker with local same-UID access reads /proc/<child>/environ
3. Attacker connects to the signer socketpair (or a new signer connection) and replays the PSK handshake
4. Attacker submits signing-digest requests as if it were the frontend

**Alternative paths:**
- PSK leaked via a crash dump or process listing if it were ever placed on argv

#### Impact

**Likelihood:** Low — Requires local same-UID access which is a high bar; environ is not network-reachable and the socketpair fd is not shared

**Technical:** Bypass of the network-isolation partition protecting the CA signing key

#### Mitigations

**Preventive controls:**
- **Unset PUPPET_CA_SIGNER_PSK from the child environment immediately after handshake, or pass it over the inherited fd rather than env** *(effort: M)* → step 2: os.Unsetenv after reading in the child, or write PSK bytes over fd:3 before first RPC

**Detective controls:**
- **Assert single-connection semantics on the signer and log any additional handshake attempts** *(effort: M)* → step 3: Signer connection audit log; alert on >1 handshake per spawn

#### Compliance Mapping

- OWASP Top 10 A07:2021 — Identification and Authentication Failures: Secret material distributed via inheritable environment weakens the isolation design
- NIST 800-53 SC-3 — Security Function Isolation: PSK exposure could subvert the signer/frontend isolation boundary

bootc:
Fixes #146.

The launcher handed the socketpair PSK to the signer and frontend children via the `PUPPET_CA_SIGNER_PSK` environment variable. Although both children called `os.Unsetenv` after reading it, that only mutates the process's own copy: the exec-time environment stays visible in `/proc/<pid>/environ` for the child's whole lifetime, readable by any same-UID process (a `PTRACE_MODE_READ` check, which Yama `ptrace_scope=1` does **not** block, unlike memory dumping) and captured verbatim by crash-dump and support tooling such as systemd-coredump — the issue's "crash dump" alternative path.

This implements the issue's suggested preventive control, with a pipe rather than the socketpair itself (sending the PSK over the channel it authenticates would be circular):

- The launcher pre-loads a dedicated pipe per child with the hex PSK, closes the write end, and passes the read end as inherited fd 4 alongside the socketpair on fd 3. The child reads the PSK followed immediately by EOF; the secret never appears in any environment.
- The handshake is now **mandatory**: since the launcher always supplies the pipe, `loadPSK` failing (or the fd being absent) aborts startup instead of silently skipping authentication. This closes a downgrade path where a stripped environment would previously have disabled the handshake without any error.
- `PUPPET_CA_SIGNER_PSK` is still stripped from child environments defensively, but nothing reads it any more.

A review pass on the first commit produced four advisories, all addressed in the second commit:

- The handshake is now **mutual**. Previously only the signer verified the frontend, so a process holding a leaked fd could still impersonate the signer. Now the signer sends a nonce, the frontend replies with its own nonce plus an HMAC proof over both, and the signer returns a counter-proof the frontend verifies before use. Direction-specific labels domain-separate the proofs so neither can be replayed as the other.
- The cross-process fd contract (ExtraFiles position ↔ fd number) is pinned by an exec-based integration test that re-execs the test binary as real signer and frontend children via the production entry points, drives a signing RPC end to end, and asserts the fail-closed path when the fd 4 pipe is absent.
- Writing that test surfaced two robustness gaps, both fixed: socketpair fds were not close-on-exec (a child spawned while either end was open inherited both ends — masked in production only by the launcher's close-before-next-spawn ordering), and `loadPSK` now verifies fd 4 is actually a pipe before reading, so a process started outside the launcher gets a clear error instead of reading from an unrelated descriptor.
- The `parsePSK` specs assert which validation fires rather than that some error occurred, and the read-error branch is covered.

On the issue's detective control (single-connection semantics): the signer already serves exactly one pre-connected socketpair via `rpc.ServeConn` — there is no listening endpoint to accept further connections, and only one handshake can occur per spawn — so no change is needed there.

Exploitability context for reviewers: the PSK alone was never sufficient to reach the signer (the socketpair has no name; obtaining an endpoint requires `PTRACE_MODE_ATTACH`-level access, at which point the attacker can read the signer's memory or the on-disk key directly), so this is defence-in-depth hardening rather than a remotely exploitable fix.

FIPS/crypto note: no new dependencies and no new cryptographic primitives — the handshake still uses only HMAC-SHA256 (`crypto/hmac`, `crypto/sha256`) with nonces and the PSK from `crypto/rand`; the second commit changes the message flow (mutual proofs with domain-separation labels), not the primitives. The secret's transport from launcher to children is stdlib `os.Pipe`.

Verified end to end on a local run: both children log `PSK handshake succeeded` (one per direction), `ps eww` shows no PSK in any process environment, and SIGTERM shutdown remains clean.


trevor-vaughan-ai:

>
> Models used:
> - claude-sonnet-5

_Reviewed at commit [`8dde479`](https://github.com/voxpupuli/openvox-ca/commit/8dde479843e38cfd16bf87ffc79ffc4fdc36ee7e) (standard effort)._

**TL;DR:** Security and design got clean approvals; the council blocked only on missing test coverage for the new PSK-pipe and handshake code paths.

**Findings:** 🔴 0 Critical, 🟠 0 High, 🟡 4 Medium, 🔵 2 Low

|---|---|---|
| 📚 Curator (code) | ✅ Approve | 1 MEDIUM |
| 🧪 Tester (code) | ❌ Changes | 2 MEDIUM, 1 LOW |

- 📚 **CA key security doc omits the process-isolation/PSK-handshake mechanism this PR reworks** ([`docs/ca-key-security.md:3`](https://github.com/voxpupuli/openvox-ca/blob/8dde479843e38cfd16bf87ffc79ffc4fdc36ee7e/docs/ca-key-security.md#L3))

  ```
  This page covers how the CA private key is protected at rest, the options for
  moving key custody off the CA host entirely, and the detective control that
  watches for anomalous destructive operations.
  ```

  💡 **Recommendation:** Add a section to docs/ca-key-security.md (or a new docs/process-isolation.md linked from it and from README.md's Documentation list) describing the launcher/signer/frontend process split, the socketpair (fd 3) and PSK-pipe (fd 4) IPC contract, and the mutual handshake -- mirroring the rationale already written into the launcher.go and signer.go SECURITY comments. Since no Docs repo is configured for this project (no "Review Council Configuration" / Docs repo entry found in the invoking repository), this cannot be auto-filed as a tracked issue; flagging it here for manual triage instead.
  **Constraint:** docs/ca-key-security.md is the established, dedicated home for CA-key protection mechanisms (encryption at rest, OpenBao custody); it should cover other protection mechanisms in the same category rather than omitting them entirely.

  cmd/openvox-ca/launcher.go and internal/signer/signer.go substantially rework how the CA key is protected at runtime: the signer/frontend split now performs a mandatory mutual challenge-response handshake, and the PSK authenticating that handshake is delivered over an inherited pipe (fd 4) instead of an environment variable specifically to avoid /proc/<pid>/environ and crash-dump exposure (see the SECURITY comments at launcher.go:92-97 and signer.go:250-255). This is a genuine key-protection control -- keeping the CA key out of the frontend process's address space entirely -- yet docs/ca-key-security.md's declared scope (quoted above) covers only at-rest encryption and OpenBao custody, with no mention of the isolated-process architecture, the socketpair (fd 3), or the PSK handshake at all. I searched the full docs/ directory (grep -rn "PSK|PUPPET_CA_SIGNER_PSK|fd 3|fd 4|ExtraFiles|socketpair" --include=*.md) and README.md/AGENTS.md and found zero references anywhere -- this control has no operator-facing documentation at all, so an operator or security auditor reading the CA's key-security page (the natural place to look) has no way to learn this protection exists, verify its guarantees, or learn that fds 3 and 4 are reserved internally by the launcher for this purpose when running in daemon/isolated mode. Because no doc ever mentioned the old PUPPET_CA_SIGNER_PSK env var, nothing here is stale from this specific diff -- but the diff is a significant, well-commented security-hardening change to an entirely undocumented mechanism, which is the gap worth closing now rather than compounding further.
  </details>

- ⚙️ **PSK pipe read on fd 4 has no timeout, so a foreign inherited fd 4 can hang startup silently** ([`internal/signer/signer.go:283`](https://github.com/voxpupuli/openvox-ca/blob/8dde479843e38cfd16bf87ffc79ffc4fdc36ee7e/internal/signer/signer.go#L283))

  ```
  hexPSK, err := io.ReadAll(io.LimitReader(r, 2*pskLen+1))
  ```

  💡 **Recommendation:** Bound the fd 4 read with a deadline (e.g. f.SetReadDeadline(time.Now().Add(N)) before the parsePSK call) or a context-scoped read, and turn a timeout into the same actionable 'process not spawned by the launcher' error already used for the non-pipe case, so operators get a message instead of a silent hang.
  **Constraint:** Runtime Observability: long-running or blocking operations should provide progress feedback or fail with an actionable message rather than hang silently.

  loadPSK (signer.go:256-276) verifies fd 4 is a FIFO via Fstat, then hands it to parsePSK, which blocks on io.ReadAll(io.LimitReader(r, 65)) until either 65 bytes accumulate or the underlying reader returns EOF/error - LimitReader only synthesizes EOF once its own counter reaches zero, it does not shorten a blocking underlying Read. If fd 4 is not the launcher's PSK pipe but some other inherited FIFO whose write end is held open by an unrelated process (the launcher_test.go comment at lines 157-158 cites lefthook's pre-push hook and 'some CI runners' as real examples of wrappers that leak an fd at this exact number), the S_IFIFO check at signer.go:265 passes, and the read blocks indefinitely: no context, no deadline (no SetReadDeadline call anywhere in loadPSK/parsePSK), and no log line explaining why the signer or frontend process never becomes ready. An operator or supervisor sees only a hung process with zero diagnostic output, rather than the clear 'fd %d is not a pipe' error the same function already produces for the non-FIFO case (signer.go:266). The test suite works around exactly this by pinning fd 4 to /dev/null (a character device, never a FIFO) specifically so its own negative test doesn't hang -- that workaround is direct evidence the underlying blocking-read behavior it was written to dodge is real.
  </details>

- 🧪 **runLauncher's new spawn-failure cleanup branches are never exercised by a test** ([`cmd/openvox-ca/launcher.go:142`](https://github.com/voxpupuli/openvox-ca/blob/8dde479843e38cfd16bf87ffc79ffc4fdc36ee7e/cmd/openvox-ca/launcher.go#L142))

  ```
  	frontendPSK, err := pskPipe(pskHex)
  	if err != nil {
  		signerCmd.Process.Kill()
  		return err
  	}
  ```

  💡 **Recommendation:** Add a test that exercises runLauncher's failure/cleanup paths directly (or extracts the spawn-and-kill sequencing into a smaller unit that can be tested without a real os.Pipe fault), asserting that the previously-started signer process is actually waited on/reaped, not merely signaled, when a later spawn step fails.
  **Constraint:** Untested code paths in core functionality (severity.md HIGH calibration)

  This PR rewrites how runLauncher wires up the two children (PSK pipe creation, kill-on-partial-failure sequencing at launcher.go:120-123, 142-146, and 155-159), but `runLauncher` itself has zero test coverage - before or after this diff. `grep -rn "runLauncher" --include=*.go .` returns only its definition in launcher.go and its call site in main.go; no *_test.go file references it. The new test file (launcher_test.go) exercises `pskPipe` in isolation and the fd-recovery/handshake contract via a hand-rolled `runPSKChild` helper, but never calls `runLauncher` or reconstructs its child-spawn-and-kill sequence. Consequently the exact defect class this code guards against - e.g. an ordering bug where `signerCmd.Process.Kill()` at line 144 is invoked without ever being followed by `signerCmd.Wait()` in this failure branch, leaving a zombie signer process on every frontendPSK-creation failure - would not be caught by the test suite. The existing re-exec test infrastructure (TestMain dispatch, pskChildCmd) already demonstrates the pattern needed to drive this: a test could set a very low ulimit or a fault-injectable pskPipe to force the frontendPSK-creation-failure branch and assert both children are reaped and no fd survives.
  </details>

- 🧪 **loadPSK's fd-absent branch (the realistic non-launcher-spawn case) has no test** ([`internal/signer/signer.go:262`](https://github.com/voxpupuli/openvox-ca/blob/8dde479843e38cfd16bf87ffc79ffc4fdc36ee7e/internal/signer/signer.go#L262))

  ```
  	var st syscall.Stat_t
  	if err := syscall.Fstat(PSKFD, &st); err != nil {
  		return nil, fmt.Errorf("PSK pipe fd %d unavailable (process not spawned by the launcher?): %w", PSKFD, err)
  	}
  ```

  💡 **Recommendation:** Add a case to the "launcher fd contract" describe block that starts the frontend/signer child role with only the socketpair in ExtraFiles (fd 3) and nothing at fd 4, asserting the child fails with the "unavailable (process not spawned by the launcher?)" message rather than the S_IFIFO message.
  **Constraint:** Missing edge case / negative-path coverage for a security-relevant fail-closed guard

  loadPSK has two independent fail-closed branches: the Fstat-error branch (fd 4 closed/nonexistent, line 262-264) and the wrong-type branch (fd 4 open but not a FIFO, line 265-267). Only the second is tested: launcher_test.go's "fails closed when fd 4 is not the launcher's PSK pipe" spec (lines 163-183) opens /dev/null and passes it as ExtraFiles[1], which lands in the S_IFIFO mismatch branch, not the Fstat branch. `grep -rn "unavailable"` across the whole repo shows the error string built by the Fstat branch appears nowhere except its own definition - no test triggers it. This is the branch a real operator would actually hit: running `PUPPET_CA_ROLE=signer openvox-ca` by hand (i.e. the exact failure mode issue #146 and this PR's SECURITY comments are written to guard against) leaves fd 4 completely absent, giving EBADF from Fstat, not a wrong-type fd. That specific, more-likely path is unverified.
  </details>

</details>

- ⚙️ **New fd 3/fd 4 inheritance contract not documented for downstream integrators** ([`docs/openbao-transit.md:341`](https://github.com/voxpupuli/openvox-ca/blob/8dde479843e38cfd16bf87ffc79ffc4fdc36ee7e/docs/openbao-transit.md#L341))

  ```
  The isolated-signer deployment (the default; see the
  [OpenBao Transit-engine key custody](ca-key-security.md#openbao-transit-engine-key-custody)
  discussion) keeps working unchanged in OpenBao
  mode: the OpenBao client and its token live inside the isolated
  `openvox-ca [signer]` child process, exactly where a local private key lives
  today.
  ```

  💡 **Recommendation:** Add a short note to the Process isolation section (or docs/container-images.md) stating that openvox-ca reserves fds 3 and 4 for launcher-to-child IPC when running in daemon/isolated mode, and that wrapper scripts or supervisors must not leave an inherited pipe open at fd 4 when exec'ing it.
  **Constraint:** Operational Documentation: environment prerequisites and process-launch contracts that affect correct operation under supervision should be explicit for operators who wrap or supervise the binary.

  This is the only prose in the repo describing the isolated-signer child-process architecture (also referenced from docs/ca-key-security.md and docs/metrics.md:28), but it says nothing about the fd 3 (socketpair) / fd 4 (PSK pipe) reservation the launcher now depends on (internal/signer/signer.go:44,48, cmd/openvox-ca/launcher.go:129,152). Searched docs/*.md, README.md, Dockerfile, Dockerfile.run, and Dockerfile.alpine for 'fd 4', 'PSKFD', 'InheritedFD', and 'ExtraFiles' -- no hits anywhere in the docs tree. The shipped container image's exec-form ENTRYPOINT (Dockerfile.run) isn't affected, but any operator who wraps this binary in their own shell script, systemd unit, or process supervisor (s6, supervisord, runit) has no documented warning that fds 3 and 4 are reserved for launcher-child IPC and must not be pre-opened before exec. The launcher_test.go comment (lines 149-162) demonstrates the project's own engineers already hit this class of problem with lefthook and CI runners leaking an fd at position 4.
  </details>

- 🧪 **Success-path fd-contract spec has no cleanup guard if a mid-test Expect fails** ([`cmd/openvox-ca/launcher_test.go:136`](https://github.com/voxpupuli/openvox-ca/blob/8dde479843e38cfd16bf87ffc79ffc4fdc36ee7e/cmd/openvox-ca/launcher_test.go#L136))

  ```
  		Expect(signerCmd.Start()).To(Succeed(), "starting signer child")
  		signerSock.Close()
  		signerPipe.Close()
  		Expect(frontendCmd.Start()).To(Succeed(), "starting frontend child")
  		frontendSock.Close()
  		frontendPipe.Close()
  ```

  💡 **Recommendation:** Register `DeferCleanup` for `signerCmd.Process.Kill` immediately after `signerCmd.Start()` succeeds (and similarly for frontendCmd), matching the pattern already used in the "fails closed" spec, so cleanup does not depend on the rest of the spec body executing without failure.
  **Constraint:** Test isolation: resource cleanup should not depend on every prior assertion succeeding

  If `frontendCmd.Start()` fails, Ginkgo's fail handler aborts the spec via panic before `signerCmd.Wait()` is ever reached, leaving the already-started signer child unwaited. Unlike the sibling spec ("fails closed when fd 4 is not the launcher's PSK pipe", lines 163-183) which registers `DeferCleanup` for its socket and file handles up front, this spec relies on straight-line execution reaching the final `Wait()` calls to reap both children. The orphaned process is eventually killed by the `DeferCleanup(cancel)` context-timeout guard at cleanup time and reaped when the test binary itself exits, so this is not an observable hang, but it is an inconsistency with the more defensive sibling spec and a real (bounded) process leak within a single test run.
  </details>

</details>

---

bootc:
Thanks — all six addressed, plus three further council rounds. Pushed as `2cc4f32`, `d5082ff`, `1fe5432`, `cafdca6`, `4072742`; CI green. Every fix is mutation-verified.

**The six findings, as filed:**

- **The fd 4 read had no bound.** A foreign FIFO whose write end is held open passes the type check and never reaches EOF, and `io.LimitReader` does not shorten a blocking read — so the signer hung before logging anything. This is the one that grew; see below.
- **`runLauncher` only signalled the signer** when the frontend failed to start. `Kill` without `Wait` leaves a zombie for as long as the launcher lives. It reaps now, and the spawn-and-clean-up sequence is extracted into `spawnChild`/`killAndReap` so it can be tested at all.
- **Both of `loadPSK`'s refusals are covered** — and the way the review suggested reaching the absent-fd-4 branch does not work, which is the useful part: closing fd 4 in a child frees the slot and the Go runtime reuses it before `main` runs, so the check sees whatever took it. The validation is extracted as `checkPSKFD(fd)` and both branches are driven directly.
- **Cleanup no longer depends on straight-line execution**, and `docs/ca-key-security.md` now documents the process split, which process holds the key, the fd contract and the mutual handshake — the control that keeps the key out of the network-facing process was missing from a page whose stated scope was key protection.

**What the follow-up rounds found. Two of these are the reason I would not have shipped this after round 1:**

**The deadline I added never worked.** Four of five reviewers traced it to the same place: `exec.Cmd.Start` calls `File.Fd` on every `ExtraFiles` entry, and `Fd` clears `O_NONBLOCK` on the open file description the child's copy shares — so `os.NewFile` in the child returns a file whose `SetReadDeadline` can only return `os.ErrNoDeadline`, which I was tolerating. It bound only on an in-process `os.Pipe`, which is exactly what my spec used. Making the error fatal failed the end-to-end spec with *"file type does not support deadline"* on a perfectly correct spawn. The bound is a timer and a goroutine now, which depends on neither the descriptor's provenance nor the platform — setting the descriptor non-blocking again would also arm the deadline, but on Darwin kqueue does not reliably wake a reader when the last FIFO writer closes, which would risk delaying the *legitimate* start.

**Nothing pinned the property this branch is named after.** The impostor spec forges a *random* proof, which HMAC rejects whether or not the two directions are domain-separated. Unify the labels and the whole suite stays green — while an attacker holding a leaked socketpair descriptor and no PSK impersonates the signer by reflecting the frontend's own proof back, after which the frontend hands it CSR digests and accepts the results as CA signatures. There is a reflection spec now, and a table pinning all three `handshakeProof` inputs plus one asserting the challenge is fresh per handshake.

**An unrecognised `PUPPET_CA_ROLE` silently disabled key isolation.** No default in the dispatch: a non-empty value skips the launcher and a value that is not `frontend` skips the dial to the signer, so `PUPPET_CA_ROLE=fronted` served the API with the CA private key in its own address space, exit code 0, no warning. Pre-existing — but it is the doc *this branch added* that tells operators the fd rules matter "when you set `PUPPET_CA_ROLE` yourself", so the page asserted a guarantee the code did not enforce. Refused by name now.

**fd 3 had none of the discipline this branch built for fd 4.** `connFromFD` wrapped it unexamined and closed it *before* checking whether the conversion failed — so a role process whose own log file took fd 3 (likely, once `--log-file` is set) had that descriptor closed out from under it, and got a `getsockopt` error naming neither the launcher nor the contract. `checkSocketFD` is the sibling of `checkPSKFD` now, and the close happens only after the conversion succeeds.

**Three of my own specs could not fail.** When I added `sock.Close()` to the failure path the descriptor floor legitimately moved, and I relaxed `Equal(before)` to `<=` — the one direction the socket close already guarantees, so both leak specs passed with the PSK pipe leaked. They hold the descriptor and assert `Close()` returns `os.ErrClosed` now. Nothing asserted `Cmd.Env` either (dropping `PUPPET_CA_DAEMON=1` makes every child re-daemonise and exit, so the CA never starts), nor the `ExtraFiles` slice itself — swapping the two entries or dropping the pipe left the suite green and produced a CA whose children both refuse to start.

Also: the anti-aliasing clip needed a fixture with room for *both* appends before it proved anything; the wrapper-script paragraph I wrote stated the fd 4 rule backwards and attributed descriptor handling to `exec`; and this branch had orphaned an `os.Unsetenv` errcheck exclusion by deleting the only call it justified.

Issue #138, opened 2026-07-24
Decompose etcd inventory into per-entry keys (InventoryStore/CertIndex)


bootc:
## Summary

Make the **etcd** backend store the certificate inventory as decomposed per-entry keys (implementing `InventoryStore`, and later `CertIndex`) instead of the single append-only `inventory/data` blob it uses today.

Part of the pre-1.0.0 storage-interface review. Sibling issues: #137 (SQL cert index), #139 (Redis decomposition).

## Motivation

The current etcd inventory append is **read-the-entire-inventory, concatenate, write-the-entire-inventory**: [`AppendLine`](../blob/main/internal/storage/etcd.go) does `Get(whole blob)` then a ModRevision-guarded `Put(whole blob)`, plus a whole-blob HMAC recompute. That's **O(size-of-entire-inventory) per certificate issued** — O(n²) to issue n certs. At fleet scale (100k nodes, multi-MB inventory) every single signing rewrites megabytes.

Second gain: blob backends have **no cross-replica duplicate-serial guarantee** today — only SQL does, via its unique index (see the `ErrDuplicateSerial` comment in `storage.go`). Decomposition closes that gap for etcd.

## Feasibility — clean fit

etcd is a sorted MVCC store with multi-key transactions, which is essentially what `InventoryStore` wants:

- **Entries**: `inventory/entries/<zero-padded-seq>` → entry bytes. Range scan returns issuance order; a monotonic counter key allocates `seq`.
- **Serial uniqueness**: `inventory/by-serial/<serial>` marker key. `Txn.If(CreateRevision(serialKey) == 0).Then(...)` gives atomic, cross-replica "insert only if this serial is unused" — the guarantee currently exclusive to SQL.
- **Subject → latest serial**: `inventory/by-subject/<subject>` → serial, written in the same txn. O(1) instead of a full scan.
- **Hash-chain head**: one `inventory/hmac` key, advanced in the same txn as the entry insert, so entry + head never diverge across replicas.

**Accurate wrinkle:** `chainInventoryMAC` is a keyed hash we must compute in Go, and etcd txns are compare-then-op, not read-compute-write. So `AppendEntry` becomes the same optimistic loop the backend already runs today: read head → `newHead(prev)` in Go → `Txn.If(head unchanged AND by-serial absent).Then(put entry, head, subject-index, serial-marker)` → retry on conflict. It reuses the existing ModRevision-guard-plus-retry shape.

[...]

PR #151, opened 2026-07-25
Extend inventory table as certificate index (CertIndex capability)

### LINKED ISSUE #137: Extend inventory table as certificate index (add CertIndex capability)
## Summary

Extend the structured inventory table (`puppet_ca_inventory`) so it doubles as a **certificate index**, and add an optional `CertIndex` backend capability that lets SQL backends answer `certificate_status(es)` and expiry queries from indexed columns instead of the current *list-certs → per-subject `GetCert` → parse PEM → per-subject `IsRevoked`* scan.

This came out of an architecture review of the backend/front-end storage interface. The conclusion there was that the layering (`StorageService` domain layer over a pluggable blob `Backend`, with optional capability interfaces like `InventoryStore`/`Locker`) is sound and the blob key namespace is a genuine portability/Puppet-compat asset — **not** something to rip out. The one decision with a long half-life, worth making before 1.0.0 because it shapes the migrations, is: **is the inventory table our canonical certificate index?** This issue proposes committing to "yes".

## Motivation

[`handleGetStatuses`](../blob/main/internal/api/handlers.go) currently does, on every call:

- `ListCerts` → for each subject `GetCert` (read PEM) → X.509 parse → `IsRevoked` (CRL check).

That's O(N) blob reads + O(N) parses + O(N) revocation checks per request. For a fleet of tens of thousands of nodes this is a real scaling cliff, and `?state=` filtering still parses every PEM before discarding non-matches.

The certificate metadata we need is **immutable for the life of the cert** (subject, serial, validity, fingerprint, SANs, auth extensions never change once issued). The only mutable fact is revocation. So we can denormalise the display fields safely — they cannot drift from the PEM because the PEM cannot change either — keep the PEM as the authoritative artefact, and treat the row as a rebuildable projection of it.

## Proposed schema

Grow `puppet_ca_inventory` rather than adding a parallel `certificates` table (a second table would duplicate serial/subject/validity and create a two-way consistency burden). Inventory already *is* the issued-cert log.

```go
type sqlInventoryRow struct {
    bun.BaseModel `bun:"table:puppet_ca_inventory,alias:inv"`

    ID        int64  `bun:"id,pk,autoincrement"`               // issuance order + chain order (unchanged)
    Serial    string `bun:"serial,notnull,type:varchar(128)"`  // UNIQUE (unchanged)
    Subject   string `bun:"subject,notnull,type:varchar(512)"`
    NotBefore string `bun:"not_before,notnull,type:varchar(64)"`
    NotAfter  string `bun:"not_after,notnull,type:varchar(64)"`

    // new: denormalised display projection (immutable, filled at signing)
    Fingerprint string `bun:"fingerprint_sha256,type:varchar(95)"`
    DNSAltNames string `bun:"dns_alt_names,type:text"`   // JSON array, or NULL
    AuthExts    string `bun:"auth_extensions,type:text"`  // JSON object, or NULL

    // new: the one mutable bit
    State     string     `bun:"state,notnull,type:varchar(16)"`  // "signed" | "revoked"
    RevokedAt *time.Time `bun:"revoked_at"`                       // NULL unless revoked
}
```

Indices: keep `UNIQUE(serial)`; add `INDEX(subject)`, `INDEX(state)`, `INDEX(not_after)`.

## Proposed capability interface

Fold into `InventoryStore` (or a sibling it embeds) — the backend that owns the inventory rows is the one that can answer these:

```go
type CertRecord struct {
    InventoryEntry                    // Serial, Subject, NotBefore, NotAfter
    State       string                // "signed" | "revoked"
    Fingerprint string
    DNSAltNames []string
    AuthExts    map[string]string
}

type CertIndex interface {
    Statuses(ctx context.Context, stateFilter string) ([]CertRecord, error) // "" == all
    SetRevoked(ctx context.Context, serial string, at time.Time) error
}
```

`handleGetStatuses` probes for the capability (exactly like `asInventoryStore`) and collapses to a single indexed query, keeping the existing `ListCerts`/parse path verbatim as the fallback for blob backends. Pending CSRs (`requested`) aren't issued certs and stay on the `csr/` list-and-parse path — that population is small and short-lived.

## Open questions / decisions for @trevor-vaughan

1. **Denormalise the display fields (fingerprint/SANs/auth) or not?** If we don't, the list path still parses one PEM per row for those fields — we've cheapened filtering/counting but not the full payload. Since the fields are immutable I lean towards denormalising (cost: wider row + one-time backfill). Agree?
2. **Is `state`/`revoked_at` authoritative or a cache of the CRL?** The signed CRL remains the source of truth; I'd treat these columns as a projection the revoke path writes transactionally alongside the CRL update, rebuildable from the CRL during repair/migrate — the same relationship the row has to the PEM. Objections?
3. Anything about this shape that would bite the FIPS/boringcrypto contract or the existing migration framework that I'm not seeing?

## Scope / notes

- Non-breaking: it's behind `StorageService` and a capability probe, so it changes no REST contract and can land independently.
- The `CertRecord` field set should be **fixed once** and shared, because the etcd/redis inventory-decomposition issues (#138 / #139) would project the same fields. Let's settle this struct before implementing any of the three.

Part of the pre-1.0.0 storage-interface review. Sibling issues: #138 (etcd decomposition), #139 (Redis decomposition).

bootc:
Implements #137: the structured inventory table (`puppet_ca_inventory`) now doubles as a certificate index, and SQL backends advertise a new optional `CertIndex` capability that lets `GET /certificate_statuses` collapse from an O(N) *list-certs → read-PEM → parse → CRL-check* scan into indexed queries. Blob backends (filesystem, etcd, Redis) keep the existing scan path verbatim via the same capability-probe pattern as `InventoryStore`/`Locker` — no REST contract changes anywhere.

Following the leanings in the issue's open questions:

1. **Display fields are denormalised** (`fingerprint_sha256`, `dns_alt_names`, `auth_extensions`, filled at signing). They are immutable for the life of the cert, so they cannot drift from the PEM; the PEM stays the authoritative artefact.
2. **`state`/`revoked_at` are a projection of the signed CRL**, written by the revoke path right after the CRL is re-signed (log-and-continue on failure, including the already-revoked retry branch). The CRL remains the source of truth.
3. **Rebuildability is enforced, not just claimed**: CA startup runs an index repair pass that backfills projections from stored PEMs (refusing serial mismatches) and reconciles revocation state against the CRL in both directions. This is also what converges an index populated by `storage migrate` from a blob backend, whose imported rows arrive projection-less; until repair runs, the handler falls back to the PEM per projection-less record.

Design points worth reviewer attention:

- **Integrity chain coverage is unchanged.** The hash chain still covers only the canonical inventory line. Chaining the new columns would add nothing: each one is independently verifiable against a signed artefact (PEM or CRL).
- **`Statuses` is gated on blob existence and returns only the latest issuance per subject**, so cleaned or superseded historical rows are never listed and the output matches what the scan path would have produced. The API suite pins the index path to values derived from the stored PEM — the exact artefact the fallback would have parsed.
- **Migration hygiene:** the 20260201 migration is frozen onto a v1 copy of the row model so fresh databases create the original shape and replay the new ALTERs, identically to databases upgraded in place. The down migration avoids bun's `DropIndexQuery`, which emits bare `DROP INDEX` (invalid on MySQL). Both directions plus a full index round-trip run against SQLite, PostgreSQL and MySQL in the unit/integration suites.
- **Shared field producers:** the colon-separated SHA-256 fingerprint and the auth-extension display map now have single producers (`ca.SHA256ColonFingerprint`, `ca.AuthExtensionMap`) used by both the status API and the index projection, so the two cannot disagree on format.
- **`CertRecord` is the shared struct** the etcd/Redis decompositions (#138/#139) would project into, per the issue's note to settle it once.

**FIPS/crypto note:** no new dependencies; the only crypto touched is `crypto/sha256` and `crypto/x509` from the standard library, on paths that already used them (fingerprinting and cert parsing), so the boringcrypto contract is unaffected.

Closes #137.

### DIFF
--- a/docs/development/inventory-store.md
+++ b/docs/development/inventory-store.md
@@ -61,12 +61,13 @@ type InventoryEntry struct {
 // render → scan → reparse round-trip. Backends that do not implement it keep
 // using the KeyInventory blob via AppendLine/Get.
 type InventoryStore interface {
-    // AppendEntry inserts e and advances the integrity head atomically.
+    // AppendEntry inserts rec and advances the integrity head atomically.
     // newHead computes the chained head MAC from the previous head (nil when
     // the inventory is empty); the backend MUST invoke it inside the same
     // transaction/lock that serialises appends so the chain cannot fork under
-    // concurrent appenders.
-    AppendEntry(ctx context.Context, e InventoryEntry, newHead func(prev []byte) []byte) error
+    // concurrent appenders. rec is a CertRecord (see "The certificate index"
+    // below); only its canonical InventoryEntry fields feed the chain.
+    AppendEntry(ctx context.Context, rec CertRecord, newHead func(prev []byte) []byte) error
 
     // Entries returns every entry in issuance order, for the OCSP index build
     // and for chain verification.
@@ -136,6 +137,51 @@ blob HMAC ≠ a chain head over the same entries). So after the copy,
 recomputes the destination's integrity head from its entries
 (`RebuildInventoryHMAC`). This resolves the otherwise-spurious
 `ErrInventoryTampered` that copying a foreign `inventory_hmac` would cause.
+
+### The certificate index (`CertIndex`)
+
+The inventory table doubles as a **certificate index**: alongside the four
+canonical columns, each row carries a denormalised display projection
+(`fingerprint_sha256`, `dns_alt_names`, `auth_extensions`, filled at signing)
+and the one mutable fact, revocation (`state`, `revoked_at`). Backends owning
+the rows may advertise a second optional capability:
+
+```go
+type CertIndex interface {
+    // One record per subject with a stored certificate (latest issuance),
+    // optionally filtered by state ("signed"/"revoked").
+    Statuses(ctx context.Context, stateFilter string) ([]CertRecord, error)
+    SetRevoked(ctx context.Context, serial string, at time.Time) error
+    ClearRevoked(ctx context.Context, serial string) error
+    SetProjection(ctx context.Context, serial string, proj CertProjection) error
+}
+```
+
+`GET /certificate_statuses` probes it (via `StorageService.CertStatuses`,
+mirroring `asInventoryStore`) and collapses from an O(N) *list-certs →
+read-PEM → parse → CRL-check* scan to indexed queries; blob backends keep the
+scan path verbatim. Pending CSRs are not issued certificates and stay on the
+`csr/` list-and-parse path.
+
+Design rules that keep the index honest:
+
+- **The PEM and the signed CRL stay authoritative.** The projection columns
+  are immutable copies of fields fixed at signing (so they cannot drift); the
+  revocation columns are written by the revoke path right after the CRL is
+  re-signed. Every column is rebuildable from the artefacts.
+- **The hash chain covers only the canonical columns.** Chaining the
+  projection would add nothing: each projected field is independently
+  verifiable against a signed artefact.
+- **Blob-imported rows carry no projection** (`fingerprint_sha256` NULL). The
+  CA runs an index repair pass at startup (`rebuildCertIndex`) that backfills
+  projections from the stored PEMs and reconciles `state` against the CRL —
+  this is what makes a `storage migrate` from a blob backend converge. Until
+  repair runs, the statuses handler falls back to the PEM per projection-less
+  record.
+- **Statuses is gated on blob existence.** Historical rows survive cleaning
+  (that is the inventory's job), so the index reports only subjects whose
+  `cert/<subject>` blob still exists, and only their latest issuance —
+  matching what the scan path would have listed.
 
 ## Scope
 
--- a/docs/development/storage-internals.md
+++ b/docs/development/storage-internals.md
@@ -157,10 +157,55 @@ Most logical keys live in a key-value table, `puppet_ca_blobs`:
 
 Migrations are managed by [bun](https://bun.uptrace.dev/)'s migrator. On every
 start the backend runs any pending migrations and records applied versions in
-its own `bun_migrations` table; a `bun_migration_locks` table serialises
-concurrent runners so multiple replicas can start against the same database
-safely. Migrations are defined as Go functions, so one definition emits
-dialect-correct DDL across SQLite, PostgreSQL, and MySQL/MariaDB.
+its own `bun_migrations` table. Migrations are defined as Go functions, so one
+definition emits dialect-correct DDL across SQLite, PostgreSQL, and
+MySQL/MariaDB.
+
+bun does **not** serialise concurrent migration runners on its own. It creates a
+`bun_migration_locks` table and exposes `Migrator.Lock`/`Unlock`, but `Migrate`
+never calls them, and it records a migration as applied *before* running it. Two
+processes starting together — separate replicas, or the signer and frontend of
+one replica — will therefore both see a version as unapplied, both record it,
+and both run its DDL; the loser fails on work the winner already did, and the
+version stays recorded and is never retried. `EnsureReady` closes that hole with
+four measures:
+
+- It holds the backend's own distributed lock (`sql-schema-migrate`) across the
+  whole run, taken before any migration state is read.
+- It passes `WithMarkAppliedOnSuccess`, so a failed migration is retried on the
+  next start rather than remembered as done.
+- Each migration's DDL runs in a transaction where the dialect has transactional
+  DDL (PostgreSQL and SQLite), and every statement is idempotent, so MySQL —
+  which commits implicitly on each DDL statement — recovers by re-running.
+- The run gets its own timeout rather than the per-statement request timeout, so
+  a slow index build cannot be cut off part-way.
+
+SQLite has no distributed lock and falls back to a process-local mutex. Two
+processes sharing one file can still race there; the transactional, idempotent
+migrations mean the loser fails cleanly and succeeds on its next start rather
+than leaving the schema half-changed.
+
+#### Recovering a half-migrated schema
+
+A database migrated by a build without those measures can hold a version in
+`bun_migrations` whose DDL only partly ran — typically visible as a startup
+failure naming a column that does not exist, reported as an inventory integrity
+failure because the integrity check is the first thing to read the table.
+Confirm it by listing the table's columns and the recorded versions:
+
+```sql
+SELECT column_name FROM information_schema.columns
+ WHERE table_name = 'puppet_ca_inventory' ORDER BY ordinal_position;
+SELECT id, name, group_id, migrated_at FROM bun_migrations ORDER BY id;
+```
+
+Two rows for one version, recorded milliseconds apart, are the fingerprint of
+two runners having raced. To recover, stop every replica, delete the duplicate
+rows for the affected version, and start one replica: migrations are idempotent,
+so the run completes the missing DDL and leaves the already-applied statements
+alone. Applying the remaining DDL by hand and leaving the rows in place works
+equally well and is the safer choice on a large table, where the missing
+statement may be a long index build you would rather schedule.
 
 On MySQL/MariaDB, the first-run migration widens the `data` column to `LONGBLOB`
 (MySQL's default `BLOB` caps at 64 KiB, too small for large blobs such as the
@@ -178,6 +223,16 @@ indexed by subject:
 | `serial` | certificate serial (unique) |
 | `subject` | certificate subject (indexed) |
 | `not_before` / `not_after` | validity window, stored as the inventory.txt strings |
+| `fingerprint_sha256` | SHA-256 fingerprint; `NULL` means "no projection, read the PEM" |
+| `dns_alt_names` | subject alternative names, JSON array; `NULL` when empty |
+| `auth_extensions` | Puppet auth extensions, JSON object; `NULL` when empty |
+| `state` | `signed` or `revoked`, projected from the signed CRL (indexed) |
+| `revoked_at` | revocation time; `NULL` unless revoked |
+
+The last five columns make the table double as the certificate index; see
+[the inventory store](inventory-store.md) for what reads them and how a missing
+projection falls back to the stored PEM. `not_after` is indexed alongside
+`state` for consumers that have not landed yet (see the migration's own note).
 
 This turns appends and revocation lookups (`LatestSerialForSubject`) into
 single-row operations instead of scanning the whole inventory. Integrity uses a
--- a/docs/storage-backends.md
+++ b/docs/storage-backends.md
@@ -314,7 +314,7 @@ cadir: /var/lib/puppet-ca                # still needed for per-subject keys
 sql_dsn: "postgres://puppetca:secret@db.example.com:5432/puppetca?sslmode=require"
 
 sql_request_timeout_sec: 10              # per-operation timeout (default 10)
-sql_max_open_conns: 0                    # 0 = database/sql default
+sql_max_open_conns: 0                    # 0 = database/sql default; min 4 when set
 sql_max_idle_conns: 0                    # 0 = database/sql default
 
 # Optional mTLS to PostgreSQL (alternative to sslmode/ssl params in the DSN).
@@ -386,6 +386,16 @@ The SQL backends share one set of config keys and environment variables:
 
 The pool-tuning and TLS settings apply only to the networked SQL dialects;
 SQLite ignores them.
+
+`sql_max_open_conns`, when set to a non-zero value below 4, is raised to 4 and
+the effective value is logged at startup. The distributed locks are
+session-scoped (PostgreSQL advisory locks and MySQL `GET_LOCK` both bind to
+their connection), so each lock held occupies a connection for its lifetime
+while the work under it needs another. A `migrate` run nests three deep — the
+bootstrap lock, the migration lock inside it, and the migration's own
+statements. A smaller pool would not merely be slow, it would deadlock, so the
+floor is not negotiable. Leave the setting at 0 unless you have measured a
+reason to cap it.
 
 ---
 
--- a/docs/storage-backends.md
+++ b/docs/storage-backends.md
@@ -263,6 +263,13 @@ backend creates and upgrades its own tables automatically on startup, so
 multiple replicas can start against the same database safely. `cadir` is still
 required for per-subject keys and ancillary local state.
 
+SQL backends additionally maintain a certificate index: `GET
+/certificate_statuses` (`puppetserver ca list`) is answered from indexed
+columns instead of reading and parsing every stored certificate, which matters
+for large fleets. The index is a rebuildable projection of the stored
+certificates and the CRL — after migrating from another backend it is
+backfilled automatically on the next server start.
+
 ### SQLite backend
 
 Stores the entire CA in one SQLite database file — a convenient, dependency-free
--- a/internal/api/certindex_statuses_test.go
+++ b/internal/api/certindex_statuses_test.go
@@ -1,0 +1,203 @@
+// Copyright (C) 2026 Chris Boot
+// Copyright (C) 2026 Vox Pupuli and contributors
+//
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 2 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+package api_test
+
+import (
+	"bytes"
+	"context"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/json"
+	"encoding/pem"
+	"net/http"
+	"net/http/httptest"
+	"path/filepath"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/voxpupuli/openvox-ca/internal/api"
+	"github.com/voxpupuli/openvox-ca/internal/ca"
+	"github.com/voxpupuli/openvox-ca/internal/storage"
+	"github.com/voxpupuli/openvox-ca/internal/testutil"
+)
+
+// generateCSRWithSANs builds a CSR carrying explicit DNS subject alternative
+// names, which the signing path copies onto the issued certificate and the
+// certificate index projects into dns_alt_names.
+func generateCSRWithSANs(commonName string, dnsNames []string) []byte {
+	key, err := rsa.GenerateKey(rand.Reader, 2048)
+	Expect(err).NotTo(HaveOccurred())
+	template := &x509.CertificateRequest{
+		Subject:            pkix.Name{CommonName: commonName},
+		DNSNames:           dnsNames,
+		SignatureAlgorithm: x509.SHA256WithRSA,
+	}
+	csrBytes, err := x509.CreateCertificateRequest(rand.Reader, template, key)
+	Expect(err).NotTo(HaveOccurred())
+	return pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE REQUEST", Bytes: csrBytes})
+}
+
+// This suite runs the certificate_statuses endpoint against a SQL backend, so
+// the handler serves from the certificate index instead of walking the stored
+// PEMs. The assertions derive every expected value from the stored PEM — the
+// authoritative artefact the fallback path would have parsed — pinning the
+// index path to byte-identical output.
+var _ = Describe("Certificate statuses via the certificate index", func() {
+	var (
+		ctx   = context.Background()
+		store *storage.StorageService
+		myCA  *ca.CA
+		mux   http.Handler
+	)
+
+	BeforeEach(func() {
+		dir := GinkgoT().TempDir()
+		backend, err := storage.NewSQLBackend(storage.SQLConfig{
+			Dialect: storage.SQLitePure,
+			DSN:     "file:" + filepath.Join(dir, "ca.db"),
+		})
+		Expect(err).NotTo(HaveOccurred())
+		DeferCleanup(func() { _ = backend.Close() })
+		store = storage.NewWithBackend(backend, dir)
+		myCA = ca.New(store, ca.AutosignConfig{Mode: "off"}, "puppet.test")
+
+		Expect(store.EnsureDirs(ctx)).To(Succeed())
+		Expect(store.SaveCAKey(ctx, cachedKeyPEM)).To(Succeed())
+		Expect(store.SaveCACert(ctx, cachedCrtPEM)).To(Succeed())
+		Expect(store.UpdateCRL(ctx, cachedCrlPEM)).To(Succeed())
+		Expect(store.WriteSerial(ctx, "0001")).To(Succeed())
+		Expect(store.TouchInventory(ctx)).To(Succeed())
+		Expect(myCA.Init(ctx)).To(Succeed())
+
+		mux = api.New(myCA).Routes()
+	})
+
+	submitAndSign := func(subject string, csrPEM []byte) {
+		rr := httptest.NewRecorder()
+		mux.ServeHTTP(rr, httptest.NewRequest("PUT", "/certificate_request/"+subject, bytes.NewReader(csrPEM)))
+		Expect(rr.Code).To(Equal(http.StatusOK), "PUT certificate_request/"+subject)
+		body, err := json.Marshal(api.PutStatusBody{DesiredState: "signed"})
+		Expect(err).NotTo(HaveOccurred())
+		rr = httptest.NewRecorder()
+		mux.ServeHTTP(rr, httptest.NewRequest("PUT", "/certificate_status/"+subject, bytes.NewReader(body)))
+		Expect(rr.Code).To(Equal(http.StatusNoContent), "PUT certificate_status/"+subject)
+	}
+
+	getStatuses := func(query string) []api.CertStatusResponse {
+		rr := httptest.NewRecorder()
+		mux.ServeHTTP(rr, httptest.NewRequest("GET", "/certificate_statuses/any"+query, nil))
+		Expect(rr.Code).To(Equal(http.StatusOK), "GET certificate_statuses"+query)
+		var statuses []api.CertStatusResponse
+		Expect(json.Unmarshal(rr.Body.Bytes(), &statuses)).To(Succeed())
+		return statuses
+	}
+
+	It("serves indexed statuses identical to what the stored PEM would produce", func() {
+		sans := []string{"idx-node.example.com", "idx-node.alt.example.com"}
+		submitAndSign("idx-node", generateCSRWithSANs("idx-node", sans))
+
+		statuses := getStatuses("")
+		Expect(statuses).To(HaveLen(1))
+		got := statuses[0]
+
+		// Derive the expected response from the authoritative PEM, exactly as
+		// the non-indexed fallback path would.
+		certPEM, err := store.GetCert(ctx, "idx-node")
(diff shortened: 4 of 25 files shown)


trevor-vaughan-ai:

>
> Models used:
> - claude-sonnet-5

_Reviewed at commit [`a74ab48`](https://github.com/voxpupuli/openvox-ca/commit/a74ab482c36f55fc41b7cc9b8aeca739abc166b0) (standard effort)._

**TL;DR:** The council reviewed PR #151 (voxpupuli/openvox-ca), which grows the certificate inventory table into a CertIndex capability for indexed certificate queries; the council verdict is REQUEST CHANGES, driven by scope-bundling and missing tests around safety-critical code paths.

**Findings:** 🔴 0 Critical, 🟠 2 High, 🟡 8 Medium, 🔵 4 Low

|---|---|---|
| 📚 Curator (code) | ✅ Approve | 1 MEDIUM, 2 LOW |
| 🧭 Guard (code) | ❌ Changes | 1 HIGH |
| 🧪 Tester (code) | ❌ Changes | 1 HIGH, 6 MEDIUM, 1 LOW |

- 🧭 **Migration-locking correctness fix bundled into the CertIndex PR** ([`docs/storage-backends.md:390`](https://github.com/voxpupuli/openvox-ca/blob/a74ab482c36f55fc41b7cc9b8aeca739abc166b0/docs/storage-backends.md#L390))

  ```
  `sql_max_open_conns`, when set to a non-zero value below 4, is raised to 4 and
  the effective value is logged at startup.
  ```

  💡 **Recommendation:** Confirm with the author whether the EnsureReady locking fix was pre-existing behavior being documented, or new code introduced by this PR. If new, split it into its own PR/commit with its own rationale and tests, separate from the CertIndex feature, so each can be reviewed and reverted independently.
  **Constraint:** Intent Drift (scope creep)

  Issue #137 (linked issue for this PR) scopes the change to "growing the puppet_ca_inventory table into a certificate index" - a denormalised display projection plus revocation state, explicitly required to be non-breaking behind a capability probe. This batch's docs additions go well beyond that: docs/development/storage-internals.md's diff replaces the previously-true claim "a bun_migration_locks table serialises concurrent runners so multiple replicas can start against the same database safely" with a correction that bun's Migrate never actually serialises runners, and documents four new EnsureReady measures (distributed lock, WithMarkAppliedOnSuccess, transactional/idempotent DDL, dedicated timeout) plus a new "Recovering a half-migrated schema" runbook. docs/storage-backends.md then documents a behavioral change: sql_max_open_conns values 1-3 are now silently overridden to 4 to avoid a lock-nesting deadlock in that same migration path. None of this is about the certificate index's display projection or revocation columns - it is a general-purpose SQL migration race-condition fix, unrelated to CertIndex, bundled into the same PR without being called out in the issue or the PR description excerpt provided. This inflates the changeset's review surface and blast radius beyond the stated scope.
  </details>

- 🧪 **backfillCertProjection's serial-mismatch safety guard is never exercised by a test** ([`internal/ca/certindex.go:186`](https://github.com/voxpupuli/openvox-ca/blob/a74ab482c36f55fc41b7cc9b8aeca739abc166b0/internal/ca/certindex.go#L186))

  ```
  	if recSerial.Cmp(cert.SerialNumber) != 0 {
  		slog.Warn("Certificate index repair: stored certificate serial does not match index record, leaving projection empty",
  ```

  💡 **Recommendation:** Add a case to the "on a SQL backend" or migration Context in internal/ca/certindex_test.go that leaves an index row with a serial that does not match the subject's stored PEM (e.g. via store.AppendInventory with a mismatched serial, or by importing then overwriting the stored cert directly), runs rebuildCertIndex (via Init on a fresh ca.CA), and asserts the record's Fingerprint stays empty rather than being backfilled from the mismatched PEM.
  **Constraint:** Missing edge case coverage for boundary-sensitive code / untested code path in core functionality (severity.md, tester calibration table)

  backfillCertProjection's own doc comment names this branch as protection against a real, expected operational scenario ("e.g. a crash window between blob and inventory writes") -- it refuses to stamp a subject's index row with another certificate's projection when the stored PEM's serial disagrees with the row's serial. A repo-wide search (`grep -rn "does not describe the stored certificate\|leaving projection empty\|serial does not match" --include=*_test.go`) turns up no matches: neither internal/ca/certindex_test.go nor internal/storage/sql_certindex_test.go exercises this path. The two certindex_test.go scenarios (revocation reconciliation, migration backfill) only cover the case where the record's serial already matches the stored PEM. A regression here (e.g. the Cmp check silently dropped or inverted) would go undetected and would let the repair pass overwrite a subject's fingerprint/SANs/auth-extensions with a different certificate's data -- exactly the corruption this guard exists to prevent.
  </details>

</details>

<details><summary>🟡 MEDIUM (8)</summary>

- 📚 **CertIndex design doc omits the StorageService method-name mapping its own convention establishes** ([`docs/development/inventory-store.md:160`](https://github.com/voxpupuli/openvox-ca/blob/a74ab482c36f55fc41b7cc9b8aeca739abc166b0/docs/development/inventory-store.md#L160))

  ```
  `GET /certificate_statuses` probes it (via `StorageService.CertStatuses`,
  mirroring `asInventoryStore`) and collapses from an O(N) *list-certs →
  read-PEM → parse → CRL-check* scan to indexed queries; blob backends keep the
  scan path verbatim. Pending CSRs are not issued certificates and stay on the
  `csr/` list-and-parse path.
  ```

  💡 **Recommendation:** Add a bullet list after the CertIndex interface block mirroring the InventoryStore section's format, e.g.: '`MarkCertRevoked` → structured: `SetRevoked`; blob: no-op (revocation is display-only there)', '`ClearCertRevoked` → structured: `ClearRevoked`', '`SetCertProjection` → structured: `SetProjection`'. No Docs repo is configured for this project, so file this directly rather than through a forge issue.
  **Constraint:** Doc convention set by the same file: the immediately preceding InventoryStore section maps every backend-interface method to its StorageService caller name in an explicit bullet list ('`AppendInventory` → structured: ... call `AppendEntry`', '`LatestSerialForSubject` (new) → structured: backend query', '`computeInventoryHMAC` → structured: fold the chain over `Entries`').

  The new `CertIndex` section (lines 141-184) declares a four-method backend interface (`Statuses`, `SetRevoked`, `ClearRevoked`, `SetProjection`) but only ever names one StorageService-level caller, `StorageService.CertStatuses`, for the read path. The three write-path StorageService methods this same PR introduces and calls in this batch -- `MarkCertRevoked` (internal/ca/revoke.go:97,121,134; internal/ca/certindex.go:848,903) which projects onto the interface's `SetRevoked`, `ClearCertRevoked` (internal/ca/certindex.go:855) which projects onto `ClearRevoked`, and `SetCertProjection` (internal/ca/certindex.go:903) which projects onto `SetProjection` -- are never named anywhere in inventory-store.md, storage-internals.md, or storage-backends.md. A contributor reading this design doc to understand how the ca package talks to a CertIndex-capable backend has no way to learn these three method names without reading the source, breaking the exact mapping pattern the InventoryStore section one page above establishes for the sibling optional-capability interface.
  </details>

- ⚙️ **Migration timeout budget is a hardcoded constant with no operator override** ([`internal/storage/sql.go:86`](https://github.com/voxpupuli/openvox-ca/blob/a74ab482c36f55fc41b7cc9b8aeca739abc166b0/internal/storage/sql.go#L86))

  ```
  	// sqlMigrationTimeout bounds a whole migration run, replacing the
  	// per-statement RequestTimeout for the duration of EnsureReady. It has to
  	// cover the slowest legitimate migration (an index build over a large
  	// inventory) plus any wait for a peer replica holding the migration lock,
  	// because a run cut short is exactly what leaves a schema half-migrated.
  	sqlMigrationTimeout = 10 * time.Minute
  ```

  💡 **Recommendation:** Add a MigrationTimeout field to SQLConfig (defaulting to the current 10 minutes) so operators with unusually large inventories or constrained I/O can raise the budget without a code change.
  **Constraint:** Operator criterion 1 (hardcoded environment-specific values that should be parameterized)

  sqlMigrationTimeout bounds both the distributed-lock wait and the migration DDL itself, and the comment explicitly acknowledges the budget has to cover an index build over a large inventory plus a wait for a peer holding the lock -- exactly the 'fleets of tens of thousands of nodes' scenario this PR's issue (#137) targets. SQLConfig (same file, lines 96-117) exposes RequestTimeout, MaxOpenConns and MaxIdleConns as operator-tunable fields, but no equivalent field exists for the migration budget, so an operator whose environment needs more than 10 minutes (slow disk I/O, a heavily contended migration lock, or a very large inventory) has no way to raise it short of a code change. This is compounded by migrationDDL (sql_migrations.go): on PostgreSQL and SQLite the whole migration runs in one transaction, so a timeout discards all progress, and WithMarkAppliedOnSuccess means the migration is retried on the next start -- but every retry hits the identical unconfigurable 10-minute wall, so a genuinely slow environment can never complete the migration without a binary patch.
  </details>

- 🧪 **Malformed-serial skip in rebuildCertIndex has no test coverage** ([`internal/ca/certindex.go:129`](https://github.com/voxpupuli/openvox-ca/blob/a74ab482c36f55fc41b7cc9b8aeca739abc166b0/internal/ca/certindex.go#L129))

  ```
  		if _, ok := n.SetString(rec.Serial, 16); !ok {
  			slog.Warn("Certificate index repair: malformed serial in index record",
  ```

  💡 **Recommendation:** Add a test that seeds one well-formed record and one record with a non-hex serial (e.g. via a direct AppendInventory blob line), runs the repair pass, and asserts the well-formed record is still reconciled correctly while the malformed one is left with its prior state and does not abort the loop for other records.
  **Constraint:** Missing edge case coverage for boundary-sensitive code (tester calibration table)

  When an index record's serial fails to parse as hex, the repair loop logs a warning and `continue`s past both the revocation-reconciliation and (already-run) projection-backfill work for that record, leaving it untouched. No test in internal/ca/certindex_test.go or internal/storage/sql_certindex_test.go constructs a record with a malformed serial to verify the loop degrades gracefully (skips only that record, still processes the rest, doesn't panic on the subsequent `revoked[serialHexStr(n)]` lookup logic).
  </details>

- 🧪 **certStatusFromRecord's serial/time validation branches are untested** ([`internal/api/handlers.go:704`](https://github.com/voxpupuli/openvox-ca/blob/a74ab482c36f55fc41b7cc9b8aeca739abc166b0/internal/api/handlers.go#L704))

  ```
  	serialInt := new(big.Int)
  	if _, ok := serialInt.SetString(rec.Serial, 16); !ok {
  		return CertStatusResponse{}, false
  	}
  	nb, err := time.Parse(storage.InventoryTimeFormat, rec.NotBefore)
  	if err != nil {
  		return CertStatusResponse{}, false
  	}
  	na, err := time.Parse(storage.InventoryTimeFormat, rec.NotAfter)
  	if err != nil {
  		return CertStatusResponse{}, false
  	}
  ```

  💡 **Recommendation:** Add a package-internal test file (package api) that calls certStatusFromRecord directly with a storage.CertRecord carrying a non-hex Serial, then one with a malformed NotBefore, then one with a malformed NotAfter, asserting ok=false for each.
  **Constraint:** Missing edge case coverage for boundary-sensitive code (tester calibration table)

  certStatusFromRecord has four independent `ok=false` branches (empty fingerprint, unparsable serial, unparsable NotBefore, unparsable NotAfter). internal/api/certindex_statuses_test.go only exercises the empty-fingerprint branch (the "falls back to the stored PEM for a projection-less record" test) -- the serial/NotBefore/NotAfter parse-failure branches have zero coverage from any test in the changeset. A regression in any of these three checks (e.g. dropping the NotAfter parse guard) would let a malformed record's zero-value fields leak into the API response undetected. The codebase already has precedent for package-internal (`package api`, not `api_test`) unit tests that could call this unexported function directly -- see internal/api/ratelimit_test.go.
  </details>

- 🧪 **Retried-revocation index re-projection has no regression test** ([`internal/ca/revoke.go:93`](https://github.com/voxpupuli/openvox-ca/blob/a74ab482c36f55fc41b7cc9b8aeca739abc166b0/internal/ca/revoke.go#L93))

  ```
  			slog.Debug("Certificate already revoked", "serial", serialStr)
  			// Still project the state into the certificate index: a retried
  			// revocation may be exactly the case where the CRL write landed
  			// but the index update did not.
  			c.markCertRevokedIndex(ctx, serialStr, entry.RevocationTime)
  ```

  💡 **Recommendation:** Add a case that calls myCA.Revoke(ctx, subject) twice for the same subject and asserts store.CertStatuses still reports the record as revoked with a non-nil RevokedAt after the second (already-revoked) call.
  **Constraint:** Missing regression test for a named recovery scenario (tester calibration table: "Missing regression test for bug fix")

  This line is new behavior added specifically to recover from a documented failure mode (CRL write succeeded, index write did not). No test in the changeset revokes the same subject/serial twice to exercise the "already revoked" branch and confirm the second call re-projects MarkCertRevoked. internal/ca/certindex_test.go's revocation test (line 134, "clears a revocation the CRL no longer corroborates when repair runs") calls Revoke exactly once per subject. This is straightforward to test with existing test infrastructure (no mocks needed -- just call Revoke twice) and covers a scenario the code comment explicitly calls out as a reason for the change.
  </details>

- 🧪 **Blob-gated Statuses filtering ("only subjects with a stored certificate") is verified only on SQLite** ([`internal/storage/sql_certindex_test.go:177`](https://github.com/voxpupuli/openvox-ca/blob/a74ab482c36f55fc41b7cc9b8aeca739abc166b0/internal/storage/sql_certindex_test.go#L177))

  ```
  	It("returns only the latest issuance per subject, and only subjects with a stored certificate", func() {
  		ctx := context.Background()
  		svc, b := newCertIndexService()

  		for _, line := range []string{
  			"0001 2024-01-01T00:00:00UTC 2029-01-01T00:00:00UTC /node1",
  			"0002 2024-01-02T00:00:00UTC 2029-01-02T00:00:00UTC /node2",
  			"0003 2024-01-03T00:00:00UTC 2029-01-03T00:00:00UTC /node1",
  		} {
  			Expect(svc.AppendInventory(ctx, line)).To(Succeed(), "AppendInventory")
  		}
  		// node1 has a stored cert; node2 was cleaned (rows remain, blob gone).
  		putCertBlob(b, "node1")
  ```

  💡 **Recommendation:** Add the row-without-blob scenario to certIndexRoundTrip (or a new shared helper) so it runs against MySQL and Postgres too, e.g. append a second subject and only Put a cert blob for one of them, then assert the other is excluded from Statuses.
  **Constraint:** Missing edge case coverage for boundary-sensitive code / dialect coverage gap in a multi-dialect test suite

  This is the only test exercising SQLBackend.Statuses' "stored[r.Subject]" gating -- the branch that excludes inventory rows whose certificate blob was removed by clean/cleanup -- and it lives only under Describe("SQLiteCertIndex", ...) (line 140). The shared certIndexRoundTrip helper that MySQL and Postgres reuse (sql_mysql_integration_test.go:264-272, sql_postgres_integration_test.go:247-255) always Puts a cert blob for every subject it appends (certIndexRoundTrip, lines 65-67), so it never creates a row-without-blob scenario. The MAX(id)/GROUP BY subquery combined with the List(certPrefix)-derived exclusion set is real per-row, per-dialect SQL logic that ships untested against real PostgreSQL/MySQL.
  </details>

- 🧪 **No regression test that CertIndex survives OverlayBackend wrapping, unlike its InventoryStore sibling** ([`internal/storage/backend.go:260`](https://github.com/voxpupuli/openvox-ca/blob/a74ab482c36f55fc41b7cc9b8aeca739abc166b0/internal/storage/backend.go#L260))

  ```
  func asCertIndex(b Backend) (CertIndex, bool) {
  	for {
  		if idx, ok := b.(CertIndex); ok {
  			return idx, true
  		}
  		unwrapper, ok := b.(interface{ Unwrap() Backend })
  		if !ok {
  			return nil, false
  		}
  		b = unwrapper.Unwrap()
  	}
  }
  ```

  💡 **Recommendation:** Add a test that wraps a SQL backend with NewOverlayBackend (e.g. overriding KeyCAKey, as InventoryMigrationRoundTripOverlayDestination does) and asserts CertStatuses/MarkCertRevoked/etc. still see the capability and behave correctly through the wrapper.
  **Constraint:** Missing regression test for a documented failure class already seen in production for the sibling capability

  internal/storage/sql_inventory_test.go:190-259 ("InventoryMigrationRoundTripOverlayDestination") documents, at length, a real production incident: OverlayBackend did not implement InventoryStore itself, so the naive type assertion failed to see through the wrapper and silently downgraded the integrity scheme, causing ErrInventoryTampered on startup. asCertIndex (backend.go:260-273) uses the exact same probe-then-Unwrap() pattern that was the fix for that incident, but no test exercises CertIndex through an OverlayBackend-wrapped SQL backend the way sql_inventory_test.go does for InventoryStore. The only capability-probing test for CertIndex, "degrades gracefully on backends without the capability" (sql_certindex_test.go:281-293), uses a bare filesystem backend (New(GinkgoT().TempDir())), which never touches the Unwrap loop at all -- it exits on the first iteration with ok=false. The unwrap branch of asCertIndex is therefore never exercised by any test in this changeset.
  </details>

- 🧪 **Migration atomicity/idempotency guarantee never exercised under an actual mid-migration failure** ([`internal/storage/sql_migrations.go:52`](https://github.com/voxpupuli/openvox-ca/blob/a74ab482c36f55fc41b7cc9b8aeca739abc166b0/internal/storage/sql_migrations.go#L52))

  ```
  // migrationDDL runs one migration's statements as a unit. On PostgreSQL and
  // SQLite, whose DDL is transactional, that unit is a transaction: the migration
  // either applies whole or not at all, and a cancelled context can no longer
  // leave the schema half-changed. MySQL/MariaDB commit implicitly on every DDL
  // statement, so a transaction there would be a comforting lie — its migrations
  // stay recoverable by being idempotent instead (addColumnIfMissing and friends),
  ```

  💡 **Recommendation:** Add a test that registers a throwaway migration (or wraps an existing one) whose migrationDDL callback deliberately fails partway through a multi-statement sequence, then asserts: on SQLite/Postgres the table is unchanged (no columns added) and on MySQL the columns added before the failure point persist and a subsequent EnsureReady completes the rest -- mirroring the "recovering a half-applied migration" test but driven by a real failure instead of a manufactured one.
  **Constraint:** Missing edge case coverage for boundary-sensitive code (severity.md HIGH/MEDIUM boundary; treated as MEDIUM given the difficulty of fault injection)

  This comment documents a specific reliability contract: on Postgres/SQLite a failing migration must leave the schema completely untouched (transactional rollback), while on MySQL a failing migration must leave it in a state addColumnIfMissing/createIndexIfMissing can finish on retry. No test in sql_migrations_test.go or sql_migrations_integration_test.go actually forces migrationDDL's inner fn to fail mid-run and then asserts the resulting schema state. The closest test, "recovering a half-applied migration" (sql_migrations_test.go:77-106), only verifies recovery from a half-state that the test manufactures directly via dropCertIndexSchema/DELETE FROM bun_migrations -- it never proves migrationDDL itself produces (MySQL) or refuses to produce (Postgres/SQLite) that half-state when a statement inside it actually errors. A future migration that forgets to route a statement through migrationDDL, or a bug in the MySQL/dialect branch, would not be caught by any test here.
  </details>

</details>

<details><summary>🔵 LOW (4)</summary>

- 📚 **"filled at signing" undersells the projection-write paths shown in this batch** ([`docs/development/inventory-store.md:145`](https://github.com/voxpupuli/openvox-ca/blob/a74ab482c36f55fc41b7cc9b8aeca739abc166b0/docs/development/inventory-store.md#L145))

  ```
  (`fingerprint_sha256`, `dns_alt_names`, `auth_extensions`, filled at signing)
  ```

  💡 **Recommendation:** Change 'filled at signing' to 'filled at signing or import' (or similarly reference both `issueLeafLocked` and `ImportCertificate`) so the design rule bullet that follows ('Blob-imported rows carry no projection') reads as the exception it is, not a restatement of the general case. No Docs repo is configured for this project, so file this directly rather than through a forge issue.
  **Constraint:** Documentation accuracy -- the described behavior should match the code paths that write the described field.

  internal/ca/importcert.go:202 (`proj := certProjectionFor(cert)`) computes and stores the same display projection at `ImportCertificate` time, not only at signing -- it is passed non-nil into `AppendInventoryRecord`, so an imported certificate never goes through the null-projection/backfill path the doc later describes for 'blob-imported rows'. The parenthetical 'filled at signing' therefore describes only one of the two paths (internal/ca/signing.go:436-449 and internal/ca/importcert.go:202) that populate these columns at write time, which could lead a reader to assume `ImportCertificate` rows start projection-less like migrated blob rows do.
  </details>

- 📚 **CertIndex capability has no tracked blog-post opportunity** ([`internal/storage/backend.go:217`](https://github.com/voxpupuli/openvox-ca/blob/a74ab482c36f55fc41b7cc9b8aeca739abc166b0/internal/storage/backend.go#L217))

  ```
  // CertIndex is an optional capability of InventoryStore backends: the backend
  // that owns the structured inventory rows can answer certificate-status
  // queries from indexed columns instead of forcing the caller into a
  // list-certs → read-PEM → parse → CRL-check scan per subject.
  ```

  💡 **Recommendation:** Once a Docs repo is configured for this project, file a blog-opportunity issue there. Until the Docs repo is known, file this manually in whatever repository tracks content/marketing issues for openvox-ca.
  **Constraint:** Curator criterion 4 (Blog Opportunity Identification): significance threshold "New major capability" met -- this PR (issue #137) grows puppet_ca_inventory into a certificate index, adding a new public interface (CertIndex, 4 methods), a new schema migration (20260725000000_certindex.go), and new StorageService entry points (CertStatuses, MarkCertRevoked, ClearCertRevoked, SetCertProjection).

  This batch implements a genuinely new backend capability -- indexed certificate-status queries replacing a PEM-scan -- significant enough to warrant a release/feature announcement. No "Docs repo" is configured in this council run (no Review Council Configuration section and no .review-council config found in the repository), so per the Bash Access Restriction I did not invoke `gh` to search for an existing `blog`-labeled issue and cannot confirm whether one already exists.
  </details>

- 🧪 **AppendInventoryRecord's "ignored on non-CertIndex backends" contract has no direct test** ([`internal/storage/storage.go:197`](https://github.com/voxpupuli/openvox-ca/blob/a74ab482c36f55fc41b7cc9b8aeca739abc166b0/internal/storage/storage.go#L197))

  ```
  // AppendInventoryRecord is AppendInventory with an optional certificate-index
  // projection. Issuance paths (signing, import) pass the display fields
  // denormalised from the certificate they just stored; backends with the
  // CertIndex capability persist them on the new row (state "signed"), all
  // others ignore them. A nil proj records the entry without a projection,
  ```

  💡 **Recommendation:** Add an assertion (e.g. in the same "degrades gracefully" test) that calls svc.AppendInventoryRecord(ctx, line, &CertProjection{...}) against the filesystem backend and asserts it succeeds and behaves identically to AppendInventory without a projection.
  **Constraint:** Missing property/fuzz/contract test (general case)

  The doc comment states a specific contract for backends without CertIndex: a non-nil proj passed to AppendInventoryRecord is silently ignored rather than erroring or being persisted somewhere unexpected. No test in this changeset calls AppendInventoryRecord with a non-nil CertProjection against a non-SQL backend (e.g. New(GinkgoT().TempDir())) to confirm this. The existing "degrades gracefully on backends without the capability" test (sql_certindex_test.go:281-293) only exercises CertStatuses/MarkCertRevoked/ClearCertRevoked/SetCertProjection, not AppendInventoryRecord itself.
  </details>

- ⚙️ **No log signal that a (potentially minutes-long) migration run has started or finished** ([`internal/storage/sql.go:346`](https://github.com/voxpupuli/openvox-ca/blob/a74ab482c36f55fc41b7cc9b8aeca739abc166b0/internal/storage/sql.go#L346))

  ```
  	migrator := migrate.NewMigrator(b.db, sqlMigrations, migrate.WithMarkAppliedOnSuccess(true))
  	if err := migrator.Init(ctx); err != nil {
  		return fmt.Errorf("initialising migrations: %w", err)
  	}
  	if _, err := migrator.Migrate(ctx); err != nil {
  		return fmt.Errorf("running migrations: %w", err)
  	}
  	return nil
  }
  ```

  💡 **Recommendation:** Emit an slog.Info before acquiring the migration lock and another after migrator.Migrate succeeds, so a long-running-but-healthy migration is visible in logs instead of looking indistinguishable from a stall.
  **Constraint:** Operator criterion 5 (long-running operations provide progress feedback)

  EnsureReady's own documentation budgets up to sqlMigrationTimeout (10 minutes) for a migration that includes an index build over a large inventory, and the migration lock wait can add further delay from a peer replica. Nothing in EnsureReady logs that a migration run is starting, is still waiting on the distributed lock, or has completed -- the only slog call in this function is a single warning path (the failed lock-release warning); the pool-floor clamp warning lives in a different function, NewSQLBackend, not in EnsureReady. An operator watching startup logs during a slow-but-healthy multi-minute migration has no way to distinguish it from a hang until it either finishes or the process is killed by an external timeout.
  </details>

</details>

---

bootc:
Thanks — ten of the fourteen are fixed, four are answered rather than changed, and a follow-up council round found four more things worth having. Pushed as `9dedc71` and `c252e20`.

**The HIGH and the untested guards.** `backfillCertProjection`'s refusal to fill a projection from a certificate whose serial does not match the row is pinned now, including that startup still succeeds. Reaching the malformed-serial branch turned out to need a stored certificate blob as well as the row — `CertStatuses` reports only subjects that have one, so a row without a blob never reaches the repair loop, which is why that branch looked unreachable from the inventory alone. Also pinned: `Revoke`'s re-projection on its already-revoked early return (the retry path — if that exit skipped the write, a row that missed the first projection would stay wrong for ever), `AppendInventoryRecord`'s ignored-projection contract on blob backends, `certStatusFromRecord`'s four rejections, and `CertIndex` surviving an `OverlayBackend` — the regression its `InventoryStore` sibling already had, because that bug shipped once.

**The MEDIUMs on the operator surface.** `sql_migration_timeout_sec` is configurable (default ten minutes) and a migration run now logs before and after, naming the group and count.

**What the follow-up round found, which the original pass did not:**

**The knob I had just added could not be set the way the CA is deployed.** Three of four reviewers found it independently: the config key went into the YAML examples, and the environment variable was never added — while `docs/storage-backends.md` says in the sentence above its table that the SQL backends "share one set of config keys and environment variables", and lists every sibling. An operator deploying by environment, which is how the container images are configured, silently got ten minutes: exactly the wall the knob exists to raise. The spec named "applies all SQL-related env vars" passed throughout. It exists now, with the table row, pinned at each of its four hops.

**`certificate_statuses` answered revocation from the index row, not the CRL.** Every other status path in the file derives state from the signed CRL. The row is a projection maintained best-effort — a swallowed `SetRevoked` logs and carries on — and the repair pass that reconciles it runs only at startup, so on a CA that restarts quarterly the list view disagreed with the single-subject view for a quarter.

**The indexed branch dropped subjects the scan branch lists.** `Statuses` reports the *intersection* of stored certificates and inventory rows, so the crash window between those two writes left a certificate the index cannot see, while the scan path lists it from the blobs. Same endpoint, opposite answers by backend — and worse than a missing row: the certificate vanished, never landed in `seen`, and its pending CSR was then reported as `requested` for a host that already holds a certificate this CA is serving.

**An interrupted index repair claimed success.** The loop never checked its context, so a cancelled pass — SIGTERM, a startup probe, or the 60-second bootstrap-lock budget it inherits when it loses the bootstrap race — emitted one warning per remaining record and then logged "Certificate index repaired" for a quarter-done pass.

One thing I wrote and then withdrew, since the reasoning matters: I first made a serial mismatch in the statuses fallback *omit* the subject, by analogy with the repair refusing that pairing. The analogy does not hold — the repair would *write* the PEM's fields onto the row, whereas the handler uses nothing of the row, so the response describes the stored certificate accurately and omitting would drop a real certificate. An existing spec pinned exactly that case, which is how the over-strict version was caught.

**Four findings answered rather than changed:**

- **The migration-locking fix sharing this PR.** It is already its own self-contained commit with its own tests, plus a follow-up acting on its review. Splitting it into a separate PR changes the stack's merge order, so that is your call — and worth noting that its lock is what makes this PR's DDL safe on replicas, so splitting would ship the schema change ahead of its guard.
- **Blob-gated `Statuses` filtering verified only on SQLite**, and **migration atomicity under a real mid-migration failure.** Both want cross-backend or fault-injection harnesses of their own rather than a bolt-on. One reviewer noted `certIndexMigrationRollback` already gives the fixture shape for the second.
- **A blog-post opportunity.** No Docs repo is configured, so there is nowhere to file it.

And one deferred with a reason: `WithLock` probes `Locker` with a bare type assertion where both capability siblings unwrap, and no spec builds a real `OverlayBackend` over a lock-capable base — it is correct today only because `OverlayBackend` forwards `AcquireLock` itself. Pre-existing, and the same defect class this branch closed for `CertIndex`, but routing it through an `asLocker` helper touches every backend's locking path.

Issue #139, opened 2026-07-24
Decompose Redis/Valkey inventory into structured entries (InventoryStore/CertIndex)


bootc:
## Summary

Make the **Redis/Valkey** backend store the certificate inventory as decomposed structures (implementing `InventoryStore`, and later `CertIndex`) instead of the single append-only `inventory:data` blob it uses today.

Part of the pre-1.0.0 storage-interface review. Sibling issues: #137 (SQL cert index), #138 (etcd decomposition).

## Motivation

Same as the etcd case: the current Redis inventory append is **read-the-whole-blob, concatenate, write-the-whole-blob** — the [`appendLuaScript`](../blob/main/internal/storage/redis.go) `GET`s the entire inventory and `SET`s it back. That's **O(size-of-entire-inventory) per certificate issued** (O(n²) overall). And like etcd, blob storage gives **no cross-replica duplicate-serial guarantee** — only SQL does.

## Feasibility — workable via Lua, but fiddlier than etcd

Redis has the data structures (a `LIST`/`ZSET` for ordered entries, a `HASH` for serial→entry and subject→serial) and single-command atomicity, with multi-key atomicity via Lua — the pattern the backend already uses for append and locks:

- **Serial uniqueness**: `HSETNX inventory:by-serial <serial> ...` returns 0 if present — atomic uniqueness in one op.
- **Append + chain + indices**: one Lua script does *check serial unique → RPUSH entry → set subject index → set head*, atomically server-side.

**Same HMAC wrinkle as etcd:** `chainInventoryMAC` is a keyed hash computed in Go, so the chain can't be folded into the server-side script. Resolve it the same way: compute `newHead` in Go, pass `prevExpected` + `newHead` into the script, and have the script abort if the stored head ≠ `prevExpected` (retry in Go). Optimistic concurrency, mirroring the etcd approach.

**Durability caveat (unchanged):** Redis atomicity is single-primary; an async-replicated Sentinel failover can lose a just-appended entry — exactly as it can today, and as the lock code already documents at the top of `AcquireLock`. Decomposition doesn't make this worse; it just moves from one big value to many small ones.

## Why it doesn't ripple upward

Same as etcd: the decomposed backend still serves `KeyInventory` as rendered `inventory.txt` text via the `InventoryStore` blob shim, so `Migrate` and the OCSP index build are untouched. Purely internal to the backend.

## Cost — the fiddly one

[...]

PR #154, opened 2026-07-26
Decompose the etcd inventory into per-entry keys (InventoryStore/CertIndex)

### LINKED ISSUE #138: Decompose etcd inventory into per-entry keys (InventoryStore/CertIndex)

bootc:
Implements #138: the etcd backend stores the certificate inventory as one key per issued certificate instead of the single append-only `inventory/data` blob, implementing `InventoryStore` and — since #137 settled the `CertRecord` field set — `CertIndex` as well.

**Stacked on #151** (`feature/cert-index`): only the commits from "Decompose the etcd inventory into per-entry keys" onwards belong to this PR; the diff will collapse once #151 merges.

## Why

The blob append was read-everything, concatenate, write-everything: O(size of the inventory) per signing, O(n²) to issue n certificates, plus a whole-blob HMAC recompute per append. At fleet scale every signing rewrote megabytes. Decomposition also closes the duplicate-serial gap: blob backends had no cross-replica uniqueness guarantee — that was previously exclusive to SQL.

## Design

Zero-padded `inventory/entries/<seq>` keys hold JSON records in issuance order; `inventory/by-serial/<serial>` existence is the atomic duplicate-serial guard; `inventory/by-subject/<subject>` gives O(1) latest-serial lookups; `inventory/seq` is both sequence allocator and the fence every mutation guards on and re-puts. etcd transactions are compare-then-op, so the chained integrity head is computed in Go between a read and a fence-guarded commit — the same optimistic ModRevision-retry shape the blob append already used, now with O(1) appends.

- **Bulk rewrites are batched** under etcd's default `--max-txn-ops`. Every prune batch commits a head covering exactly the entries that remain, so no observer ever sees entries and head out of sync, and a crash mid-prune leaves a valid partially-pruned inventory. One cleanup pass removes at most 900 entries so a large backlog drains over scheduled runs instead of blowing the CRL lock budget; `PruneEntries` returns every entry actually removed — accumulated across batches and retries, even alongside an error — and `CleanupExpiredCerts` finishes the CRL/blob cleanup for them on a context that survives the prune's deadline.
- **In-place upgrade.** `EnsureReady` converts a pre-decomposition blob under a distributed lock: the blob is verified against its stored whole-blob HMAC first (unverifiable or mismatching states fail startup, as the old code would have), stays authoritative until the final commit so an interrupted conversion resumes, and mixed-version writes are detected and refused. A CA upgraded while its inventory is empty has its verified whole-blob head removed so the first verification re-baselines cleanly.
- **Certificate index.** `GET /certificate_statuses` is answered from the decomposed entries with the same rebuildable-projection semantics as SQL. Serials duplicated in a legacy blob (possible pre-conversion) cannot be addressed by the one-to-one by-serial index: index writes for them are refused rather than aliased onto the wrong bearer, `Statuses` reports them with a new `CertStateUnknown`, and the handler derives their real state from the signed CRL.
- The `KeyInventory` logical key stays served through a render/parse shim, so `Migrate` and the OCSP index build are untouched; filesystem ⇄ etcd migrations work in both directions with the head rebuilt by `MigrateService`.

The interface change to note: `InventoryStore.PruneEntries` now takes a step function (`advanceHead`) instead of a whole-fold callback, and its contract documents partial completion, the authoritative-even-on-error return, and per-call bounding. The SQL implementation is updated to match; its behaviour is unchanged.

## Review

The branch went through three rounds of automated multi-persona review, finishing with a unanimous APPROVE and zero retracted findings across 49 raised; every CRITICAL/HIGH/MEDIUM was fixed in code (interruption safety, partial-prune accounting, fail-closed legacy verification, ambiguous-serial state resolution) rather than documented away. The etcd integration suite grew from 6 to ~30 specs, including chain tamper negative controls, multi-batch prune consistency observed between commits, conversion resume/refusal states, and both ends of the partial-prune contract.

No new dependencies and no new cryptography: `go.etcd.io/etcd/api/v3` moves from indirect to direct at the version already in the module graph, and the hash chain uses the existing `crypto/hmac`/SHA-256 helpers.

Closes #138.

### DIFF
--- a/cmd/openvox-ca/cert_cleanup.go
+++ b/cmd/openvox-ca/cert_cleanup.go
@@ -65,7 +65,10 @@ func cleanupExpiredOnce(ctx context.Context, c *ca.CA, retain time.Duration) {
 	removed, err := c.CleanupExpiredCerts(ctx, retain)
 	switch {
 	case err != nil:
-		slog.Warn("Expired-certificate cleanup failed", "error", err)
+		// removed can be non-zero here: a batched backend (etcd) may have
+		// durably pruned entries before failing, and the operator needs to
+		// know state changed permanently despite the error.
+		slog.Warn("Expired-certificate cleanup failed", "removed", removed, "error", err)
 	case removed > 0:
 		slog.Info("Expired certificates cleaned up", "removed", removed)
 	default:
--- a/docs/development/inventory-store.md
+++ b/docs/development/inventory-store.md
@@ -215,39 +215,56 @@ Rules that keep the decomposed structure coherent:
 - **Appends are O(1)** — six puts guarded on the fence plus
   `CreateRevision(by-serial/<serial>) == 0`, which makes duplicate-serial
   rejection atomic cluster-wide (previously a SQL-only guarantee).
-- **Bulk rewrites are batched but every commit is consistent.** Prunes and
-  imports larger than one transaction (bounded well under etcd's default
-  `--max-txn-ops` of 128) are split into batches, and each batch writes a head
-  covering exactly the entries that remain after it. A concurrent verifier
-  never sees entries and head out of sync, and a crash mid-prune leaves a
-  valid, partially-pruned inventory rather than a spurious tamper alarm.
-  Because a batched prune can partially complete, `PruneEntries` returns every
-  entry actually removed — accumulated across batches and retries, even
-  alongside an error — so `CleanupExpiredCerts` can always finish the CRL and
-  blob cleanup for what was deleted (see the contract in `backend.go`). Prune
-  batches run newest-first, which keeps the intermediate heads cheap: each is
-  a cached prefix fold over the untouched older entries resumed across the
-  survivor tail, rather than a full refold per batch.
+- **Bulk rewrites are batched; prune commits are individually consistent.**
+  Prunes and imports larger than one transaction (bounded well under etcd's
+  default `--max-txn-ops` of 128) are split into batches. Each *prune* batch
+  writes a head covering exactly the entries that remain after it, so a
+  concurrent verifier never sees entries and head out of sync and a crash
+  mid-prune leaves a valid, partially-pruned inventory rather than a spurious
+  tamper alarm. *Import* batches carry no head at all — the head is left for
+  `RebuildInventoryHMAC` (migration) or dropped in the final commit (legacy
+  conversion) — which is exactly why the legacy blob stays authoritative
+  until the import's final commit and why the marker-guard/resume machinery
+  exists. Because a batched prune can partially complete, `PruneEntries`
+  returns every entry actually removed — accumulated across batches and
+  retries, even alongside an error — so `CleanupExpiredCerts` can always
+  finish the CRL and blob cleanup for what was deleted (see the contract in
+  `backend.go`). Prune batches run newest-first, which keeps the intermediate
+  heads cheap (each is a cached prefix fold over the untouched older entries
+  resumed across the survivor tail), and one call removes at most a bounded
+  number of batches so a huge backlog cannot blow the caller's lock budget —
+  deferred matches stay present and consistent for later runs.
 - **Legacy blobs are decomposed in place.** `EnsureReady` detects a non-empty
   pre-decomposition `inventory/data` blob, takes a distributed lock
   (`inventory-decompose`), verifies the blob against its stored whole-blob
-  HMAC (the key is a backend blob, so it is available; a mismatch fails
-  startup with `ErrInventoryTampered` exactly as the old code would have),
-  imports the lines into entry keys, and empties the marker only in the final
-  commit. The verified HMAC is deleted in the same import — it is not a chain
-  head, so it cannot carry over — and the next verification re-baselines from
-  the imported entries; only the import window itself is uncovered. Because
-  the blob stays authoritative until that final commit, an interrupted import
-  is detected on the next start (the partial entries are the import-written
-  prefix of the blob) and redone from the intact blob; entries that are *not*
-  such a prefix mean a mixed-version cluster wrote both forms, which is
-  refused with an explicit error rather than guessed at. Duplicate serials in
-  the legacy blob — possible, since blob backends never had a cluster-wide
-  uniqueness guarantee — are imported verbatim with a warning, with the
-  by-serial index pointing at each serial's newest bearer. All replicas must
-  still upgrade together: an old-version writer appending to the blob
-  mid-import is detected via the marker guard and the import restarts, but
-  the race only closes once the old writers are gone.
+  HMAC (the key is a backend blob, so it is available), imports the lines
+  into entry keys, and empties the marker only in the final commit.
+  Verification is fail-closed: a mismatch — or a stored HMAC that cannot be
+  verified because the key is missing or malformed — fails startup with
+  `ErrInventoryTampered`, exactly as the pre-decomposition code would have;
+  the operator acknowledges a lost baseline by deleting the stored
+  `inventory/hmac` key. The verified HMAC is deleted in the same import — it
+  is not a chain head, so it cannot carry over — and the next verification
+  re-baselines from the imported entries; only the import window itself is
+  uncovered. A CA upgraded while its inventory is *empty* has no import to
+  drop the head as part of, so `EnsureReady` handles that case separately:
+  when zero entries exist and the stored head verifies as the whole-blob MAC
+  of an empty inventory, it is deleted so the first verification re-baselines
+  cleanly (any other head over zero entries is left for verification to
+  flag). Because the blob stays authoritative until the import's final
+  commit, an interrupted import is detected on the next start (the partial
+  entries are the import-written prefix of the blob) and redone from the
+  intact blob; entries that are *not* such a prefix mean a mixed-version
+  cluster wrote both forms, which is refused with an explicit error rather
+  than guessed at. Duplicate serials in the legacy blob — possible, since
+  blob backends never had a cluster-wide uniqueness guarantee — are imported
+  verbatim with a warning; their by-serial keys carry an ambiguity sentinel
+  that keeps the serial reserved against reissue but makes certificate-index
+  writes for it explicit no-ops, since a one-to-one index cannot say which
+  bearer such a write is meant for. All replicas must still upgrade together:
+  an old-version writer appending to the blob mid-import is detected via the
+  marker guard and the import restarts, but the race only closes once the old
+  writers are gone.
 - **Certificate-index writes stay off the chain.** `SetRevoked` /
   `ClearRevoked` / `SetProjection` rewrite a single entry key guarded on its
   own ModRevision (the mutable fields are not chain input), so index repair
--- a/docs/development/inventory-store.md
+++ b/docs/development/inventory-store.md
@@ -261,7 +261,10 @@ Rules that keep the decomposed structure coherent:
   verbatim with a warning; their by-serial keys carry an ambiguity sentinel
   that keeps the serial reserved against reissue but makes certificate-index
   writes for it explicit no-ops, since a one-to-one index cannot say which
-  bearer such a write is meant for. All replicas must still upgrade together:
+  bearer such a write is meant for. `Statuses` reports those records with
+  `CertStateUnknown`, the statuses handler derives their real state from the
+  signed CRL, and the startup repair pass skips them (they can never
+  converge). All replicas must still upgrade together:
   an old-version writer appending to the blob mid-import is detected via the
   marker guard and the import restarts, but the race only closes once the old
   writers are gone.
--- a/docs/development/inventory-store.md
+++ b/docs/development/inventory-store.md
@@ -221,15 +221,33 @@ Rules that keep the decomposed structure coherent:
   covering exactly the entries that remain after it. A concurrent verifier
   never sees entries and head out of sync, and a crash mid-prune leaves a
   valid, partially-pruned inventory rather than a spurious tamper alarm.
+  Because a batched prune can partially complete, `PruneEntries` returns every
+  entry actually removed — accumulated across batches and retries, even
+  alongside an error — so `CleanupExpiredCerts` can always finish the CRL and
+  blob cleanup for what was deleted (see the contract in `backend.go`). Prune
+  batches run newest-first, which keeps the intermediate heads cheap: each is
+  a cached prefix fold over the untouched older entries resumed across the
+  survivor tail, rather than a full refold per batch.
 - **Legacy blobs are decomposed in place.** `EnsureReady` detects a non-empty
-  pre-decomposition `inventory/data` blob, takes a distributed lock, imports
-  the lines into entry keys, and empties the marker. The stored whole-blob
-  HMAC is deleted in the same import: it is not a chain head and the backend
-  does not hold the key to translate it, so the next verification re-baselines
-  from the imported entries. Tamper detection therefore does not cover the
-  decomposition window itself — and all replicas must upgrade together, since
-  an old-version writer appending to the blob mid-import is detected and
-  retried, but the race only closes once the old writers are gone.
+  pre-decomposition `inventory/data` blob, takes a distributed lock
+  (`inventory-decompose`), verifies the blob against its stored whole-blob
+  HMAC (the key is a backend blob, so it is available; a mismatch fails
+  startup with `ErrInventoryTampered` exactly as the old code would have),
+  imports the lines into entry keys, and empties the marker only in the final
+  commit. The verified HMAC is deleted in the same import — it is not a chain
+  head, so it cannot carry over — and the next verification re-baselines from
+  the imported entries; only the import window itself is uncovered. Because
+  the blob stays authoritative until that final commit, an interrupted import
+  is detected on the next start (the partial entries are the import-written
+  prefix of the blob) and redone from the intact blob; entries that are *not*
+  such a prefix mean a mixed-version cluster wrote both forms, which is
+  refused with an explicit error rather than guessed at. Duplicate serials in
+  the legacy blob — possible, since blob backends never had a cluster-wide
+  uniqueness guarantee — are imported verbatim with a warning, with the
+  by-serial index pointing at each serial's newest bearer. All replicas must
+  still upgrade together: an old-version writer appending to the blob
+  mid-import is detected via the marker guard and the import restarts, but
+  the race only closes once the old writers are gone.
 - **Certificate-index writes stay off the chain.** `SetRevoked` /
   `ClearRevoked` / `SetProjection` rewrite a single entry key guarded on its
   own ModRevision (the mutable fields are not chain input), so index repair
@@ -274,3 +292,9 @@ Each phase is a separate commit.
    backends: latest-wins lookups, chain tamper detection (modify / insert /
    delete), byte-identical render, and a filesystem ⇄ sqlite migration
    round-trip that verifies integrity on both sides.
+5. **Certificate index** (issue #137, a later extension). Extend the SQL
+   inventory table with the projection/state columns, define `CertIndex`,
+   serve `certificate_statuses` from it, and add the startup repair pass.
+6. **etcd decomposition** (issue #138, a later extension). Implement
+   `InventoryStore` and `CertIndex` on the etcd backend with per-entry keys,
+   including the in-place legacy blob conversion described above.
--- a/docs/development/inventory-store.md
+++ b/docs/development/inventory-store.md
@@ -35,8 +35,9 @@ an internal and on-disk-compatibility concern only.
 ## Goal
 
 Let backends that can do better store the inventory as **structured records**
-(e.g. a SQL table) while preserving exact behaviour for backends that keep the
-blob (filesystem, etcd, redis/valkey). This is opt-in per backend.
+(e.g. a SQL table, or one etcd key per entry) while preserving exact behaviour
+for backends that keep the blob (filesystem, redis/valkey). This is opt-in per
+backend.
 
 ## Design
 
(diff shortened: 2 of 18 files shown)


trevor-vaughan-ai:
> **Obsolete.** Superseded by the [current Review Council verdict](https://github.com/voxpupuli/openvox-ca/pull/154#issuecomment-5308656778) for commit `f6a5760`. 

>
> Models used:
> - Capable tier
> - Standard tier
> - claude-sonnet-5

_Reviewed at commit [`7ff62e7`](https://github.com/voxpupuli/openvox-ca/commit/7ff62e7ce038ff9f10a10377730e31bfdab6e7b7) (standard effort)._

**TL;DR:** A fleet-scale etcd inventory PR needs changes: 3 HIGH findings on untested concurrency-critical paths, plus 1 MEDIUM test and 1 MEDIUM docs finding.

**Findings:** 🔴 0 Critical, 🟠 3 High, 🟡 2 Medium, 🔵 0 Low

|---|---|---|
| 🧪 Tester (code) | ❌ Changes | 3 HIGH, 1 MEDIUM |

- 🧪 **mutateRecordBySerial's CAS retry loop has no concurrent-writer test** ([`internal/storage/etcd_certindex.go:146`](https://github.com/voxpupuli/openvox-ca/blob/7ff62e7ce038ff9f10a10377730e31bfdab6e7b7/internal/storage/etcd_certindex.go#L146))

  ```
  func (b *EtcdBackend) mutateRecordBySerial(ctx context.Context, serial string, mutate func(rec *CertRecord) bool) error {
  	serialPhys := b.invPhys(etcdInvSerialSub + serial)
  	for attempt := range etcdMaxTxnRetries {
  ```

  💡 **Recommendation:** Add an integration test (build tag etcd_integration) firing several goroutines calling SetRevoked/SetProjection against the same serial concurrently (mirroring EtcdInventoryConcurrentAppends), asserting the final state is consistent and no update is lost. Add a hook-based fence-conflict test and a retry-exhaustion test analogous to the PruneInventory ones to pin the 'too many concurrent writers' path.
  **Constraint:** testing_conventions (lang-go.md): concurrency-sensitive code needs tests that actually drive concurrent writers into conflict, not just single-writer paths; the review focus explicitly calls for checking conflict-retry coverage.

  SetRevoked, ClearRevoked, and SetProjection all funnel through mutateRecordBySerial (lines 146-211), whose read-mutate-CAS-write loop is the only concurrency guard on index writes - the function's own doc comment cites 'SetRevoked racing SetProjection during index repair' as the scenario it protects against. Every test that exercises this path - the whole 'EtcdCertIndex' Describe block in internal/storage/etcd_integration_test.go (lines 1425-1544: 'serves the projected latest issuance...', 'projects revocation idempotently...', 'backfills the projection...', 'keeps the integrity chain untouched...', 'treats unknown serials as no-ops...', 'treats a dangling by-serial pointer as a no-op...') - calls SetRevoked/ClearRevoked/SetProjection sequentially from a single goroutine. None spins up concurrent goroutines to force a ModRevision conflict and prove the retry-then-succeed path works, and none exhausts etcdMaxTxnRetries to prove the 'too many concurrent writers' error (line 210) is reachable. By contrast, the sibling AppendLine/PruneEntries paths got exactly this treatment (EtcdInventoryConcurrentAppends with 4 concurrent writers; the pruneBatchHook-based fence-conflict and retry-exhaustion tests at lines 760-898). A regression in mutateRecordBySerial's guard would not be caught by this suite even though Revoke() (which calls SetRevoked) is documented elsewhere in this PR as a replica-safe, concurrently-invoked operation.
  </details>

- 🧪 **rebuildCertIndex's CertStateUnknown skip branch is never exercised** ([`internal/ca/certindex.go:117`](https://github.com/voxpupuli/openvox-ca/blob/7ff62e7ce038ff9f10a10377730e31bfdab6e7b7/internal/ca/certindex.go#L117))

  ```
  	for _, rec := range records {
  		if rec.State == storage.CertStateUnknown {
  			// The backend cannot address this record's serial one-to-one
  			// (duplicated legacy serial on etcd): index writes for it are
  			// refused, so backfill and reconciliation can never converge.
  			// Readers derive its state from the CRL instead; skip it rather
  			// than re-attempt and warn on every start.
  			continue
  		}
  ```

  💡 **Recommendation:** Add a ca-package test (in certindex_test.go) that seeds a CertRecord with State == storage.CertStateUnknown, either via a small fake CertIndex-implementing backend (the unknownStateBackend pattern already used in certindex_statuses_test.go) or the real etcd path under etcd_integration, then calls ca.New(...).Init(ctx) and asserts the record is left alone (no backfill/reconciliation attempt) and excluded from the 'missing' backfill-progress count.
  **Constraint:** Coverage completeness: new conditional branch in core startup-repair logic with no direct or indirect test.

  This PR adds two new branches to rebuildCertIndex gated on storage.CertStateUnknown: the missing-projection counter exclusion at line 105 and the skip-and-continue quoted above (lines 117-124). CertStateUnknown appears in only two test files in the whole changeset. internal/api/certindex_statuses_test.go's 'derives an unknown-state record's state from the CRL' test injects the CertStateUnknown record via a Statuses() override on a fake backend *after* myCA.Init(ctx) has already run and rebuildCertIndex has already completed over an empty index - the unknown record is created only at request time and never flows through rebuildCertIndex; that test verifies internal/api/handlers.go's derive-from-CRL branch instead. internal/storage/etcd_integration_test.go's 'tolerates duplicate serials in the legacy blob' test verifies EtcdBackend.Statuses() output directly, not anything in the ca package. internal/ca/certindex_test.go (read in full, 258 lines) - the file that actually drives rebuildCertIndex via ca.New(...).Init(ctx) - never references CertStateUnknown and only backs onto storage.NewSQLBackend, which never produces CertStateUnknown (only the etcd backend does). So this skip logic, whose entire stated purpose is to prevent an infinite re-attempt-and-warn loop on every CA startup for ambiguous legacy-duplicate records, has zero coverage: a regression that removed the `continue` would pass the full suite undetected.
  </details>

- 🧪 **New CRL-failure-joined-with-prune-error path is untested** ([`internal/ca/cleanup.go:122`](https://github.com/voxpupuli/openvox-ca/blob/7ff62e7ce038ff9f10a10377730e31bfdab6e7b7/internal/ca/cleanup.go#L122))

  ```
  		// A CRL failure must not abort the rest: the "clean up NOW" rule above
  		// applies to the cert blobs just as much, and the deletion loop
  		// already logs-and-swallows its own per-subject failures.
  		crlErr := c.dropCRLEntriesLocked(cleanupCtx, removedSerials)
  ```

  💡 **Recommendation:** Add a test using a backend wrapper (the file already has the pattern via partialPruneBackend) that makes GetCRL/UpdateCRL fail independently of PruneEntries, and assert: CleanupExpiredCerts returns an error wrapping that CRL failure, removed still reports the pruned count, and the stored certificate for the pruned subject is still deleted despite the CRL step failing.
  **Constraint:** Coverage completeness: behavior change in core cleanup path with no test for the new branch.

  This diff changes CleanupExpiredCerts's control flow: previously a dropCRLEntriesLocked failure returned immediately, skipping the per-subject stored-cert deletion loop entirely. Now the error is captured as crlErr, the deletion loop always runs regardless, and the final return is errors.Join(pruneErr, crlErr) (line 131). I found no test in internal/ca/cleanup_test.go (read in full) that ever causes dropCRLEntriesLocked to return a non-nil error. Both new specs added by this PR inject failures only via the PruneEntries wrapper (partialPruneBackend); dropCRLEntriesLocked's own two failure sources - c.readStoredCRL(ctx) and c.signCRLLocked(ctx, ...) - are never forced to fail, so: (a) pruneErr == nil && crlErr != nil is unexercised, (b) both errors non-nil together (verifying errors.Join actually surfaces both) is unexercised, and (c) nothing proves the stored-cert deletion loop still runs when only the CRL step fails - the exact scenario the comment at lines 119-121 describes as the point of this change.
  </details>

</details>

- ⚙️ **Docs promise conversion progress logging the code doesn't emit** ([`docs/storage-backends.md:164`](https://github.com/voxpupuli/openvox-ca/blob/7ff62e7ce038ff9f10a10377730e31bfdab6e7b7/docs/storage-backends.md#L164))

  ```
  large fleet expect it to take a while (progress is logged).
  ```

  💡 **Recommendation:** Either add periodic progress logging inside replaceInventoryOnce's batch loop (mirroring rebuildCertIndex's progressEvery pattern) so the conversion phase is observable at fleet scale, or narrow the doc claim at docs/storage-backends.md:164 to state that only the projection backfill reports progress and the inventory conversion does not.
  **Constraint:** Operational documentation must describe actual observable behavior (Operator criterion 5/7: runtime observability, operational documentation)

  This sentence promises progress logging for both phases of the first post-upgrade start: "the inventory conversion and a per-certificate projection backfill". The backfill phase does log progress - internal/ca/certindex.go's rebuildCertIndex logs "Certificate index repair: backfill progress" every 1000 records (progressEvery = 1000, checked at line 127-130). The inventory-conversion phase does not: I grepped every slog.* call in internal/storage/etcd_inventory.go and internal/storage/etcd.go and the only log lines decomposeLegacyInventory/replaceInventoryOnce emit are "Resuming interrupted etcd inventory decomposition" (once, only when resuming a crashed import), a duplicate-serials warning (once, if applicable), and "Decomposed legacy etcd inventory blob into per-entry keys" (once, at the very end) - there is no per-batch or periodic log inside the batch loop in replaceInventoryOnce (internal/storage/etcd_inventory.go lines 765-799), which is exactly the loop that does the conversion work. At the fleet scale this PR targets (100k nodes, issue #138), a legacy blob decomposes at etcdImportBatch=32 entries per transaction - roughly 3,125 sequential etcd round-trips with zero intermediate log output - which is precisely the "apparent hang" scenario the backfill's own progress-logging comment (certindex.go lines 98-102) says it exists to avoid, but that reasoning was not applied to the conversion phase.
  </details>

- 🧪 **New heavily-concurrent etcd integration suite runs without -race** ([`magefile.go:710`](https://github.com/voxpupuli/openvox-ca/blob/7ff62e7ce038ff9f10a10377730e31bfdab6e7b7/magefile.go#L710))

  ```
  	return sh.RunV("go", "test", "-tags", "etcd_integration", "-count=1", "./internal/storage/...")
  ```

  💡 **Recommendation:** Add -race to the go test invocation for Test.BackendsEtcd.
  **Constraint:** testing_conventions (lang-go.md): "Race detection: `go test -race` (flag missing race detection in concurrent code tests)"

  internal/storage/etcd_integration_test.go, which this Test.BackendsEtcd mage target runs, is substantially new in this PR and heavily concurrent: EtcdInventoryConcurrentAppends spins up 4 goroutines writing through two StorageService/backend instances sharing an in-process etcd cluster; EtcdBackendAcquireLockSerialisesConcurrentCallers and EtcdBackendWithLockCrossBackend run 6 and 4 concurrent goroutines against shared atomic counters and a shared *clientv3.Client; EtcdLegacyInventoryDecompose's 'lets two replicas decompose concurrently' test races two *EtcdBackend instances' EnsureReady calls. None of this runs under the race detector - this invocation has no -race flag (and none of the sibling backend suites - redis_integration, postgres_integration, mysql_integration, openbao_integration - enable it either, confirmed by grepping every go test invocation in magefile.go). This PR substantially increases the concurrent code exercised by exactly this un-raced suite (new mutation paths, new pruneBatchHook/importBatchHook seams, new per-entry-key CAS loops), so a data race introduced by the decomposition work would not be caught by CI.
  </details>

</details>

---

bootc:
Context for this PR: moving etcd onto `InventoryStore` also closes the etcd half
of #204, which turned out to be a live CI flake on the Redis suite rather than a
theoretical gap — mechanism and repro in
[this comment](https://github.com/voxpupuli/openvox-ca/issues/204#issuecomment-5258671756).

The short version: the blob path writes a whole-blob HMAC computed from the bytes
it read before its own append, so two replicas appending concurrently can leave
the stored head covering a blob that no longer exists; the next verifying read
(revocation, OCSP, the renewal serial lookup) then fails with
`ErrInventoryTampered`. The hash-chain head this PR moves etcd to advances
atomically with the row, so the window disappears.

Nothing to change here as far as I can see — just worth capturing in the PR
description or the storage-internals wording, since after this lands Redis is the
only distributed backend still on the racy scheme (#139).

bootc:
Replying to the [review council comment](https://github.com/voxpupuli/openvox-ca/pull/154#issuecomment-5253553320) — all five findings are now addressed, the last three in 1485755:

- **`mutateRecordBySerial` concurrency (HIGH)**: added in 1485755. A four-goroutine, two-replica race over one serial (mirroring `EtcdInventoryConcurrentAppends`) proves every `SetRevoked`/`ClearRevoked`/`SetProjection` lands atomically with the chain undisturbed, and a new `mutateRecordHook` seam drives both the conflict-then-succeed path (injected ModRevision bump, retried write asserted to land) and retry exhaustion (persistent bumps surface "too many concurrent writers").
- **`rebuildCertIndex` unknown-state skip (HIGH)**: already closed in 91d2240 — `internal/ca/certindex_test.go` drives `Init` over a relabelled `CertStateUnknown` record via a wrapper backend and asserts no index write targets it (writes log), it is excluded from the backfill count, a sibling is still repaired, and the stored row is byte-identical afterwards (snapshotted through the unwrapped backend, per a follow-up in 5794220).
- **CRL-failure path in `CleanupExpiredCerts` (HIGH)**: closed in 5794220 (CRL write fails after a successful prune: blobs still deleted, count still reported, error surfaced) and completed in 1485755 with the both-errors case — prune and CRL failures in one pass both survive the `errors.Join` and the blob cleanup still runs.
- **Conversion progress logging (MEDIUM)**: 1485755 makes the doc claim true rather than narrowing it — the import batch loop now logs every 1024 entries, so a fleet-scale conversion reads as a moving import instead of thousands of silent transactions.
- **`-race` on the etcd suite (MEDIUM)**: 1485755 adds `-race` to `Test.BackendsEtcd` (with `CGO_ENABLED=1` explicitly, since the repository default is `CGO_ENABLED=0`). The full suite — including the new contention specs — passes under the detector. The sibling backend suites still run without it; happy to follow up there separately if wanted.

Worth noting your review ran at 7ff62e7, and the branch has moved on substantially since:

- **Rebased onto main** after #151 merged. The statuses handler deriving every indexed record's state from the signed CRL superseded this branch's own handler-side unknown-state resolution (dropped in the rebase); the storage-side `CertStateUnknown` remains as the contract for backend consumers, and the repair pass skips such records. The lock documentation moved to the now-authoritative `docs/development/locking.md`, which also picked up the `inventory-decompose` row and, later, the backend-internal lock-name convention.
- **Three further local review-council rounds** (9 → 6 → 1 findings, all fixed, ending in a unanimous APPROVE): the ambiguity sentinel now survives partial prunes — a by-serial key is released only when the last bearer of a duplicated legacy serial is pruned, with `Statuses` deriving ambiguity from the sentinel itself; the 900-entry prune bound became a genuine per-call budget threaded across fence-conflict retries; `Statuses` reads the entry range and the sentinels in one transaction, closing a torn-snapshot window; and the removed-alongside-error contract gained producer-side specs on both the blob fallback and the CRL-failure arm.
- Separately, this branch closes the etcd half of #204 (context in [this comment](https://github.com/voxpupuli/openvox-ca/pull/154#issuecomment-5258678249)): the append transaction commits the entry and its integrity head atomically, so the two-write HMAC race that bites the blob backends cannot occur on etcd.

Head is 1485755; all suites (including etcd under `-race`) and lint are green there.

trevor-vaughan-ai:
## 🟢 Review Council: APPROVE

>
> Models used:
> - claude-sonnet-5

_Reviewed at commit [`f6a5760`](https://github.com/voxpupuli/openvox-ca/commit/f6a576046d4c8ad3229e32e2e023250fa7268530) (standard effort)._

**TL;DR:** Five-persona review approved the PR outright; only two confirmed, low-severity documentation gaps surfaced, no code defects found.

**Findings:** 🔴 0 Critical, 🟠 0 High, 🟡 1 Medium, 🔵 1 Low

|---|---|---|
| 📚 Curator (code) | ✅ Approve | 1 MEDIUM, 1 LOW |

- 📚 **InventoryStore interface code sample omits PruneEntries** ([`docs/development/inventory-store.md:79`](https://github.com/voxpupuli/openvox-ca/blob/f6a576046d4c8ad3229e32e2e023250fa7268530/docs/development/inventory-store.md#L79))

  ```
      // LatestSerialForSubject returns the most recently issued serial for
      // subject, wrapping fs.ErrNotExist when the subject has no entry.
      LatestSerialForSubject(ctx context.Context, subject string) (string, error)
  }
  ```

  💡 **Recommendation:** Add a PruneEntries entry (with a short doc comment, mirroring backend.go's) to the InventoryStore interface code sample in the "Optional capability interface" section of docs/development/inventory-store.md so the reference definition matches internal/storage/backend.go.
  **Constraint:** Documentation completeness: a reference doc's illustrative interface definition must match the actual interface it documents

  This PR adds ~90 new lines to this same file (the "### The etcd decomposition" section) describing PruneEntries's batching, partial-completion, and returned-entries-on-error contract in detail, and changes PruneEntries's signature itself (whole-fold callback -> step-function advanceHead) in internal/storage/backend.go. But the canonical `InventoryStore interface { ... }` code sample near the top of this same file (## Design > Optional capability interface) only lists AppendEntry, Entries, and LatestSerialForSubject -- it never gained a PruneEntries entry, even though the interface in internal/storage/backend.go has four methods (AppendEntry, Entries, LatestSerialForSubject, PruneEntries). A reader following the doc top-down sees an interface that appears to have three methods, then reads extensive prose about a fourth method's contract for a method that, per the code sample, does not exist. A future backend implementer (e.g. the planned Redis decomposition, issue #139) copying this code sample as a template would miss PruneEntries entirely.
  </details>

</details>

- 📚 **AGENTS.md locking summary not updated for the new backend-internal lock category** ([`docs/development/locking.md:40`](https://github.com/voxpupuli/openvox-ca/blob/f6a576046d4c8ad3229e32e2e023250fa7268530/docs/development/locking.md#L40))

  ```
  Backend-internal locks live beside the backend that takes them, for the same
  import-direction reason: `etcdDecomposeLockName` (`"inventory-decompose"`) in
  [etcd_inventory.go](../../internal/storage/etcd_inventory.go). They are no
  less protocol for it.
  ```

  💡 **Recommendation:** Add a one-line note to AGENTS.md's "Locking and concurrency" section acknowledging that backend-internal locks (taken via Backend.AcquireLock directly, e.g. etcd's inventory-decompose) are a second recognised pattern alongside the CA-layer StorageService.WithLock names, pointing to locking.md for detail.
  **Constraint:** Project context document (AGENTS.md) should stay in sync with the reference docs it summarizes

  This PR introduces a new class of lock into locking.md: a "backend-internal" lock (`inventory-decompose`) taken directly via `Backend.AcquireLock` rather than through `StorageService.WithLock`, and locking.md (changed by this PR) documents it in both prose and the lock-name table. AGENTS.md at the repo root -- which explicitly states "This file is authoritative for repository conventions" and gives what it calls "the short form" of locking.md's rules for any AI or human agent touching shared CA state -- still describes the whole locking system as "Mutations serialise on cluster-wide named locks via StorageService.WithLock (bootstrap, crl, subject:<name>)", with no mention of the new backend-internal category or its different acquisition path. A contributor or AI agent reading only AGENTS.md's short form (not this PR's changed locking.md) would not learn that a second lock-acquisition pattern now exists in the codebase.
  </details>

</details>

---

bootc:
Replying to the [16 August review](https://github.com/voxpupuli/openvox-ca/pull/154#issuecomment-5308656778) — both findings were real and are fixed in 06fe20a:

- **`InventoryStore` sample omits `PruneEntries` (MEDIUM)**: the code sample in `docs/development/inventory-store.md` now carries a fourth, abridged `PruneEntries` entry matching the actual signature in `backend.go` (the `advanceHead` step function) and summarising the consistency/bounding/returned-alongside-error contract, deferring to `backend.go` for the full text. The three-method sample predated the interface gaining the method and was exactly the template a future implementer — the planned Redis decomposition included — would have copied incomplete.
- **AGENTS.md locking short form (LOW)**: the `WithLock` bullet now names the backend-internal `Backend.AcquireLock` pattern this branch introduced (etcd's `inventory-decompose`) as a second recognised acquisition path, pointing at `locking.md` for when each applies.

No further review-council loop was run for this commit: it is two documentation edits with no code, test, or behaviour change, so a five-persona pass would be ceremony. Head is 06fe20a; the branch remains rebased on current main and mergeable.

trevor-vaughan (approved):

PR #157, opened 2026-07-27
Add a Helm chart, released and versioned with the project

bootc:
Adds a Helm chart for deploying openvox-ca on Kubernetes, released in lockstep
with the binaries and published to GHCR as an OCI artefact.

> Stacked on #142 — this targets `feature/version-stamping`, because it extends
> the release machinery that PR introduces (the version constant, the shared tag
> gate, `release:prepare`). Review it after #142, or read only the commits above
> that branch.

## The chart

`charts/openvox-ca/` renders a Deployment, Service, ConfigMap, ServiceAccount
and PVC, plus opt-in Ingress, Gateway API routes, ServiceMonitor,
NetworkPolicy, PodDisruptionBudget, HPA and Kubernetes-export RBAC.

The central design decision is that the chart **does not mirror the server's
configuration surface**. `config` is written verbatim to
`/etc/puppet-ca/config.yaml`, so every key in `docs/configuration.md` —
including the nested `kubernetes_export` and `openbao` blocks — is reachable
with no chart value of its own, and stays reachable as the server grows more.
The convenience blocks (`tls`, `ca`, `metrics`, `kubernetesExport`,
`persistence`, …) do the Kubernetes half of a feature — mount the Secret, open
the port, create the RBAC — *and* set the config keys pointing at whatever they
mounted. `config` is merged over the top, so an explicit key always wins.

Other choices worth a reviewer's attention:

- **Alpine is the default image.** An empty `image.tag` resolves to
  `<appVersion>-alpine`; an explicit tag is used verbatim, which is how the
  CentOS Stream variant is selected without a dedicated value for the choice.
  A `-dev` appVersion resolves to the rolling `edge` tag, because that is what
  an unreleased tree actually has published.
- **`cadir` is a retained PVC by default.** The filesystem backend keeps the
  CA's private key there, so silently regenerating it on restart would be far
  worse than a pending PVC on a cluster with no default StorageClass. That also
  dictates `strategy: Recreate`, since a surging pod cannot mount a
  ReadWriteOnce volume.
- **TLSRoute, not HTTPRoute**, is the Gateway API route to reach for: openvox-ca
  terminates TLS itself and authenticates agents by client certificate, so a
  terminating proxy silently disables every admin endpoint. Both are available,
  with the trade-off documented at each.
- **A server certificate is required.** The server refuses to serve plain HTTP
  on a non-loopback address, so the chart refuses to render that install rather
  than handing over a crash-looping pod, naming the same remedies the binary's
  own error does.

## Versioning and publishing

`Chart.yaml`'s `version` and `appVersion` are both the `internal/version`
constant. `mage release:prepare` rewrites them alongside `version.go`, and four
places refuse a mismatch: `mage chart:version`, the shared `verify-release-tag`
action, the pre-push hook, and the publish workflow.

`helm-chart.yml` pushes to `oci://ghcr.io/<owner>/<repo>-charts` — a package of
its own, since one GHCR package cannot hold both container images and a chart.
Release tags publish behind the same verify gate as the other two tag
workflows, and wait for the tag's alpine image to exist first, so a published
chart can never name an image that does not. Pushes to `main` republish the
rolling `0.9.0-dev` chart; anything else is a version somebody means to tag, and
is skipped with a notice.

## What CI checks

A new `chart` job, wired into the `CI success` fan-in the tag gate depends on,
running against **both Helm 3.21.3 and 4.2.3**:

- `mage chart:version` — the chart still tracks `internal/version`
- `mage chart:validate` — six fixtures rendered and checked against the real
  Kubernetes and CRD schemas, with the minimal fixture additionally held to the
  `kubeVersion` floor
- `mage chart:test` — 38 assertions on what the templates actually resolve to
  (tag resolution, merge precedence, which kind `rbac.scope` selects, probe
  scheme, RBAC narrowing) and on each precondition refusing what it claims to
- `mage chart:package` — the packaging step the publish workflow performs

## Review history

Two automated review passes ran against this branch; their findings are fixed
in the last three commits and the fixes are asserted, not just made. The
substantive ones were: a default install that crash-looped while NOTES reported
success; a publish path that would have shipped a release-numbered chart from
`main` outside the verify gate; a TLS precondition that was wrong in both
directions (`[::1]` accepted though the server rejects it, TLS via `env`
refused though it works); export RBAC that failed open with no targets
configured; and an assertion that hard-coded the `-dev` image tag and so would
have gone red on the exact commit `release:prepare` produces.

## Before merging

The publish path cannot be exercised by CI. It needs a fork rehearsal —
dispatching the *Helm chart* workflow with `push=true` — which also confirms
the new image-wait step. The GHCR chart package is created private on first
push and needs making public by hand, as for the container images.


trevor-vaughan-ai:
> **Obsolete.** Superseded by the [current Review Council verdict](https://github.com/voxpupuli/openvox-ca/pull/157#issuecomment-5308572731) for commit `7886994`. 

>
> Models used:
> - claude-sonnet-5

_Reviewed at commit [`f21237f`](https://github.com/voxpupuli/openvox-ca/commit/f21237f1507a5d0ec831fa288913b45130b2f8e4) (standard effort)._

**TL;DR:** Review Council requests changes on PR #157's Helm chart addition: 4 of 5 personas approved, but the curator persona's 4 verified documentation/NOTES.txt findings (plus 8 more test-coverage and consistency findings elsewhere) keep the council from a clean approve; nothing found is CRITICAL or HIGH.

**Findings:** 🔴 0 Critical, 🟠 0 High, 🟡 10 Medium, 🔵 3 Low

|---|---|---|
| 🛡️ Adversary (code) | ✅ Approve | 1 MEDIUM |
| 📚 Curator (code) | ❌ Changes | 2 MEDIUM, 2 LOW |
| 🧪 Tester (code) | ✅ Approve | 5 MEDIUM, 1 LOW |

<details><summary>🟡 MEDIUM (10)</summary>

- 🛡️ **Unrestricted patch on all Secrets/ConfigMaps when export targets are unknown, with no operator-facing warning** ([`charts/openvox-ca/templates/rbac.yaml:40`](https://github.com/voxpupuli/openvox-ca/blob/f21237f1507a5d0ec831fa288913b45130b2f8e4/charts/openvox-ca/templates/rbac.yaml#L40))

  ```
    {{- else if $unknownTargets }}
    # The export config lives in a ConfigMap the chart does not render, so the
    # target names are unknown and patch cannot be narrowed. Grant it in full,
    # visibly, rather than appearing to restrict something.
    - apiGroups: [""]
      resources: ["secrets", "configmaps"]
      verbs: ["patch"]
    {{- end }}
  ```

  💡 **Recommendation:** Either fail() the existingConfigMap + kubernetesExport.rbac.create combination the same way the default-ServiceAccount case is failed, or at minimum surface a NOTES.txt WARNING (mirroring the pattern already used for the other risky combinations in openvox-ca.notes) so the operator sees the unrestricted-patch grant at install time instead of only in a template comment.
  **Constraint:** Adversary RBAC review: avoid overly broad permissions (wildcard resourceNames, cluster-wide bindings where namespace-scoped would do) - CWE-269 Improper Privilege Management / OWASP A01:2021 Broken Access Control

  When kubernetesExport.enabled + kubernetesExport.rbac.create are true and the operator sets existingConfigMap (openvox-ca.exportTargetNames returns "unknown", _helpers.tpl:749-761), the exporter's Role/ClusterRole gets `patch` on every Secret and ConfigMap in scope with no resourceNames filter - i.e. RBAC treats the absent list as "every resource" (as the chart's own comment at rbac.yaml:48-52 acknowledges). With kubernetesExport.rbac.scope: ClusterRole this becomes patch on every Secret/ConfigMap in the entire cluster; even with the default Role scope it is unrestricted patch within the bound namespaces. The chart already treats an analogous RBAC risk as fatal - _helpers.tpl:342-345 hard-fails install when kubernetesExport.rbac.create would bind to the namespace's default ServiceAccount ("granting create/patch on Secrets to every pod in the namespace") - but this comparably broad grant is only a rendered code comment, invisible to `helm install` output and absent from NOTES.txt (grep of openvox-ca.notes in _helpers.tpl shows warnings for missing TLS, ephemeral storage, multi-replica filesystem backend, autosign=true, and missing NetworkPolicy for metrics, but none for this RBAC path).
  </details>

- 📚 **NOTES.txt never warns about HTTPRoute/Ingress breaking mTLS, unlike every other security-relevant deployment state** ([`charts/openvox-ca/templates/_helpers.tpl:481`](https://github.com/voxpupuli/openvox-ca/blob/f21237f1507a5d0ec831fa288913b45130b2f8e4/charts/openvox-ca/templates/_helpers.tpl#L481))

  ```
  {{- if not (or .Values.puppetServers (dig "puppet_server" "" $config) (dig "puppet_server_file" "" $config)) }}

  NOTE: no puppetServers are listed, so no CN is granted admin API access over
  mTLS. Your OpenVox/Puppet Server compilers need to be listed here (or via
  config.puppet_server) before they can sign, revoke, or list certificates.
  {{- end }}

  Full configuration reference: https://github.com/voxpupuli/openvox-ca/blob/main/docs/configuration.md
  Chart documentation:          https://github.com/voxpupuli/openvox-ca/blob/main/docs/helm-chart.md
  {{- end -}}
  ```

  💡 **Recommendation:** Add a NOTES.txt block (in the openvox-ca.notes define) that fires when `.Values.gateway.httpRoute.enabled` is true, warning that HTTPRoute terminates TLS at the Gateway and disables client-certificate authentication, mirroring the existing WARNING blocks' tone and the guidance already given in docs/helm-chart.md around TLSRoute vs HTTPRoute.

  The `openvox-ca.notes` template (charts/openvox-ca/templates/_helpers.tpl, lines 773-848; rendered verbatim as charts/openvox-ca/templates/NOTES.txt) walks through every security-relevant deployment characteristic and warns the operator at `helm install` time: missing TLS, ephemeral filesystem/sqlite storage, multi-replica filesystem/sqlite, autosign=true, unrestricted metrics, restricted egress, and missing puppetServers (the block quoted above is the last of these, immediately followed by the notes' closing footer). The notes template itself never references `.Values.ingress`, `.Values.gateway.httpRoute`, or `.Values.gateway.tlsRoute`, even though `ingress.yaml`, `httproute.yaml`, and `tlsroute.yaml` are templated resources in this same batch. `httproute.yaml`'s own inline comment states the operationally critical fact that an HTTPRoute makes the Gateway terminate TLS so 'openvox-ca never sees the agent's client certificate and every mTLS-authenticated endpoint stops authenticating' - exactly the class of install-time gotcha the other NOTES.txt warnings exist to surface. An operator who enables `gateway.httpRoute` gets no equivalent warning at install time, breaking the internal convention this template otherwise follows consistently.
  </details>

- 📚 **AGENTS.md mage-target table omits chart:package, which ci.yml's new chart job invokes directly** ([`AGENTS.md:27`](https://github.com/voxpupuli/openvox-ca/blob/f21237f1507a5d0ec831fa288913b45130b2f8e4/AGENTS.md#L27))

  ```
  | `mage chart:version` | Assert `charts/openvox-ca/Chart.yaml` still tracks `internal/version` |
  ```

  💡 **Recommendation:** Add a row for `mage chart:package` to the mage-target table in AGENTS.md, describing that it packages charts/openvox-ca into a .tgz (as CI now does before the Helm chart workflow re-packages and pushes it).

  AGENTS.md's 'Common commands' table documents `mage chart:validate`, `mage chart:test`, and `mage chart:version` (lines 25-27) as the reference for the Helm chart's mage targets. `.github/workflows/ci.yml`'s new `chart` job in this batch runs a fourth target directly - `run: mage chart:package` (step 'Package the chart') - that has no row in this table. `Package()` is a real, distinct method on the `Chart` mg.Namespace alongside `Version()`, `Lint()`, `Validate()`, and `Test()`, so this isn't a naming variant of an already-documented target. The table's established convention (three of five Chart namespace targets already listed, one per row) is broken by silently omitting the one CI now exercises.
  </details>

- 🧭 **Route backend/port selection logic duplicated verbatim across two templates** ([`charts/openvox-ca/templates/tlsroute.yaml:2`](https://github.com/voxpupuli/openvox-ca/blob/f21237f1507a5d0ec831fa288913b45130b2f8e4/charts/openvox-ca/templates/tlsroute.yaml#L2))

  ```
  {{- $backend := ternary "metrics" "https" (eq .Values.gateway.tlsRoute.backendPort "metrics") }}
  {{- $port := ternary .Values.metrics.port (.Values.service.port | int) (eq $backend "metrics") }}
  ```

  💡 **Recommendation:** Extract the $backend/$port computation into a helper in _helpers.tpl (e.g. openvox-ca.routeBackend / openvox-ca.routePort taking the relevant backendPort value as an argument) and call it from ingress.yaml, httproute.yaml, and tlsroute.yaml, consistent with how probeScheme and the other decision helpers are already factored out.

  The two-line backend/port resolution in tlsroute.yaml:2-3 is byte-for-byte identical to httproute.yaml:2-3, differing only in which Values key is read (.Values.gateway.tlsRoute.backendPort vs .Values.gateway.httpRoute.backendPort). ingress.yaml:2 computes the same $backend ternary a third time for its own backendPort field. _helpers.tpl already establishes the chart's own pattern for exactly this kind of decision logic - openvox-ca.probeScheme, openvox-ca.needsAPIAccess, openvox-ca.tlsConfigured, and openvox-ca.configFullyKnown all centralize a piece of cross-cutting logic into one named template rather than repeating it per resource. The route backend/port selection breaks that established pattern: it is the same maintenance-sensitive logic (which Service port a route or ingress targets) duplicated instead of extracted, so a future change to the metrics-vs-https selection rule (e.g. adding a third backend option) has to be made in two or three places and can silently drift if one is missed.
  </details>

- ⚙️ **kubeconform version pin in CONTRIBUTING.md is not covered by the Renovate custom manager that tracks it everywhere else** ([`CONTRIBUTING.md:22`](https://github.com/voxpupuli/openvox-ca/blob/f21237f1507a5d0ec831fa288913b45130b2f8e4/CONTRIBUTING.md#L22))

  ```
    (`go install github.com/yannh/kubeconform/cmd/kubeconform@v0.7.0`)
  ```

  💡 **Recommendation:** Either add `CONTRIBUTING.md` to the customManager's `managerFilePatterns` so Renovate keeps all three occurrences in lockstep, or have CONTRIBUTING.md reference the version symbolically (e.g. point at `mage chart:validate`'s own error message) instead of repeating the literal pin.

  The same pinned string `go install github.com/yannh/kubeconform/cmd/kubeconform@v0.7.0` is duplicated in three places: `.github/workflows/ci.yml:99`, `magefile.go:681`, and `CONTRIBUTING.md:22`. This PR adds a Renovate customManager (renovate.json:41-53) whose `managerFilePatterns` is `["/^\\.github/workflows/.*\\.ya?ml$/", "/^magefile\\.go$/"]` - it does not include `CONTRIBUTING.md`. The first time Renovate bumps kubeconform, `ci.yml` and `magefile.go` will move together but `CONTRIBUTING.md` will silently keep advertising the old version to contributors, who will then install a kubeconform release that may diverge in behavior (schema validation output, exit codes) from what CI and `mage chart:validate` actually enforce.
  </details>

- 🧪 **chartConfigChecksum swallows render failures into pass-through sentinel strings** ([`magefile.go:742`](https://github.com/voxpupuli/openvox-ca/blob/f21237f1507a5d0ec831fa288913b45130b2f8e4/magefile.go#L742))

  ```
  func chartConfigChecksum(sets ...string) string {
  	out, err := helmTemplate(sets)
  	if err != nil {
  		return "<render failed>"
  	}
  ```

  💡 **Recommendation:** Have chartConfigChecksum return (string, error) and fail the case explicitly (increment `failures`, print a FAIL line) when the second render errors or the checksum cannot be found, rather than encoding the failure as a magic string fed back into the assertion it is supposed to be checking.

  chartConfigChecksum is used exactly once, at line 982, to build the `notWants` value for the "a config change rolls the pods" case: `notWants: []string{chartConfigChecksum(tls, "config.crl_validity_days=7")}`. If that second helm render fails (or the checksum regex fails to match, handled the same way two lines below), the function does not propagate the error - it returns the literal string "<render failed>" (or "<no checksum rendered>"), which is then used as the thing the primary render's output must *not* contain. Since neither sentinel string can ever appear in real helm output, the assertion passes trivially whenever the helper's own render breaks, instead of failing loudly and pointing at the real problem. This is exactly the 'test helper swallows an error' pattern: a regression that breaks rendering with the second override, or changes the checksum annotation's format, degrades this case to a silent no-op rather than a reported failure.
  </details>

- 🧪 **Lowercase kubernetesExport target kind is schema-legal but never rendered or asserted on** ([`charts/openvox-ca/values.schema.json:164`](https://github.com/voxpupuli/openvox-ca/blob/f21237f1507a5d0ec831fa288913b45130b2f8e4/charts/openvox-ca/values.schema.json#L164))

  ```
  "kind": { "type": "string", "enum": ["Secret", "ConfigMap", "secret", "configmap"] },
  ```

  💡 **Recommendation:** Add a chart:test render case (or a ci/ fixture) that sets a target with a lowercase kind and asserts it renders the same Secret/ConfigMap resource as the capitalized form, or narrow the schema enum to the forms the chart is actually documented to accept.

  The schema explicitly accepts `kind: secret` / `kind: configmap` (lowercase) as valid values for kubernetesExport.targets[].kind, alongside the capitalized Kubernetes kind names. Every fixture and every chart:test render/reject case uses only `Secret` or `ConfigMap`. Because the schema is the only place in this batch that names the lowercase forms, and nothing in ci/ or magefile.go's Chart.Test renders a target with `kind: secret`, there is no coverage proving the chart actually normalizes or otherwise correctly handles the case the schema was written to allow - a template-side regression in that handling (e.g. a case-sensitive comparison picking the wrong resource kind) would pass CI undetected.
  </details>

- 🧪 **envFrom's documented TLS-precondition bypass has zero test coverage** ([`magefile.go:1072`](https://github.com/voxpupuli/openvox-ca/blob/f21237f1507a5d0ec831fa288913b45130b2f8e4/magefile.go#L1072))

  ```
  		{
  			name:    "no TLS on a non-loopback address, which the server refuses to serve",
  			sets:    []string{},
  			wantErr: "refuse to start",
  		},
  ```

  💡 **Recommendation:** Add a chartRenderCase for envFrom (no tls set, envFrom pointing at a Secret) asserting the render succeeds, and a comparable deliberate case for args, matching the pattern already used for existingConfigMap at magefile.go:1023.

  docs/helm-chart.md:153-154 and values.yaml:269 both document that `existingConfigMap`, `args` and `envFrom` each cause the chart to stop checking the TLS precondition, because the chart can no longer see the full configuration. No chart:test case exercises this for `envFrom` at all - no render case sets envFrom without tls to prove the precondition is correctly skipped. `args` fares only slightly better: it is set (with no tls block) in the ci/existing-config-values.yaml fixture consumed by chart:lint/chart:validate, so a total breakage of the bypass would fail CI, but no chart:test case deliberately asserts on it. A regression that made the check fire unconditionally when envFrom supplies TLS via the environment, or that silently stopped firing outside the three documented modes, would go uncaught.
  </details>

- 🧪 **New chart version/appVersion tag gate has no automated test** ([`.lefthook/pre-push/verify-tags.sh:30`](https://github.com/voxpupuli/openvox-ca/blob/f21237f1507a5d0ec831fa288913b45130b2f8e4/.lefthook/pre-push/verify-tags.sh#L30))

  ```
  if [ "$chart_version" != "$want" ] || [ "$chart_app_version" != "$want" ]; then
  ```

  💡 **Recommendation:** Add a lightweight bats/shellspec (or a Go test that shells out to the script against fixture git commits) covering: matching version/appVersion (allows push), mismatched chart version (refuses with the chart-specific message), and a commit where Chart.yaml is missing or malformed (parses to empty strings and still refuses rather than passing silently).
  **Constraint:** Coverage completeness: new functional logic in a diff should carry test coverage, not rely solely on manual exercise

  This diff adds a second verification block to verify-tags.sh (lines 27-35) that parses charts/openvox-ca/Chart.yaml out of the tagged commit via `git show` and refuses the push if version/appVersion don't match the tag. No shell-test harness, venom suite, or go test invokes or exercises this script - `test/` contains only compose/puppet/migration integration scripts, and `.lefthook.yml` only wires the script to run as a git hook, not as a test target. The only other place with equivalent logic (.github/actions/verify-release-tag/action.yml, the server-side gate this script mirrors) is likewise plain bash with no test coverage. Because this is untested shell logic manipulating `git show` output through several sed pipelines, a mistake in the parsing (e.g. quoting, the appVersion regex, or the multi-line `chart` variable capture) would silently pass or silently over-reject, and nothing in CI would catch it before a developer hits it at push time.
  </details>

- 🧪 **No render assertion locks down defaultMode: 0600 on the TLS/CA/passphrase secret volumes** ([`charts/openvox-ca/templates/deployment.yaml:234`](https://github.com/voxpupuli/openvox-ca/blob/f21237f1507a5d0ec831fa288913b45130b2f8e4/charts/openvox-ca/templates/deployment.yaml#L234))

  ```
  - name: tls
            secret:
              secretName: {{ .Values.tls.existingSecret }}
              # 0600 plus the pod's fsGroup: the kubelet ORs group access in when
              # fsGroup is set, so the container user can read it and nothing
              # else can.
              defaultMode: 0600
  ```

  💡 **Recommendation:** Add a chartRenderCase (or extend an existing tls-case) asserting `defaultMode: 0600` is present for the tls, ca, and ca-key-passphrase volumes when their respective existingSecret values are set.
  **Constraint:** Security test coverage: security-relevant code paths (file permissions on Secret-backed volumes holding the CA private key, its passphrase, and the server TLS key) should have tests exercising them

  The tls, ca, and ca-key-passphrase secret volumes (deployment.yaml lines 233-252) each set `defaultMode: 0600` with a comment explaining this is deliberate least-privilege permissioning for the CA's key material. The chart's test harness in magefile.go has no chartRenderCase in Chart.Test() that asserts `defaultMode: 0600` appears for any of these three volumes; the ci/ fixtures that set tls.existingSecret/ca.existingSecret/caKeyPassphrase.existingSecret (full-values.yaml, filesystem-values.yaml, postgres-ha-values.yaml) are only schema-validated by kubeconform, which accepts any valid integer or omitted defaultMode. A future edit that drops or changes this field (e.g. while refactoring the volumes block) would pass helm lint, kubeconform, and every existing chart:test case while silently loosening permissions on the CA's private key and TLS key material.
  </details>

</details>

<details><summary>🔵 LOW (3)</summary>

- 📚 **Chart namespace comment omits chart:test, the target this PR's own docs call "required"** ([`magefile.go:63`](https://github.com/voxpupuli/openvox-ca/blob/f21237f1507a5d0ec831fa288913b45130b2f8e4/magefile.go#L63))

  ```
  type Chart mg.Namespace   // chart:version  chart:lint  chart:validate  chart:package
  ```

  💡 **Recommendation:** Add `chart:test` to the comment on magefile.go:63, e.g. `// chart:version  chart:lint  chart:validate  chart:test  chart:package`.

  This PR adds the `Chart mg.Namespace` line and its enumerating comment (magefile.go, new in this diff). The comment lists four sub-targets but omits `chart:test`, even though `func (Chart) Test() error` is defined at magefile.go:886 in this same diff and is the specific target CONTRIBUTING.md:89-92, AGENTS.md:26, and charts/openvox-ca/README.md:525 all call a required CI check ("mage chart:validate and mage chart:test are required checks too"). The established convention in this file is that an mg.Namespace comment is a complete inventory of that namespace's targets - `Build mg.Namespace // build:all build:fips build:dist build:distVariant` (line 60) lists exactly its four `func (Build) ...` methods with none missing. The new `Chart` comment breaks that pattern for a target the PR itself treats as load-bearing.
  </details>

- 📚 **values.yaml header comment gets the physical layout backwards** ([`charts/openvox-ca/values.yaml:6`](https://github.com/voxpupuli/openvox-ca/blob/f21237f1507a5d0ec831fa288913b45130b2f8e4/charts/openvox-ca/values.yaml#L6))

  ```
  # every key in docs/configuration.md is reachable without the chart needing a
  # value of its own for each one. The blocks above it (tls, ca, puppetServers,
  # metrics, kubernetesExport, …) are conveniences that wire up the Kubernetes
  ```

  💡 **Recommendation:** Change "The blocks above it" to "The blocks below it" (or reorder the file so config comes last) in values.yaml:6.

  The file's own header comment says the convenience blocks (tls, ca, puppetServers, metrics, kubernetesExport) sit "above" `config` in the file. `config` is defined at values.yaml:65; tls, ca, metrics, and kubernetesExport are all defined further down (lines 130, 139, 187, 221 respectively) - below `config`, not above it. puppetServers (line 112) is also below config, just closer to it. A reader using this comment as a map of the file will look in the wrong direction. Neither the chart README's "How configuration works" section nor docs/helm-chart.md's "How the chart is configured" section repeats this directional claim - the error is isolated to this one comment.
  </details>

- 🧪 **Chart.Test reimplements a pass/fail runner instead of using the project's go-test convention** ([`magefile.go:880`](https://github.com/voxpupuli/openvox-ca/blob/f21237f1507a5d0ec831fa288913b45130b2f8e4/magefile.go#L880))

  ```
  // Test asserts what the chart actually renders, which neither `helm lint` nor
  // kubeconform can: both are satisfied by valid YAML carrying the wrong values.
  // These cases cover the logic a reader has to trust — image-tag resolution,
  // config merge precedence, probe scheme selection, which kind rbac.scope
  // picks — and the preconditions, which are only worth having if they really
  // fire.
  func (Chart) Test() error {
  ```

  💡 **Recommendation:** Consider moving the render/reject case tables and their execution into a real `_test.go` (e.g. behind a build tag, run via `go test -tags chart_integration`), with Chart.Test as a thin mage wrapper. Not blocking: the current approach is functionally sound and well-documented.

  Every other Go-level integration suite in this magefile is a thin mage wrapper around `go test` (e.g. BackendsEtcd, BackendsRedisGo), with the actual assertions living in `_test.go` files under go test's own framework. Chart.Test breaks that pattern: it hand-rolls its own table iteration, a manual `failures := 0` counter, and printf-based reporting instead of `t.Run`/`t.Errorf`. This loses standard tooling the rest of the suite gets for free - `go test -run`, `-v`, parallel execution, structured JSON output for `tparse` (already used by Test.Unit), and IDE/CI failure annotations tied to a specific subtest. Nothing here is incorrect, but it is a structural inconsistency, and 39+ hand-written cases is enough volume that the missing tooling is a real ergonomic cost.
  </details>

</details>

---

bootc:
All 12 actionable findings are addressed as of d3da555. The 13th (`Chart.Test`
hand-rolling its runner) is deliberately unchanged — reasoning at the bottom.
The branch is also rebased onto `main` now that #142 has merged, so it targets
`main` directly.

### MEDIUM

| Finding | Resolution |
|---|---|
| Unrestricted patch on Secrets/ConfigMaps with no operator-facing warning | Took the NOTES route rather than `fail()`, since `existingConfigMap` plus export is a legitimate configuration the chart simply cannot narrow. `openvox-ca.notes` now emits a WARNING naming the scope (`cluster-wide` or `namespace-wide`) and that patch is granted on every Secret and ConfigMap, so it appears at install time instead of only in a template comment. |
| NOTES never warns that HTTPRoute breaks mTLS | Added, firing on `gateway.httpRoute.enabled`, in the same register as the other warnings: the Gateway terminates TLS, so openvox-ca never sees the client certificate and every mTLS-authenticated endpoint stops authenticating. |
| AGENTS.md omits `chart:package` | Added, describing the `dist/openvox-ca-<version>.tgz` output and that CI runs it before the publish workflow packages for real. |
| Route backend/port logic duplicated across three templates | Extracted to `openvox-ca.routeBackendName` and `openvox-ca.routeBackendPort`, both taking the relevant `backendPort` as an argument. `ingress.yaml`, `httproute.yaml` and `tlsroute.yaml` all call them. |
| kubeconform pin in CONTRIBUTING.md outside the Renovate custom manager | Added `CONTRIBUTING.md` to `managerFilePatterns`, so all three occurrences move together. |
| `chartConfigChecksum` swallows render failures into sentinel strings | Now returns `(string, error)`. A failed render or an unmatched checksum fails the case explicitly instead of producing a string that can never appear in Helm output — which had made the assertion pass precisely when the helper broke. |
| Lowercase export target `kind` schema-legal but never exercised | Added a render case asserting a lowercase `kind` passes through as written. |
| `envFrom`'s TLS-precondition bypass untested | Added cases for both documented bypasses: `envFrom` and a replaced `args`. |
| Tag gate has no automated test | Done, and it was the most productive finding in the set — see below. |
| `defaultMode: 0600` not locked down | Asserted for all three key-material volumes, anchored to each `secretName` so the assertion cannot pass by matching a `defaultMode` elsewhere in the document. That mattered: the first version of this assertion was satisfied by the explanatory YAML comment sitting between the two lines, so it could not have failed. |

### LOW

| Finding | Resolution |
|---|---|
| `Chart` namespace comment omits `chart:test` | Added, so the comment is a complete inventory as with `Build`. |
| values.yaml header says "above" for blocks that are below | Corrected to "below". |

### On the tag-gate tests

This one repaid the effort well beyond the finding. The gate is now covered by
Ginkgo specs — matching tag, mismatched `internal/version`, mismatched chart
`version` and `appVersion`, an unparseable `Chart.yaml` failing closed rather than
passing on an empty parse, and a branch push being ignored — and writing them
exposed a defect no amount of reading would have found.

The fixtures built their throwaway repositories from `os.Environ()`. git exports
`GIT_DIR`, `GIT_WORK_TREE`, `GIT_INDEX_FILE` and `GIT_OBJECT_DIRECTORY` to the
hooks it runs, and those outrank a child process's working directory — so under a
real `git push`, every fixture `git add -A` and `git commit` operated on the
repository being pushed. Six commits named `fixture` landed on this branch and
moved it off the real work. Nothing was lost, but note that neither
`mage test:magefile` nor CI can reproduce it: with no hook there is no `GIT_DIR`
to inherit, which is why it survived five green review rounds.

So the specs plant the leaked environment deliberately now, against a decoy
repository, and require the gate to still *fail closed* — asserting the accept
path alone was useless, because the gate skips any commit it cannot resolve and
therefore exits silent, which is exactly what an accept assertion checks for.
There is also a spec running the script the way git actually does, with `GIT_DIR`
naming the repository being pushed; every other spec lets it discover the
repository from the working directory, which production never does.
`.lefthook.yml` clears what git exports before running the suite, and the rule is
in AGENTS.md for the next test that shells out to git.

### Not changed

`Chart.Test` still hand-rolls its table runner rather than living in a `_test.go`
behind a build tag. The recommendation is reasonable and the reasoning about lost
tooling (`-run`, `-v`, `tparse`, per-subtest annotations) is correct, but it is a
structural change to a harness that grew from 38 to 77 cases across this review
cycle, and doing it while the cases were still churning would have meant
rewriting it repeatedly. Worth doing as its own change, against a settled harness,
rather than folded in here — the finding itself notes it is not blocking.

One factual correction to the surrounding claim: the magefile's Ginkgo suite
(`mage test:magefile`) does now hold the chart's non-template logic — version
parsing, the pin cross-check, the tag gate — so the pattern is followed for
everything except the render cases themselves.

trevor-vaughan-ai:
## 🟢 Review Council: APPROVE

>
> Models used:
> - claude-sonnet-5

_Reviewed at commit [`7886994`](https://github.com/voxpupuli/openvox-ca/commit/7886994639517e05d6973602121359446b5fdc46) (standard effort)._

**TL;DR:** Council approves the Helm chart PR with three medium duplication/coverage findings, all confirmed, none blocking.

**Findings:** 🔴 0 Critical, 🟠 0 High, 🟡 3 Medium, 🔵 0 Low

|---|---|---|
| 🧭 Guard (code) | ✅ Approve | 2 MEDIUM |
| 🧪 Tester (code) | ✅ Approve | 1 MEDIUM |

- 🧭 **Export RBAC rules duplicated verbatim: ClusterRole vs Role** ([`charts/openvox-ca/templates/rbac.yaml:32`](https://github.com/voxpupuli/openvox-ca/blob/7886994639517e05d6973602121359446b5fdc46/charts/openvox-ca/templates/rbac.yaml#L32))

  ```
  rules:
    # create cannot be restricted by resourceNames — the object has no name yet
    # at admission time — but patch can, so overwriting an *existing* Secret is
    # held to the objects actually configured as export targets. Names come from
    # the merged config, so targets set through config.kubernetes_export count.
    - apiGroups: [""]
      resources: ["secrets", "configmaps"]
      verbs: ["create"]
    {{- if $names }}
    - apiGroups: [""]
      resources: ["secrets", "configmaps"]
      verbs: ["patch"]
      resourceNames:
        {{- toYaml $names | nindent 6 }}
  ```

  💡 **Recommendation:** Factor the `$rawNames`/`$unknownTargets`/`$names` resolution and the `rules:` block into a single named template (e.g. `openvox-ca.exportRules`) in `_helpers.tpl`, called once from each of the ClusterRole and Role branches, so the two scopes cannot drift apart.
  **Constraint:** Structural Coherence / Zero-Waste - established chart pattern (centralize shared decision logic in one place) not applied consistently

  The 28-line block that computes and renders the export Role/ClusterRole `rules:` (resolving `exportTargetNames`, branching on `$unknownTargets`/`$names`, and emitting the `create`/`patch` rules) is duplicated byte-for-byte between the ClusterRole branch (lines 32-63) and the Role branch (lines 101-132), instead of being factored into a shared named template the way the rest of the chart's decision logic is. A future change to the patch-narrowing rule applied to one branch and not the other would leave `rbac.scope: Role` and `rbac.scope: ClusterRole` installs granting different permissions for the same configuration - silently, since `chart:test` asserts each branch's current behavior but not that the two stay in lockstep. This is exactly the failure class the PR body names as the branch's recurring defect: "the recurring defect was one setting decided in more than one place." AGENTS.md's own new section for this PR states the opposite intent for chart logic: "Preconditions live in one place, the `openvox-ca.validate` helper" - the same centralization discipline was not applied here.
  </details>

- 🧭 **extraArgs --config scan duplicated across two predicates** ([`charts/openvox-ca/templates/_helpers.tpl:412`](https://github.com/voxpupuli/openvox-ca/blob/7886994639517e05d6973602121359446b5fdc46/charts/openvox-ca/templates/_helpers.tpl#L412))

  ```
  {{- $configOverridden := false -}}
  {{- range .Values.extraArgs -}}
  {{- $arg := . | toString -}}
  {{- if or (eq $arg "--config") (hasPrefix "--config=" $arg) }}{{ $configOverridden = true }}{{ end -}}
  {{- end -}}
  ```

  💡 **Recommendation:** Extract the extraArgs `--config` scan into its own named template (e.g. `openvox-ca.configOverriddenByExtraArgs`) and have both `configFileKnown` and `configFullyKnown` include it, so the detection logic has exactly one definition.
  **Constraint:** Structural Coherence / Zero-Waste - duplicated business logic in a security-sensitive predicate the PR itself flags as bug-prone

  The 5-line loop that scans `.Values.extraArgs` for a `--config`/`--config=` override is copied verbatim into both `openvox-ca.configFileKnown` (lines 412-416) and `openvox-ca.configFullyKnown` (lines 425-429), rather than factored into one shared sub-template both predicates call. The PR description states this exact detection ("a `--config` in `extraArgs`... each took a separate round to close") was already the site of repeated bugs during this branch's twelve review rounds. A future fix to the detection applied to only one of the two copies would leave `configFileKnown` and `configFullyKnown` disagreeing about whether extraArgs overrides the config file - reintroducing the same class of defect the PR spent multiple rounds closing.
  </details>

- 🧪 **chartValuesFiles' *.yml branch and empty-fixture guard are unexercised by any test** ([`magefile.go:732`](https://github.com/voxpupuli/openvox-ca/blob/7886994639517e05d6973602121359446b5fdc46/magefile.go#L732))

  ```
  for _, ext := range []string{"*.yaml", "*.yml"} {
  ```

  💡 **Recommendation:** Give chartValuesFiles() (or a thin wrapper) an injectable directory parameter, the same seam checkModuleTidy already uses, and add two Ginkgo cases: one with a temp dir holding only a *.yml fixture (proving the second glob pattern is live) and one with an empty temp dir (proving the "no fixture values files found" error fires).
  **Constraint:** Missing edge case / branch coverage for new code the changeset introduced (testing_conventions, lang-go.md); the function's own comment identifies the exact failure mode the missing test would catch.

  chartValuesFiles() is new in this changeset and globs both *.yaml and *.yml under charts/openvox-ca/ci/, with the comment two lines above this loop warning: "Both YAML extensions are collected: globbing only *.yaml would skip a *.yml fixture in silence, which is that same rubber stamp one file at a time." Its only test, Describe("chartValuesFiles") in magefile_chart_test.go (lines 59-74), asserts the returned list is non-empty and contains the floor fixture name, but never proves the *.yml pattern contributes anything: `ls charts/openvox-ca/ci/` shows all six fixtures (edge-cases-values.yaml, existing-config-values.yaml, filesystem-values.yaml, full-values.yaml, minimal-values.yaml, postgres-ha-values.yaml) use the .yaml extension, so that arm of the loop is never taken by any test run. The len(files)==0 error branch at magefile.go:739-741 ("no fixture values files found under %s") is likewise never triggered. Both gaps are structural rather than an oversight: chartValuesFiles() hardcodes chartDir/ci instead of taking an injectable directory the way checkModuleTidy takes one, so exercising either branch today would mean writing into the real charts/openvox-ca/ci/ directory from a test. A future edit that silently broke the *.yml arm would make chart:lint/chart:validate stop validating any .yml fixture added later, with nothing in the suite to catch it - precisely the "rubber stamp" failure the comment calls out.
  </details>

</details>

---

trevor-vaughan (approved):

PR #160, opened 2026-07-29
Let the CA run as an intermediate under an external root

bootc:
openvox-ca can only ever be its own root: `setup`, `import` and `migrate` emit no
certificate signing request, `openvox-ca-ctl import` demands a `--private-key`
that a Transit-held key can never yield, and `bootstrapCA` only self-signs. This
adds the missing half of the round trip, so the CA can run as an intermediate
under an external root while its private key stays wherever `ca_key_provider`
puts it.

Two new subcommands, both offline, both on the **server** binary rather than
`openvox-ca-ctl` — they must reach the storage backend and key provider named in
the server's own configuration, which `openvox-ca-ctl` cannot do (it reads a
different config schema and can only address a local filesystem directory).

```console
$ openvox-ca csr --hostname puppet.example.com --create-key --out ca-request.pem
$ # ... parent signs it, chain assembled nearest-first ...
$ openvox-ca import-ca-cert --cert-bundle signed-chain.pem
Imported CA certificate "Puppet CA: puppet.example.com" (2 certificates in chain)
```

`import-ca-cert` never needs the private key *material*. It proves the
certificate binds the key the configured provider holds and refuses otherwise —
that single check is what makes importing a key-less certificate safe. For
deployments where `ca_cert_file` mounts the certificate read-only from a
Kubernetes Secret, `--out` runs the same validation and writes the bundle to a
file for loading out of band.

### Behaviour changes worth review

- **`Init` now refuses to bootstrap when a CA key exists but the certificate
  does not**, for every `ca_key_provider` rather than only when one is
  configured. This is the state `csr --create-key` leaves behind while the
  parent signs; the previous behaviour silently generated a fresh key over the
  top and self-signed a root. A half-finished bootstrap is indistinguishable
  from an outstanding request from inside `Init`, so this does turn that case
  into a hard failure — it costs one deletion, where the other case cannot be
  undone at all.
- **`openvox-ca-ctl import` gained a stricter bundle contract.** `ImportCA` now
  delegates to the shared validator, so `--cert-bundle` must be a complete chain
  ordered nearest-first, ending at a self-signed root. A self-signed Puppet
  Server CA already satisfies this; one that ran under an external root may need
  its chain assembled first. Recorded in the migration guide.
- **Bundles carrying a private key are rejected** rather than silently filtered,
  and bundles are re-encoded from the parsed chain before storage. `ca_crt.pem`
  is stored world-readable and served unauthenticated at `GET /certificate/ca`,
  so anything in that file is published — and
  `bao write pki/intermediate/generate/exported ... format=pem_bundle` produces
  exactly the shape that would have leaked.
- **`GET /certificate/ca` and the CRL are unchanged.** This PR does not alter
  what is served, only what can be got into storage.

### Structure

The first two commits are refactors with no behaviour change: one shared subject
builder and bundle validator (replacing two divergent inline copies), and
`resolveRuntime`, which owns backend-spec construction and provider wiring for
every entry point. `withKeyProvider` is explicit at each call site because the
distinction is security-relevant — the frontend role must never construct a
provider of its own, and a caller that silently got one would open a second
authenticated session to the key backend.

The last commit addresses an automated review pass over the branch. Two findings
were real gaps between what the commands documented and what they did on one
path: `--out` skipped the key-binding proof it promised, and the orphan-key
guard did not cover the default provider. Both are closed, with specs that fail
without the fix.

Not taken from that review: rejecting `pathlen:0` at import. It permits issuing
end-entity certificates and forbids issuing further CAs, which is precisely this
CA's profile — refusing it would reject the most correctly-profiled sub-CA
certificate a parent can issue. `keyCertSign` and `cRLSign` are required when
the extension is present, and `pathlen:0` is pinned as accepted with a spec
recording why.

Docs move with the code: a new section in `operator-cli.md`, the end-to-end
procedure in `openbao-transit.md` (including the deliberate window where the
server will not start, and why), and the chain requirement in
`migrating-from-puppet-server.md`.


trevor-vaughan-ai:
## 🟢 Review Council: APPROVE

>
> Models used:
> - claude-sonnet-5

_Reviewed at commit [`2f06033`](https://github.com/voxpupuli/openvox-ca/commit/2f0603336e268fef4fc3e9959eb72d1c44b5001a) (standard effort)._

**TL;DR:** Council approved the sub-CA bootstrap PR; 8 verified findings remain (documentation drift, a DRY violation, a silenced error, and coverage gaps), none blocking, one unverifiable blog-suggestion retracted.

**Findings:** 🔴 0 Critical, 🟠 0 High, 🟡 7 Medium, 🔵 1 Low

|---|---|---|
| 📚 Curator (code) | ✅ Approve | 2 MEDIUM |
| 🧪 Tester (code) | ✅ Approve | 3 MEDIUM, 1 LOW |

<details><summary>🟡 MEDIUM (7)</summary>

- 📚 **Kubernetes csr Job example omits required --hostname/--create-key** ([`docs/openbao-transit.md:356`](https://github.com/voxpupuli/openvox-ca/blob/2f0603336e268fef4fc3e9959eb72d1c44b5001a/docs/openbao-transit.md#L356))

  ```
  args: ["csr"]
  ```

  💡 **Recommendation:** Add `--create-key` to the Job's `args` list (or add a comment stating the example targets a CA that already has a key), so the example matches its own inline comment and the plain CLI example three paragraphs above it.
  **Constraint:** Curator criterion 2, cross-reference/example consistency: documentation examples must reflect actual CLI behavior

  The plain-CLI walkthrough for "1. Create the key and request a certificate" (line 249) shows `openvox-ca csr --hostname puppet.example.com --create-key --out ca-request.pem` as the command for a fresh sub-CA bootstrap. The Kubernetes Job example presented immediately after, as the in-cluster way to run that same first step, instead sets `args: ["csr"]` with no `--create-key`. cmd/openvox-ca/csr.go's RunE returns the error "no CA key exists yet: pass --create-key to create one, or provision it out of band first" when BuildCSR reports ca.ErrKeyProviderKeyNotFound -- exercised by the Ginkgo spec "refuses without a key rather than silently creating one" in cmd/openvox-ca/csr_test.go. The Job's own inline comment above its cadir volumeMount already presupposes --create-key is used (it says --create-key writes the CA key to the cadir), so the example contradicts itself: the comment assumes the flag while args omits it. An operator following this Kubernetes recipe verbatim for a fresh sub-CA would have the Job fail immediately instead of producing a CSR. (--hostname is not part of this gap: it is a standard top-level config.yaml key the shared ConfigMap mounted by both the Deployment and the Job already supplies for normal server operation, and csr.go only overrides it when the flag is explicitly passed.)
  </details>

- 📚 **Doc cross-reference names the pre-rename `hasCAKey`, not `HasCAKey`** ([`docs/storage-backends.md:486`](https://github.com/voxpupuli/openvox-ca/blob/2f0603336e268fef4fc3e9959eb72d1c44b5001a/docs/storage-backends.md#L486))

  ```
  storage backend, and `hasCAKey` asks the provider directly — so a populated
  ```

  💡 **Recommendation:** Update the reference in docs/storage-backends.md to `HasCAKey`, matching the exported name in internal/ca/keyprovider.go.
  **Constraint:** Documentation cross-reference consistency (curator convention: doc references to code identifiers must point at identifiers that exist)

  This batch renames `(c *CA) hasCAKey` to the exported `HasCAKey` in internal/ca/keyprovider.go (doc comment and definition both updated) and updates its sole call site in internal/ca/init.go from `c.hasCAKey(ctx)` to `c.HasCAKey(ctx)`. docs/storage-backends.md's "With the CA key at a provider" section still names the old lowercase identifier `hasCAKey` when explaining why a populated Transit slot counts as present. `grep -rn "hasCAKey"` across the repo (Go source and Markdown) turns up only this one remaining reference -- everywhere else, including cmd/openvox-ca/csr.go's call site, already uses `HasCAKey`. A maintainer or contributor who greps the codebase for `hasCAKey` after reading this doc will not find the function it is describing.
  </details>

- 🧭 **writePublicFile re-implements the storage layer's atomic-write algorithm instead of reusing it** ([`cmd/openvox-ca/writefile.go:43`](https://github.com/voxpupuli/openvox-ca/blob/2f0603336e268fef4fc3e9959eb72d1c44b5001a/cmd/openvox-ca/writefile.go#L43))

  ```
  func writePublicFile(path string, data []byte) error {
  	dir := filepath.Dir(path)
  	tmp, err := os.CreateTemp(dir, filepath.Base(path)+".tmp*")
  	if err != nil {
  		return fmt.Errorf("creating a temporary file beside %s: %w", path, err)
  	}
  	tmpName := tmp.Name()
  	defer func() {
  		// No-op once the rename has succeeded; cleans up on every failure path.
  		_ = os.Remove(tmpName)
  	}()
  ```

  💡 **Recommendation:** Export the storage package's atomic-write helper (e.g. `storage.AtomicWriteFile` or a thin wrapper around it) and have cmd/openvox-ca/writefile.go call that instead of re-implementing the sequence, so both call sites share one crash-safety algorithm and cannot silently diverge on ordering (e.g. the Chmod-before-Sync vs Chmod-after-Close difference already present).
  **Constraint:** Structural Coherence / Zero-Waste (DRY as waste prevention)

  This PR (branch feature/sub-ca-bootstrap) adds cmd/openvox-ca/writefile.go's writePublicFile, which duplicates the temp-file-plus-rename algorithm (CreateTemp in the target dir, write, chmod, fsync, close, rename, best-effort parent-dir fsync) that internal/storage/filesystem.go already implements as `atomicWriteFile` (internal/storage/filesystem.go:271-315), used by every Save* method on StorageService (SaveCACert, SaveCAKey, SaveCAPubKey, UpdateCRL, etc.). writefile.go's own doc comment even names this: "The storage layer writes CA material the same way, and docs/storage-backends.md advertises that property; a file destined for the same place should not be weaker" (writefile.go:40-42) -- the author is aware of atomicWriteFile and chose to re-derive its logic rather than export and reuse it, because atomicWriteFile is unexported to package storage. The two implementations have already diverged in one detail: atomicWriteFile chmods the *file descriptor* before Sync/Close (filesystem.go:284), while writePublicFile chmods the *path* after Close (writefile.go:71) -- small evidence of the drift risk the DRY rule exists to prevent. Grep confirms no shared helper: `grep -rn "atomicWriteFile\|writePublicFile" --include=*.go .` finds exactly these two independent definitions, each with its own os.CreateTemp/Sync/Chmod/Rename sequence.
  </details>

- ⚙️ **bootstrapCA still swallows the CA public key write failure this PR fixed elsewhere** ([`internal/ca/init.go:445`](https://github.com/voxpupuli/openvox-ca/blob/2f0603336e268fef4fc3e9959eb72d1c44b5001a/internal/ca/init.go#L445))

  ```
  	pubKeyBytes, err := x509.MarshalPKIXPublicKey(key.Public())
  	if err == nil {
  		pubKeyPEM := pem.EncodeToMemory(&pem.Block{Type: "PUBLIC KEY", Bytes: pubKeyBytes})
  		_ = c.Storage.SaveCAPubKey(ctx, pubKeyPEM)
  	}
  ```

  💡 **Recommendation:** Apply the same fix bootstrapCA's sibling in caImport.go received: return (or at minimum slog.Warn) when MarshalPKIXPublicKey or SaveCAPubKey fails, rather than the bare `if err == nil { ... }` pattern that discards both error paths.
  **Constraint:** Runtime Observability / Backup & Recovery: I/O failures on documented storage artifacts must be surfaced, not silently discarded ("no error recovery for I/O" is a HIGH-Operator pattern; this instance is scoped down to MEDIUM since ca_pub.pem is a non-essential companion file, not the cert or key)

  This is the exact same anti-pattern the PR deliberately fixed in the analogous code path: the old importCAMaterial wrote the public key with `pubKeyBytes, err := x509.MarshalPKIXPublicKey(caKey.Public()); if err == nil { ... _ = store.SaveCAPubKey(...) }` (see the diff removing this at internal/ca/caImport.go's pre-image), and the new importCAMaterial (internal/ca/caImport.go:224-231) now propagates both the marshal error and the SaveCAPubKey error, wrapping the latter in incompleteImportError so the operator is told storage is inconsistent. bootstrapCA in init.go, edited by this same PR just above this block (the marshalCAKeyForStorage extraction ends at line 434), still uses the discarded-error form: if MarshalPKIXPublicKey or SaveCAPubKey fails, bootstrapCA proceeds as if nothing happened and returns nil at the end (line 486), logging "CA bootstrapped" as success. docs/storage-backends.md, docs/configuration.md, and docs/development/storage-internals.md all document `ca_pub.pem` as a fixed part of the CA storage layout; a bootstrap that silently fails to create it leaves that documented layout incomplete with zero operator-visible signal (no log line, no error, no metric).
  </details>

- 🧪 **Post-key-creation BuildCSR failure message never exercised by a test** ([`cmd/openvox-ca/csr.go:113`](https://github.com/voxpupuli/openvox-ca/blob/2f0603336e268fef4fc3e9959eb72d1c44b5001a/cmd/openvox-ca/csr.go#L113))

  ```
  if createKey && !hadCert {
  					if nowHasKey, keyErr := myCA.HasCAKey(cmd.Context()); keyErr == nil && nowHasKey {
  						return fmt.Errorf("%w\n\nA CA key now exists with no certificate, so the server "+
  							"will refuse to start. Fix the cause and re-run 'openvox-ca csr' (without "+
  							"--create-key, which is no longer needed), then install the signed chain "+
  							"with 'openvox-ca import-ca-cert'", err)
  					}
  				}
  ```

  💡 **Recommendation:** Add a KeyProvider test double (or a narrow seam in newCSRCmd) that lets a test create the key successfully and then fail the signing step, so the 'A CA key now exists with no certificate...' message path gets exercised end to end.
  **Constraint:** Coverage completeness -- untested code path in core functionality

  This branch fires when `--create-key` mints a key and BuildCSR then fails on something other than ErrKeyProviderKeyNotFound (e.g. a Transit sign grant missing after the create grant succeeded) -- exactly the 'orphaned key' scenario the surrounding comment (lines 104-111) calls out by name and that `docs/storage-backends.md`'s new 'Removing an orphaned CA key' section documents as a real operational hazard. I searched csr_test.go for the literal string 'now exists' (present only in a comment at line 51) and for any test that makes HasCAKey return true after a BuildCSR failure; none exists. Every existing `createKey && !hadCert` test (`refuses without a hostname...`, `creates no key when the subject cannot be resolved`) fails before a key is ever created, so `nowHasKey` is false and the wrapped message is never reached. The message is specific, operator-facing guidance ('re-run without --create-key... then install the signed chain') that could silently rot (e.g. get the wrong verb, or stop firing after a refactor) without any test noticing.
  </details>

- 🧪 **Frontend/key-provider isolation only unit-tested in pieces, never at the wiring point** ([`cmd/openvox-ca/main.go:433`](https://github.com/voxpupuli/openvox-ca/blob/2f0603336e268fef4fc3e9959eb72d1c44b5001a/cmd/openvox-ca/main.go#L433))

  ```
  rt, err := resolveRuntime(ctx, cfg, roleMayReachCAKey(role))
  ```

  💡 **Recommendation:** Add a test that invokes the actual composition -- e.g. call newRootCmd's RunE (or refactor the storage/key-provider setup into a small helper callable with role="frontend") and assert the resulting KeyProvider is nil, rather than relying solely on the two constituent unit tests to imply the composition is correct.
  **Constraint:** Security test coverage -- privilege boundary transitions

  The two things this line composes are each well tested in isolation: `runtime_test.go` proves `resolveRuntime(ctx, cfg, false)` leaves `KeyProvider` nil even with OpenBao configured, and a separate DescribeTable proves `roleMayReachCAKey("frontend")` is false. But no test exercises this call site itself -- the actual security guarantee (the comment above it: a frontend process 'must never construct a provider of its own'). I grepped the full repository (`grep -rln "roleMayReachCAKey\|role.*frontend"`) and found no test, in this batch or elsewhere in the tree, that drives `newRootCmd()` with `PUPPET_CA_ROLE=frontend` and asserts the resulting runtime has a nil KeyProvider; `cmd/openvox-ca/main_test.go` (pre-existing, unchanged by this PR) contains no reference to `role`, `frontend`, `KeyProvider`, or `resolveRuntime` at all. A future edit that inverts the boolean, swaps the argument order, or hardcodes `true` here would pass every existing test in this batch while reopening the exact authenticated-session-to-the-key-backend risk the SC-3 comment describes.
  </details>

- 🧪 **New goroutine-race test runs without -race, so it cannot actually catch a race** ([`internal/ca/keyprovider_test.go:536`](https://github.com/voxpupuli/openvox-ca/blob/2f0603336e268fef4fc3e9959eb72d1c44b5001a/internal/ca/keyprovider_test.go#L536))

  ```
  		var wg sync.WaitGroup
  		for i := range racers {
  			wg.Add(1)
  			go func() {
  				defer wg.Done()
  				// Separate CA values sharing one backend, which is what two
  				// processes against a shared storage backend look like.
  				subject := ca.New(store, asCfg, "puppet.test")
  				subject.CAKeyConfig = ca.KeyConfig{Algo: ca.KeyAlgoECDSA, Size: 256}
  				key, err := subject.LoadOrCreateCAKey(ctx, true)
  ```

  💡 **Recommendation:** Run the unit suite (or at minimum this package) with `go test -race ./internal/ca/...` in CI, or add a dedicated `-race` target that magefile's Test.Unit / CI invokes, so the concurrency guarantees this test documents are actually enforced.
  **Constraint:** lang-go.md testing_conventions: "Race detection: go test -race (flag missing race detection in concurrent code tests)"

  "LoadOrCreateCAKey under concurrent creation" (added by this PR) spins up 8 goroutines that concurrently call `subject.LoadOrCreateCAKey(ctx, true)` against one shared `store`, exercising `storage.WithLock` and `c.mu` in `internal/ca/keyprovider.go`. This is the only genuinely concurrent unit test in the changeset, and its entire value is proving there is no data race between the racing goroutines. I confirmed with `grep -rn "\-race" --include="*.go" --include="*.yml" --include="Makefile" .` across the repo, and separately read `magefile.go`'s `Test.Unit` (the target that runs `go test -json -cover -coverprofile=coverage.out` over the discovered package list, lines 295-320) that neither passes `-race` anywhere in the project. Without `-race`, the assertions on final key equality and blob count can pass by coincidence even if `LoadOrCreateCAKey`'s locking has a genuine data race (e.g. a read of `c.CAKey` outside `c.mu` in a future edit), so the test provides much weaker protection than its name and comments claim.
  </details>

</details>

- 🧪 **Inconsistent nil-check on pem.Decode results across sibling specs** ([`cmd/openvox-ca/csr_test.go:153`](https://github.com/voxpupuli/openvox-ca/blob/2f0603336e268fef4fc3e9959eb72d1c44b5001a/cmd/openvox-ca/csr_test.go#L153))

  ```
  		block, _ := pem.Decode([]byte(out))
  		csr, err := x509.ParseCertificateRequest(block.Bytes)
  ```

  💡 **Recommendation:** Add `Expect(block).NotTo(BeNil())` before dereferencing `block.Bytes` at each of the listed call sites, matching the pattern already used elsewhere in the same file.
  **Constraint:** Assertion quality -- consistency of defensive checks within the same file

  Several specs in this file dereference `block.Bytes` immediately after `pem.Decode` without first asserting `Expect(block).NotTo(BeNil())` -- this one (lines 153-154, 'creates the key algorithm the configuration asks for'), and the same pattern at lines 169-171 ('persists the created key so a second run reuses it'), 234-235 ('reuses an established CA certificate's subject verbatim'), 267-268 ('reproduces the stored DER subject byte for byte'), and at importcacert_test.go:213-214. Sibling specs in the very same file -- lines 100-101, 192-193, 325-326, 341-342 -- do add the nil check before use. If `out` is ever not a valid PEM CSR (e.g. a future change accidentally writes a plain error string to stdout on a path these specs don't otherwise cover), the unchecked call sites panic on a nil pointer dereference instead of failing with a clear Gomega assertion message; Ginkgo will still catch and report the panic as a failure, so this is a diagnostics/consistency issue rather than a false-pass risk.
  </details>

</details>

---

bootc:
All eight findings from the review at `2f06033` are addressed, in `0b27bb3`. Four further council rounds ran on the fixes themselves; those are the four commits after it, and several of them corrected the fix rather than the original code. Rebased onto `main` and force-pushed — the branch was conflicting before.

### The eight findings

| Finding | What changed |
| --- | --- |
| Kubernetes `csr` Job omits `--create-key` | Not taken as recommended, see below |
| Doc names pre-rename `hasCAKey` | Corrected to `HasCAKey` |
| `writePublicFile` re-implements the atomic write | `storage.AtomicWriteFile` exported and shared; the cmd-side copy is now one call |
| `bootstrapCA` swallows the CA public key write failure | Now fails, **and** `seedSupportingState` backfills the blob on the next start |
| Post-key-creation `BuildCSR` failure never exercised | `newCSRCmd` takes its runtime resolver; two specs, mutation-checked |
| Frontend/key-provider isolation untested at the wiring point | `resolveRuntimeForRole` names the composition; two specs |
| Race test runs without `-race` | `mage test:unit` runs under `-race` with `CGO_ENABLED=1` |
| Inconsistent `pem.Decode` nil checks | Added at the five sites that dereferenced without one |

### The one I did differently

Adding `--create-key` to the Job's `args` would contradict the same document's Transit policy, which deliberately excludes `create` on `transit/keys`, and the Job runs under the Deployment's ServiceAccount. An operator copying it hits `permission denied`, and the obvious remedy permanently widens the running CA's credential to key creation. The example keeps `args: ["csr"]` — matching the recommended out-of-band provisioning path — and the comment now says `--create-key` needs a `create` grant for the duration of the Job alone.

### Worth a second look

Two of the fixes above are larger than the finding that prompted them, and one has repo-wide reach:

- **`-race` on `test:unit`** affects every branch's CI job, not just this one. No pre-existing races anywhere in the tree; wall clock goes from ~54s to ~78s locally. It also makes a C toolchain a prerequisite for that target, now documented in `CONTRIBUTING.md`.
- **The `ca_pub.pem` backfill in `seedSupportingState`** came out of the follow-up rounds: failing the bootstrap on its own only deferred the silent omission by one restart, since the next start loads a valid cert+key and nothing rewrites that blob. Both the failure and the heal have specs.

Also from the later rounds, if you are reading the diff rather than the commits: the three copies of the public key's marshal-encode-store sequence became one `savePubKeyPEM`, and an `errPubKeyMarshal` sentinel I added along the way was removed again once two reviewers established the branch consuming it was unreachable — and misclassified, since it would have dropped the `incompleteImportError` advice after the point of no return.

PR #162, opened 2026-07-29
Preserve upstream CRLs when re-signing this CA's own

bootc:
`signCRLLocked` replaced the entire stored CRL blob with a single block on every
re-sign. For a self-signed root that is correct and always has been. For an
intermediate it silently destroyed the ancestor CRLs the CA must publish for
agents to do full-chain revocation checking — Puppet's default
`certificate_revocation = chain` — at the first revocation or CRL refresh after
import. The chain was there right up until the CA was used.

Re-signing now preserves every block that is not this CA's own, in order, with
ours first. On a CA with nothing upstream the stored bytes are identical to
before, so an older replica re-signing mid-rollout is safe.

### Ownership is decided by signature

Which block is "ours" is settled by verifying the CRL's signature against this
CA's certificate. Two cheaper tests were tried and both are wrong here:

- **Issuer DN** is forgeable in exactly the deployment this exists for — a
  shared root can issue two sub-CAs with the same name.
- **Authority Key Identifier** is an optional extension. `openssl ca -gencrl`
  omits it, because the stock `openssl.cnf` leaves `crl_extensions` commented
  out, so an ancestor CRL routinely has none.

The AKI version shipped in the first commit and was wrong in both directions: at
import it could not recognise the operator's own AKI-less CRL, generated a fresh
empty one and prepended it — and since every reader takes block 0, the CA then
believed nothing was revoked, so revoked certificates authenticated again and
OCSP answered Good. On re-sign it dropped ancestors it could not positively
identify, which is the same data loss this branch was written to fix. A
signature check has neither failure mode and needs no optional extension.

### Behaviour changes worth review

- **`openvox-ca-ctl import --crl-chain` rejects a chain containing any
  unparseable CRL block**, where it previously skipped it with a warning. The
  blob is served verbatim to every agent and Puppet parses all of it, so a bad
  block is better caught at import than fleet-wide. The chain is also stored
  re-encoded from what was parsed, on both branches — the upstream-only path
  used to write the operator's bytes through untouched.
- **A failed read while preserving the chain now fails the re-sign.** All four
  callers read the same blob first under the same lock, so the "no stored CRL
  yet" case was unreachable and the only live trigger was a transient backend
  fault — which flattened the chain permanently and silently. Only
  `fs.ErrNotExist` is tolerated now, matching what the package already does for
  this blob three lines away.
- **`crl_update_failures_total` now covers all four update paths.** Three
  returned the error uncounted, so a replica tripping the fail-closed check
  without revoking anything logged every tick while the counter stayed flat and
  the mixin's alert could not fire.
- **The startup warning no longer suggests `openvox-ca-ctl reissue-crl`**, which
  fails closed on the very condition that triggers the warning.

### What this does not do

Imported ancestor CRLs are stored as supplied and never refreshed — this CA
cannot re-sign another CA's list — so they age in place and must be re-imported
before their own `nextUpdate` lapses. The CRL metrics describe block 0 only, so
that expiry is outside the mixin's alert coverage. Both are now recorded in
`metrics.md`, `README.md` and the migration guide rather than left implied.

Likewise, a replica running a build older than this one still flattens the blob,
and the loss is not detectable from the CA side: every CRL metric derives from
block 0, so a chain going 2 → 1 moves `crl_number` upward and looks healthy. The
migration guide says to complete a rolling upgrade before importing a chain.

Fetching and refreshing ancestor CRLs is item 4b and deliberately out of scope
here.

### Tests

The AKI-less fixture is hand-rolled ASN.1: the standard library always stamps an
AKI, which is itself the reason a CRL without one provably is not ours. Both
blocker fixes were confirmed load-bearing by reverting each in turn and watching
the corresponding spec fail.


trevor-vaughan-ai:

>
> Models used:
> - Capable tier
> - Standard tier

_Reviewed at commit [`bf4f750`](https://github.com/voxpupuli/openvox-ca/commit/bf4f750d82ea568ca2ddf95ba2e5a23fb968b5b1) (deep effort)._

**TL;DR:** 14 issues found, mostly missing tests for new CRL failure paths and doc drift; testing and docs personas request changes, three others approve.

**Findings:** 🔴 0 Critical, 🟠 3 High, 🟡 9 Medium, 🔵 2 Low

|---|---|---|
| 📚 Curator (code) | ❌ Changes | 2 MEDIUM, 1 LOW |
| 🧭 Guard (code) | ✅ Approve | 3 MEDIUM |
| ⚙️ Operator (code) | ✅ Approve | 3 MEDIUM |
| 🧪 Tester (code) | ❌ Changes | 3 HIGH, 1 MEDIUM, 1 LOW |

- 🧪 **Clean's revoke-failure-then-delete path has no test, despite the diff documenting a newly reachable way into it** ([`internal/ca/signing.go:507`](https://github.com/voxpupuli/openvox-ca/blob/bf4f750d82ea568ca2ddf95ba2e5a23fb968b5b1/internal/ca/signing.go#L507))

  ```
  slog.Warn("Clean: revoke failed, deleting the certificate anyway; it stays a valid "+
  ```

  💡 **Recommendation:** Add a test that seeds a stored CRL not signed by the CA (as the 'CRL cache loading' Describe block in crlchain_test.go already does for Init), issues and then Cleans a certificate, and asserts Clean still succeeds, the cert is removed from storage, and the cert's serial is NOT present in the CRL (i.e. it remains an unrevoked, still-valid credential) - locking down the exact newly-reachable behavior this diff's comment describes.
  **Constraint:** Untested code paths in core functionality (severity.md, Tester HIGH row) / lang-go.md testing_conventions

  The diff rewrites the warning fired when revokeLocked fails inside Clean, and its own comment (lines 499-506) states this is now a *newly reachable* condition: 'A foreign stored CRL now reaches here (readStoredCRL refuses to re-sign a list this CA did not issue), which is a newly reachable way into this state.' That claim ties directly to this PR's own loadCRLCache change, which now *warns and continues* instead of failing closed when block 0 is not this CA's own CRL (init.go ownsCRL check) - so a CA can start successfully with a foreign CRL cached, and the first Clean() on that replica will then hit revokeLocked failing, leaving the certificate deleted from storage while still unrevoked (a valid credential until expiry). I searched every *_test.go under internal/ca for a call exercising this branch: `grep -rn "\.Clean(" *_test.go` finds only ca_test.go:1326 (revoke succeeds), ca_test.go:1343 (CSR-only, no cert), and ca_test.go:1349 (ErrNotFound); cleanup_test.go's CleanMultiple tests (lines 258-297) only exercise the same successful-revoke and not-found outcomes. No test forces revokeLocked to fail (e.g. via a foreign stored CRL as loadCRLCache now tolerates) and then asserts that Clean still deletes the cert, still returns nil, and logs the new warning text.
  </details>

- 🧪 **Revoke's reliance on centralized crlUpdateFailures counting has no regression test** ([`internal/ca/revoke.go:82`](https://github.com/voxpupuli/openvox-ca/blob/bf4f750d82ea568ca2ddf95ba2e5a23fb968b5b1/internal/ca/revoke.go#L82))

  ```
  // 1. Load CRL. readStoredCRL counts its own failures now, so this path must
  	// not add a second increment for the same event.
  ```

  💡 **Recommendation:** Add a `Revoke` entry to the existing `DescribeTable` (or an equivalent standalone spec) that stores a CRL not signed by the CA, calls `myCA.Revoke(ctx, someSubject)`, asserts the error `errors.Is` `ca.ErrForeignStoredCRL`, and asserts `CRLUpdateFailures()` is exactly 1 - mirroring the reissue/refresh entries already present.
  **Constraint:** reviewer-protocol.md Engineering Discipline / severity.md HIGH - "missing regression test" for a diff-introduced behavior change

  This diff removes revoke.go's own `c.crlUpdateFailures.Add(1)` call after `readStoredCRL` because counting moved into `readStoredCRL` itself (see crl.go's new `readStoredCRL`/`parseStoredCRL` split). crlchain_test.go adds a `DescribeTable("counts a foreign stored CRL as a CRL-update failure on every re-sign path", ...)` specifically to guard this exact class of regression (double-counting or under-counting), but its `Entry` list only drives `reissue` and `refresh` (crlchain_test.go:340-345) - there is no `Entry` for `Revoke`. I grepped the whole `internal/ca` package (`crlchain_test.go`, `crl_test.go`, `cleanup_test.go`) for `ErrForeignStoredCRL` combined with a call to `Revoke`/`revokeSerialLocked`, and found none: `crl_test.go` and `cleanup_test.go` never reference `ErrForeignStoredCRL` at all, and crlchain_test.go's only uses of it are inside the reissue/refresh DescribeTable. The DescribeTable's own comment explains this centralization exists precisely because the metric silently went flat while the CA kept logging - the same blind spot now exists, untested, on the one call site (`revokeSerialLocked`) whose diff in this PR is exactly this de-duplication.
  </details>

- 🧪 **crlChainLocked's chain-decode failure branch is never exercised by a test** ([`internal/ca/crlchain.go:135`](https://github.com/voxpupuli/openvox-ca/blob/bf4f750d82ea568ca2ddf95ba2e5a23fb968b5b1/internal/ca/crlchain.go#L135))

  ```
  	stored, err := decodeCRLChain(existing)
  	if err != nil {
  		return nil, fmt.Errorf("decoding the stored CRL chain: %w", err)
  	}
  ```

  💡 **Recommendation:** Add a spec that writes `[ourCRL, corruptBlock]` to storage (as the existing "CRL cache loading: a corrupt ancestor block" spec does), then calls `myCA.ReissueCRL(ctx)` and asserts the error contains "decoding the stored CRL chain", that the stored blob is left untouched, and that `CRLUpdateFailures()` incremented - mirroring the GetCRL-failure spec already present for the sibling branch.
  **Constraint:** severity.md HIGH - "Untested code paths in core functionality" / "Missing edge case coverage for boundary-sensitive code"

  crlChainLocked is the function this PR adds to fix the chain-preservation bug, and it has two distinct failure branches on the read of the existing stored blob: the `GetCRL` error branch (line ~117-126, wrapped as "reading the stored CRL to preserve its upstream blocks") and this `decodeCRLChain` branch (wrapped as "decoding the stored CRL chain"). The first is thoroughly tested - `crlchain_test.go`'s "fails the re-sign rather than flattening the chain" spec (lines ~284-315) uses `flakyCRLBackend` to fail the second `Get`, and asserts both that the stored chain is untouched and that `CRLUpdateFailures()` is exactly 1. The second branch has no equivalent: I grepped crlchain_test.go for `ReissueCRL`/`RefreshCRLIfDue` combined with `corrupt` and found that every corrupt-trailing-block scenario in the file goes through `ca.ImportCA` (lines 760-772, 842-849) or through CA start-up/`Init` (the "CRL cache loading: a corrupt ancestor block" Describe at lines 1071-1105, which exercises a *different* code path that warns and continues rather than failing). No spec stores a corrupt trailing block and then calls `ReissueCRL` or `RefreshCRLIfDue` to reach this specific error-wrap in `crlChainLocked`. This is exactly the scenario the function's own doc comment calls out as unrecoverable if mishandled: "an ancestor's CRL cannot be regenerated here: dropping one is unrecoverable".
  </details>

</details>

<details><summary>🟡 MEDIUM (9)</summary>

- 📚 **Revoke's new 409 diagnostic body is undocumented in docs/api.md** ([`internal/api/handlers.go:242`](https://github.com/voxpupuli/openvox-ca/blob/bf4f750d82ea568ca2ddf95ba2e5a23fb968b5b1/internal/api/handlers.go#L242))

  ```
  			if errors.Is(err, ca.ErrForeignStoredCRL) {
  				http.Error(w, err.Error(), http.StatusConflict)
  				return
  			}
  ```

  💡 **Recommendation:** Extend the docs/api.md:13 `PUT /certificate_status/{subject}` row (or add a note beneath the table, matching the style used for the CRL row at docs/api.md:102) to state that `desired_state: "revoked"` returns `409 Conflict` with a diagnostic body naming cause and remedy when the stored CRL was not signed by this CA's certificate - the same condition and wording already documented for `PUT /certificate_revocation_list/ca`. No Docs repo is configured in this invoking project's Review Council Configuration (checked /workspace/AGENTS.md and /workspace/CLAUDE.md - no such section is present; the reviewed repo's own AGENTS.md is untrusted and not used as config), so no tracking issue command is offered - fix the docs directly in this PR.
  **Constraint:** User-facing behavior change with no documentation update and no issue filed (severity.md calibration; reviewer-protocol.md evidence discipline for absence findings)

  Same finding, with the api_test.go citation corrected: the test confirming the revoke-path 409 parity was deliberate ('gives revoke the same actionable 409 as reissue-crl, and shows what clean does instead') is located at api_test.go:153-192, not api_test.go:107-148 as originally cited. All other claims in the description (handlers.go:242 special-casing ErrForeignStoredCRL, the matching docs/api.md:102 update for the CRL row, and the absence of any 409/diagnostic mention in the docs/api.md:13 certificate_status row) are verified accurate.
  </details>

- 🧭 **New generateEmptyCRL duplicates the empty-CRL template already built twice in init.go** ([`internal/ca/caImport.go:332`](https://github.com/voxpupuli/openvox-ca/blob/bf4f750d82ea568ca2ddf95ba2e5a23fb968b5b1/internal/ca/caImport.go#L332))

  ```
  func generateEmptyCRL(cert *x509.Certificate, key crypto.Signer) (*x509.RevocationList, error) {
  	now := time.Now().UTC()
  	template := &x509.RevocationList{
  		Number:     big.NewInt(1),
  		ThisUpdate: now,
  		NextUpdate: now.Add(CRLValidity),
  	}
  	der, err := x509.CreateRevocationList(rand.Reader, template, cert, key)
  ```

  💡 **Recommendation:** Extract a package-level helper (e.g. `newEmptyCRLTemplate(cert, key, validity)` or reuse `generateEmptyCRL` itself with a validity parameter) and have bootstrapCA and seedSupportingState call it, since this diff is already introducing the canonical version of this logic in generateEmptyCRL.
  **Constraint:** Structural Coherence / DRY (zero-waste)

  This diff adds a third copy, in the same `ca` package, of the exact same empty-CRL construction already present twice in internal/ca/init.go: bootstrapCA (init.go lines 453-458, `crlTemplate := &x509.RevocationList{Number: big.NewInt(1), ThisUpdate: now, NextUpdate: now.Add(c.crlValidity())}` followed by `x509.CreateRevocationList(rand.Reader, crlTemplate, c.CACert, c.CAKey)`) and seedSupportingState (init.go lines 177-182, identical shape). All three build a `Number: 1` revocation-list template with the same three fields and call CreateRevocationList. The new copy in caImport.go additionally diverges by validity source (package-level `CRLValidity` constant vs. `c.crlValidity()` used by the other two, though that divergence itself predates this diff and was carried over from the code being replaced). Three call sites building identical security-relevant CRL-template logic is exactly the maintenance-waste case the DRY/zero-waste criterion targets: a future change to the empty-CRL shape (e.g. adding an extension, or fixing the CRLValidity-vs-crlValidity() inconsistency) has three places to update, and this diff was the opportunity to add a shared helper instead of a third copy.
  </details>

- 🧪 **loadCRLCache's undecodable-chain fallback branch is untested** ([`internal/ca/init.go:550`](https://github.com/voxpupuli/openvox-ca/blob/bf4f750d82ea568ca2ddf95ba2e5a23fb968b5b1/internal/ca/init.go#L550))

  ```
  slog.Warn("Could not decode the stored CRL chain to look for this CA's own block; "+
  ```

  💡 **Recommendation:** Add a case to the 'CRL cache loading' Describe block that stores a foreign block 0 followed by an undecodable trailing block (mirroring the corrupt-chain fixture already used in the CRL-import tests, e.g. crlchain_test.go:765), calls Init, and asserts it succeeds while logging the 'Could not decode the stored CRL chain' warning and still answers from the foreign block 0.
  **Constraint:** Missing edge case coverage for boundary-sensitive code (severity.md, Tester)

  loadCRLCache's new ownership-fallback logic (init.go lines 527-553) has three sub-branches after `!c.ownsCRL(crl)`: (1) decodeCRLChain succeeds and finds this CA's own block later in the chain, (2) decodeCRLChain succeeds but finds none of ours, (3) decodeCRLChain fails outright. crlchain_test.go's 'CRL cache loading' Describe block (lines 864-948) covers branch 1 ('answers revocation from our own block when a foreign one leads') and branch 2 ('starts with a warning rather than failing when block 0 is not ours', where storage holds only the single foreign upsCRL block). I searched for coverage of branch 3 with `grep -n "Could not decode\|continuing with block 0" *_test.go` across internal/ca and found no match outside init.go itself - no test stores a PEM blob whose block 0 is foreign and whose trailing bytes are corrupt/undecodable to force decodeCRLChain to error inside this fallback search.
  </details>

- ⚙️ **Every CRL re-sign reads the stored blob from the backend twice** ([`internal/ca/crlchain.go:117`](https://github.com/voxpupuli/openvox-ca/blob/bf4f750d82ea568ca2ddf95ba2e5a23fb968b5b1/internal/ca/crlchain.go#L117))

  ```
  func (c *CA) crlChainLocked(ctx context.Context, ourCRL *x509.RevocationList) ([]byte, error) {
  	chain := []*x509.RevocationList{ourCRL}

  	existing, err := c.Storage.GetCRL(ctx)
  ```

  💡 **Recommendation:** Have readStoredCRL/parseStoredCRL retain and return the raw blob (or the full decoded chain) alongside block 0, and pass it into crlChainLocked instead of re-fetching, eliminating the second GetCRL call and halving lock hold time on HA backends.
  **Constraint:** Operator efficiency criterion: redundant file reads/API calls that could be cached or combined

  crlChainLocked is only ever reached from signCRLLocked (internal/ca/crl.go:74), which is itself only called after the caller (reissueCRLLocked at crl.go:124, RefreshCRLIfDue at crl.go:208, or revokeSerialLocked at revoke.go:84) has already fetched and parsed the same stored CRL via readStoredCRL -> parseStoredCRL, which calls c.Storage.GetCRL(ctx) at crl.go:162. All of this happens inside a single critical section held under both the cluster-wide lockNameCRL distributed lock and c.mu (see crl.go:110-117 and revoke.go:42-45), so no writer can have changed the blob between the two reads -- the second GetCRL call in crlChainLocked returns byte-identical data to the first. parseStoredCRL discards everything except block 0 (crl.go:161-191), so crlChainLocked has no way to reuse the first read's bytes and issues its own. internal/storage/storage.go:473-477 shows GetCRL has no in-process cache -- it calls straight through to s.backend.Get -- and docs/storage-backends.md documents four HA backends (etcd, redis, postgres, mysql) where that Get is a network round trip. Every revoke, reissue, and refresh operation therefore does two sequential network round trips to fetch the same bytes while holding a cluster-wide lock that blocks every other replica's CRL mutation for the duration.
  </details>

- ⚙️ **Silent ancestor-CRL loss during a rolling upgrade has no metric or alert, only an Info log** ([`internal/ca/crlchain.go:153`](https://github.com/voxpupuli/openvox-ca/blob/bf4f750d82ea568ca2ddf95ba2e5a23fb968b5b1/internal/ca/crlchain.go#L153))

  ```
  if before, after := len(stored), len(chain); after != before {
  		slog.Info("Stored CRL chain length changed while re-signing",
  			"blocks_read", before, "blocks_written", after)
  	}
  ```

  💡 **Recommendation:** Either export a gauge for the current stored CRL chain length (so a Grafana panel or alert can catch an unexpected drop) or, at minimum, raise this log to Warn and document the rolling-upgrade risk and the log line to watch for in docs/metrics.md alongside the existing ancestor-CRL blind-spot callout.
  **Constraint:** Operator upgrade/migration-path and runtime-observability criteria: version-skew detection and meaningful signals for irreversible state changes

  The comment directly above this block (crlchain.go:145-152) states the failure mode explicitly: a replica still running a pre-fix build during a rolling upgrade re-signs the CRL, drops every upstream block (because it lacks this PR's preservation logic), and the loss is permanent since this CA cannot regenerate an ancestor's CRL. The comment says detection is 'the operator's, via the log below and the chain length they expect' -- but that log is emitted at Info level (not Warn, unlike every other CRL-chain anomaly this PR adds: crlchain.go:197, 224, 232), there is no corresponding metric (grep of internal/metrics/collector.go shows counters for crl_update_failures and crl_number/crl_revoked_certificates, but nothing tracking chain length or block count), and no alert in mixin/alerts.libsonnet references it (the file's only CRL alert is PuppetCACRLUpdateFailing on puppetca_crl_update_failures_total, which this event does not increment). docs/metrics.md already documents a related but distinct ancestor-CRL blind spot in detail (the 'shipped CRL expiry alerts do not cover ancestor CRLs' callout at metrics.md:98-105) but says nothing about the rolling-upgrade chain-shrinkage risk the code comment describes, so an operator has no way to learn what signal to watch for during a fleet upgrade short of reading this Go source comment.
  </details>

- 📚 **Storage backends guide's filesystem layout still implies one CRL per file** ([`docs/storage-backends.md:45`](https://github.com/voxpupuli/openvox-ca/blob/bf4f750d82ea568ca2ddf95ba2e5a23fb968b5b1/docs/storage-backends.md#L45))

  ```
  ├── ca_crl.pem              Certificate Revocation List
  ```

  💡 **Recommendation:** Add a sentence to docs/storage-backends.md's filesystem layout description (and/or the SQL/etcd/Redis sections, which also list `crl`/`ca_crl` as a single logical key) noting that the CRL blob may contain a concatenated chain (this CA's own CRL first, ancestor CRLs after) once `--crl-chain` has been used to import one, mirroring the wording already added to docs/development/storage-internals.md.
  **Constraint:** Documentation gap: every other file in this changeset that describes `ca_crl.pem` / the `crl` logical key was updated to explain that it can now hold a concatenated chain of CRLs (this CA's own plus preserved ancestor blocks) after `--crl-chain` import - see the updated `crl` row in docs/development/storage-internals.md and the new ancestor-preservation section in docs/migrating-from-puppet-server.md, which itself links to this file for backend-specific detail.

  docs/migrating-from-puppet-server.md (changed in this PR) directs readers to docs/storage-backends.md for backend-specific CRL storage detail ("SQLite is in that list: it keeps the CRL in its database file... See [storage backends](storage-backends.md)."), and docs/development/storage-internals.md's `crl` row was rewritten in this same PR to explain the multi-block chain format. docs/storage-backends.md's filesystem-layout table, however, still labels `ca_crl.pem` as simply "Certificate Revocation List" with no mention that it can hold several concatenated CRL blocks post-import - an operator reading only this guide (the one the migration doc sends them to) would not learn that the file's format changed. I searched docs/storage-backends.md for "chain", "concatenated", and "ancestor" and found no matches.
  </details>

- 🧭 **New operator-warning paragraphs skip the file's own blockquote-callout convention** ([`docs/migrating-from-puppet-server.md:157`](https://github.com/voxpupuli/openvox-ca/blob/bf4f750d82ea568ca2ddf95ba2e5a23fb968b5b1/docs/migrating-from-puppet-server.md#L157))

  ```
  **Re-import rewrites the CA key, so two custody modes cannot use it.** `import`
  ```

  💡 **Recommendation:** Render the five new caveat paragraphs (lines 99, 107, 112, 157, 167) as `> **Label:**` blockquotes to match the convention already used elsewhere in this file and in the sibling docs edited by this same change.

  This diff adds FIVE new bold-lead caveat paragraphs to migrating-from-puppet-server.md (lines 99, 107, 112, 157, 167), not four as originally stated -- line 112 ('`import` writes to a local filesystem directory only.') was omitted from the original finding but is also new (confirmed via `git diff main...HEAD`) and equally breaks the file's `> **Label:**` blockquote convention used at line 260 and in sibling docs (api.md:76, metrics.md:98-105). The recommendation should render all five paragraphs (99, 107, 112, 157, 167) as blockquotes, not just four.
  </details>

- 🧭 **Step 3 balloons to 114 lines with an unmarked nested backend-refresh procedure** ([`docs/migrating-from-puppet-server.md:97`](https://github.com/voxpupuli/openvox-ca/blob/bf4f750d82ea568ca2ddf95ba2e5a23fb968b5b1/docs/migrating-from-puppet-server.md#L97))

  ```
  Some limits worth knowing before you rely on re-import as a refresh mechanism.
  ```

  💡 **Recommendation:** Extract the ancestor-CRL-refresh material (from "Some limits worth knowing..." through the closing bash script) into its own subsection or a separate reference doc, and leave Step 3 with a short pointer to it, so the step-by-step pattern the rest of the guide follows is preserved.

  The guide's established pattern is one focused action per `## Step N` heading with a short explanation: Step 1 is 9 lines, Step 2 is 24, Step 4 is 16, Step 5 is 23, Step 6 is 22, Step 7 (the next-largest) is 43. This diff inserts roughly 100 new lines into `## Step 3: Import the CA` (line 65) before its original closing sentence (now line 177), bringing Step 3 to 114 lines - over twice the length of any other step. The inserted material is not incremental elaboration on the `import` command; from line 97 onward it is a self-contained, differently-scoped procedure for refreshing ancestor CRLs on non-filesystem backends, complete with its own numbered caveats and its own three-command bash workaround script (lines 139-150) built around a `migrate`/`import`/`migrate --force` round trip. It is embedded directly in the middle of the linear migration walkthrough with no sub-heading, no cross-reference from the top of the step, and no acknowledgment that it is optional/advanced material - a reader doing a first-time migration must scroll past the entire backend-refresh workaround to reach the step's actual conclusion.
  </details>

- ⚙️ **Migration round-trip example ignores its own mktemp -d and backup guidance** ([`docs/migrating-from-puppet-server.md:140`](https://github.com/voxpupuli/openvox-ca/blob/bf4f750d82ea568ca2ddf95ba2e5a23fb968b5b1/docs/migrating-from-puppet-server.md#L140))

  ```
  # scratch.yaml: storage_backend: filesystem, cadir: /tmp/scratch
  openvox-ca-ctl migrate --source-config live.yaml --dest-config scratch.yaml

  # --cert-bundle and --private-key are the copies the first leg just wrote there.
  openvox-ca-ctl import --cadir /tmp/scratch \
    --cert-bundle /tmp/scratch/ca_crt.pem \
    --private-key /tmp/scratch/private/ca_key.pem \
    --crl-chain refreshed-chain.pem

  openvox-ca-ctl migrate --source-config scratch.yaml --dest-config live.yaml --force
  ```

  💡 **Recommendation:** Update the example to assign `SCRATCH=$(mktemp -d)` and use `$SCRATCH` in place of `/tmp/scratch` throughout, add an explicit backup command for the live destination before the final `--force` migrate (mirroring the Step 1 `cp -a` pattern), and append `rm -rf "$SCRATCH"` after the import step to remove the plaintext CA key copy as the prose requires.
  **Constraint:** Operational documentation for a destructive procedure must reflect the safety steps its own prose mandates (Backup and Recovery / Operational Documentation)

  The paragraph immediately above this code block (lines 132-137) gives three explicit safety instructions for this exact round trip: "Use a fresh, empty scratch directory every time - mktemp -d", "Back up the destination first", and "Remove the directory afterwards; it holds a plaintext copy of the CA key." The copy-pasteable example that follows does none of these: it hardcodes `/tmp/scratch` instead of using `mktemp -d`, takes no backup of `live.yaml`'s destination before the final `--force` migrate, and never removes `/tmp/scratch` afterward. This is not cosmetic - the same document independently confirms (and I verified against internal/ca/caImport.go and cmd/openvox-ca-ctl/migrate.go) that `migrate` resolves its destination purely from the `--dest-config` YAML file while `import` writes purely to the `--cadir` flag, with zero cross-validation between them, and that `migrate` copies but never deletes. A reused fixed scratch directory therefore risks reintroducing stale data (including certificates already cleaned, which "reappear with no inventory row and are then invisible to the expiry cleanup" per this same doc) into the live backend on a later run, and an un-backed-up final `--force` migrate has no documented recovery path if it writes bad data into `live.yaml`'s destination. The document elsewhere (Step 1) demonstrates the exact pattern for a proper backup command (`cp -a "$PUPPET_SSL" "$BACKUP_DIR"`), so the omission here is inconsistent with the document's own standard.
  </details>

</details>

- 🧪 **warnAboutAncestors' duplicate-ancestor dedup-reset only tested with exactly two copies** ([`internal/ca/crlchain.go:230`](https://github.com/voxpupuli/openvox-ca/blob/bf4f750d82ea568ca2ddf95ba2e5a23fb968b5b1/internal/ca/crlchain.go#L230))

  ```
  	for _, crl := range others {
  		if n := seen[string(crl.RawIssuer)]; n > 1 {
  			slog.Warn("Imported chain carries more than one CRL for the same ancestor; "+
  				"all of them are published as supplied",
  				"issuer", crl.Issuer.String(), "copies", n)
  			seen[string(crl.RawIssuer)] = 0
  		}
  	}
  ```

  💡 **Recommendation:** Extend the duplicate-ancestor spec (or add a new one) with three copies of the same ancestor CRL and assert the warning substring appears exactly once in the captured log buffer.
  **Constraint:** severity.md LOW - minor edge-case coverage gap on non-critical (observability) behavior

  The `seen[...] = 0` reset after warning exists so a group of 3+ duplicate ancestor CRLs produces exactly one warning instead of one per pair past the first. crlchain_test.go's "warns when the chain carries two CRLs for the same ancestor" (line 694) only constructs a group of exactly two duplicates, which cannot distinguish "warn once" from "warn once per duplicate beyond the first" - a bug that silently omitted the reset (warning on every excess duplicate) would still pass this spec with 2 copies but would over-warn with 3.
  </details>

- 📚 **New `--crl-chain` comment breaks the alignment/wrap style of the block it's appended to** ([`docs/operator-cli.md:93`](https://github.com/voxpupuli/openvox-ca/blob/bf4f750d82ea568ca2ddf95ba2e5a23fb968b5b1/docs/operator-cli.md#L93))

  ```
                                 # imported -- pass --crl-chain to replace it.
  # --crl-chain may hold several concatenated CRLs in any order. This CA's own is
  # identified by signature and moved to the front; the rest are kept and served
  # so agents can do full-chain revocation checking. Every X509 CRL block must parse; other PEM block types are ignored and not stored.
  ```

  💡 **Recommendation:** Either continue the aligned-indentation style used in lines 90-93 for the new sentences, or clearly separate the two comment blocks (e.g. a blank line) and wrap the long final line to match the ~78-character width used elsewhere in this example.
  **Constraint:** Inline documentation format consistency within the same fenced example

  The final new comment line (docs/operator-cli.md:96, 'so agents can do full-chain revocation checking. Every X509 CRL block must parse; other PEM block types are ignored and not stored.') is 133 characters long, not ~180 as originally stated. It still clearly exceeds the ~76-79 character wrap width used by every other line in the surrounding example, and lines 94-96 still start at column 0 while lines 90-93 are indented to column 32 to align under the inline comment they continue -- the style-break and readability concern is otherwise accurately described.
  </details>

</details>

---

bootc:
Thanks — all 14 addressed, plus what three further council rounds turned up. Pushed as `fee058e`, `dcb1745`, `f7b20b2`, `c49feca`; CI green. Every fix is verified by reverting it and watching a named spec fail.

**The three HIGHs, as filed.** All three were missing tests for paths this branch made reachable:

- **`Clean` deletes the certificate even when revoking it fails.** Pinned, including that the stored CRL is untouched — which is what makes the deleted certificate still a valid credential until it expires.
- **`Revoke` was the one re-sign path missing from the `crlUpdateFailures` table** — and the one whose own increment this branch removed. Re-adding it now fails the new entry.
- **`crlChainLocked`'s chain-decode failure**, plus `loadCRLCache`'s undecodable-chain fallback, which also records the consequence rather than implying it: revocation answers come from the foreign block, so a certificate this CA revoked reads as not revoked.

Of the MEDIUMs, the substantive one was the **double read of the stored blob**. One read serves both halves of the re-sign now; a spec counts backend CRL reads. That was not just a wasted round trip inside the cluster lock — the two reads could disagree, with the number and entries from one and the ancestors from the other.

**What the follow-up rounds found, which the original pass did not.** These are the ones worth your attention:

**The newest-of-ours rule was enforced on the import path and bypassed everywhere else.** Block 0 being *ours* is not block 0 being our *newest*, and a blob holding a stale export ahead of the current one passes every ownership check. That shape is what an upgrade produces: the released build's import validated block 0 and wrote the operator's bundle verbatim, so `--crl-chain stale.pem current.pem root.pem` is a stored blob this code has to read. On it:

- `loadCRLCache` cached the stale block, because its search ran only when block 0 was *foreign* — so a certificate this CA had revoked authenticated, at every tier;
- `parseStoredCRL` handed the stale block to every re-sign, which advanced from its number (regressing a sequence the docs publish as monotonic), carried its entry list forward, and then dropped every block of ours including the newer one. Those revocations were gone for good.

`selectOwnCRL` is the single place that answers "which block is ours" now, and both readers use it. The same rule also spans the bundle and storage — a stale copy of ours in a supplied bundle against a newer one in storage used to be written back, deleting the only current copy.

**A numbered CRL now beats a numberless one.** `openssl ca -gencrl` under the stock `openssl.cnf` emits a V1 list, which cannot carry `cRLNumber`, so a numberless CRL of ours is routine — and whichever block lands at position 0 is what the next re-sign advances from. Advancing from a numberless one restarts the sequence at 1, so an RFC 5280 client may keep serving the copy it has and never see a later revocation.

**`clean` is not symmetric with `revoke`, and now says so.** In the `409` state the `PUT` refuses and the `DELETE` succeeds, deleting a certificate that stays valid and is no longer in storage to clean again. Documented in `api.md` and beside the `openvox-ca-ctl clean` example, and the failure is counted now — including an inventory read failure, which on a blob backend covers an HMAC verification failure.

**Two process notes, since they affected what you can trust in the history.** The round-2 commit message described two changes that were *not in the commit* — a batch edit aborted partway and I did not check the diff against the message. All three reviewers caught it independently; both changes are now in, and that message no longer claims them. Separately, two mutations in the third round looked verified and were not: one broke the build and one never applied, and the grep I was using did not distinguish "no failures" from "did not compile". Both are confirmed now.

Deliberately not done: a gauge for the stored chain's length. It cannot see the case that actually loses ancestors — an old replica writes one block, and the next re-sign reads one and writes one — so `docs/metrics.md` documents the blind spot and the upgrade order instead, with the `WARN` to watch for.

PR #163, opened 2026-07-29
Record the current authorisation outcomes as an oracle

bootc:
The intermediate-CA work changes who counts as an admin, in three places: scoping
identity claims to the issuing CA, gating renewal on our own certificates, and
moving `certificate_status` to admin-only. The design calls the first of those
"the change that can silently widen who is an admin", and there was no artefact
recording what the answers are *now*.

This is that artefact, committed before any of those changes so it says what the
behaviour was rather than what someone later believed it should be. It is a test
plus documentation; no production code moves.

### Dependencies and merge order

This branch is based directly on `main` and depends on nothing else. It should
merge before #164 (renewal issuer gates) and #166 (client trust domains), both of
which build on it — each of those changes moves cells recorded here, and the
point of the oracle is to be in place first.

### What is recorded

Eleven client classes against twelve rows covering eleven of the nineteen
protocol routes, driven through the real mux:

| | |
|---|---|
| `none` | no client certificate |
| `own-ca-plain` | ordinary agent certificate from this CA |
| `own-ca-allowlisted` | CN in the admin allow list |
| `own-ca-pp-cli-auth` | carries `pp_cli_auth` with value `true` |
| `own-ca-pp-cli-auth-false` | carries `pp_cli_auth` with a value that is *not* `true` |
| `own-ca-admin-both` | admin by allow list *and* extension — how OpenVox Server ships |
| `own-ca-expired` | issued by this CA, past `notAfter` |
| `own-ca-issued` | genuinely issued through the CA, so recorded in the inventory |
| `own-ca-revoked` | issued by this CA, then genuinely revoked into the CRL |
| `own-ca-server-eku` | chain-valid but carrying only `serverAuth` EKU |
| `foreign-ca` | an unrelated CA, using a CN this CA's namespace also uses |

Every cell is asserted every run. A change that intends to move one updates the
recorded outcome and says why in `changedBy`, in the same commit; a change that
moves one without doing so fails here.

### How a moved cell is made to declare itself

Each row carries two digests. `fingerprint` covers the endpoint, the recorded
outcomes *and* `changedBy`, so no cell moves without a second deliberate edit
that is visible in the diff, and re-attributing a later change to an earlier one
trips it too. `baseline` covers the endpoint and outcomes as first committed and
is never refreshed, so "has this row ever moved" can still be asked after a row
has been attributed.

Neither is tamper-proof — no literal in a repository can be. What the pair buys
is that moving a cell silently requires replacing two opaque hex literals that no
honest change touches, and the file says so in those terms rather than claiming
an enforcement it does not have.

### What the digests do not cover

The digests bind what a row *records*, not the certificate that produced the
recording. That gap is wide: the eleven classes collapse into only four distinct
outcome columns, so ten of them are indistinguishable from a sibling by outcome
alone. Handing `own-ca-pp-cli-auth-false` an ordinary CA-issued certificate
reproduces its column exactly on every row — so relaxing `isAdmin` to a presence
test could have landed with every digest, every pinned list and every `changedBy`
byte-identical. The same swap hides an `ExtKeyUsageAny` relaxation behind
`own-ca-expired`'s column. Both bypasses were *cheaper* than the honest path,
which had the incentive gradient backwards.

A fixture-property spec now asserts what each class is, against the parsed
certificate: CN, allow-list membership, `pp_cli_auth` presence and value, EKU,
validity window, issuer, CRL membership, and presence in the CA inventory by
serial. Both attacks were carried out to confirm each now fails naming exactly
what was done.

### Coverage the omissions argument did not actually cover

`POST /ocsp` had no assertion anywhere in the repository. It is the only
`lookupTier` arm that grants a tier without inspecting the method, so the reason
the other omissions are safe — they fall through to `tierAdminOnly`, which three
rows pin — did not apply to it. Deleting the arm now fails a row in both the bare
and `/puppet-ca/v1`-prefixed forms.

### Documentation

`docs/api.md` gained the `clientAuth` EKU requirement, which the tier table
omitted, and had two security claims corrected. Revoking an agent's certificate
does not cut off its access to the CA API: the public tier examines no
certificate at all, so a revoked agent keeps the CA certificate, the CRL,
`/expirations`, OCSP and CSR submission. And neither revocation form prevents
re-enrolment under the same subject, because `SaveRequest` evicts a revoked
certificate for that subject before saving — so the containment levers are
autosign policy, network blocking and a forced CRL re-sign on every replica,
not the choice between revoke and clean.

`AGENTS.md` records the obligation on anyone editing the table, including the two
cases that are easy to get wrong: a new row needs both digests (write
`baseline: ""` and the suite prints a paste-ready pair), and adding a client class
re-digests every row.

### Notes from review

Five review rounds; each of the first four introduced at least one comment or
document claim the code had already outgrown, which the next round caught. That
pattern is itself the argument for the file's habit of stating what is *not*
enforced alongside what is.

Two recommendations were not taken as given. `Ordered` is not gratuitous —
`BeforeAll` requires it — so the fix was `Ordered, ContinueOnFailure` plus one
`Entry` per route rather than dropping `Ordered`. And the `baseline` gate was
left as a convention backed by diff visibility rather than made "mechanical",
because no in-repo literal can be made unforgeable and claiming otherwise was the
original defect.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -131,10 +131,11 @@ Conventions:
   `baseline:` line, which is what makes one in a diff worth stopping on.
 
   Two cases the above does not cover. A *new* row needs **both** digests, not
-  just one: write `baseline: ""` and run the suite, which prints a paste-ready
-  `fingerprint`/`baseline` pair. Leave `changedBy` empty — a new row has not
-  moved, and attributing it from birth permanently retires its baseline check.
-  "Do not edit `baseline`" applies from its second commit onwards.
+  just one: add its name to `expectedRoutes`, write `baseline: ""`, and run the
+  suite, which prints a paste-ready `fingerprint`/`baseline` pair. Leave
+  `changedBy` empty — a new row has not moved, and attributing it from birth
+  permanently retires its baseline check. "Do not edit `baseline`" applies from
+  its second commit onwards.
 
   Adding or removing a *client class* re-digests every row, so that commit sets
   `changedBy` on all of them naming the class, even though no authorisation
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -119,6 +119,13 @@ Conventions:
   conditions; sleeps make the suite flaky on loaded CI runners.
 - Keep negative and edge cases first-class: every security-relevant branch
   (rejection paths, tamper detection, auth denial) needs an explicit `It`.
+- `internal/api/authbaseline_test.go` is the recorded authorisation baseline:
+  every client class against every covered route, as the middleware behaves
+  today. A change that deliberately alters who may reach an endpoint updates the
+  affected row, sets `changedBy`, and refreshes that row's `fingerprint`, all in
+  the same commit — the suite fails otherwise, which is the point. Update
+  `docs/api.md#authorization-tiers` in step, since it publishes the same matrix
+  to operators.
 
 ### Integration suites (build-tagged)
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -141,6 +141,9 @@ Conventions:
   `changedBy` on all of them naming the class, even though no authorisation
   behaviour changed. Note what that costs: an attributed row's `baseline` is
   never compared again, so a class change retires the baseline layer repo-wide.
+  Such a commit also updates `expectedClientClasses`, gives the class a property
+  function in the fixture-property spec, and records its outcome on every row —
+  the suite enforces all three, but only after the digests stop it first.
 
   The suite cannot judge whether an attribution is *accurate*, nor whether a
   fixture still means what its class name says; reviewers should.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -123,13 +123,24 @@ Conventions:
   every client class against every covered route, as the middleware behaves
   today. A change that deliberately alters who may reach an endpoint updates the
   affected row, records the responsible change in `changedBy`, and refreshes
-  that row's `fingerprint` — all in the same commit. Leaving `changedBy` empty
-  fails permanently, because each row also carries a `baseline` digest of its
-  originally committed outcomes that is never updated. Do not edit `baseline`.
-  The suite cannot judge whether the attribution is *accurate*; reviewers
-  should.
-  Update `docs/api.md#authorization-tiers` in step, since it publishes the same
-  matrix to operators.
+  that row's `fingerprint` — all in the same commit, in that order (the suite
+  withholds the digest you need while `changedBy` is empty).
+
+  Each row also carries a `baseline` digest of its originally committed
+  outcomes. Do not edit it: no legitimate change to an existing row touches a
+  `baseline:` line, which is what makes one in a diff worth stopping on.
+
+  Two cases the above does not cover. A *new* row needs a `baseline` of its own
+  first outcomes, written once with `changedBy` left empty — "do not edit
+  `baseline`" applies from its second commit onwards. Adding or removing a
+  *client class* re-digests every row, so that commit sets `changedBy` on all of
+  them naming the class, even though no authorisation behaviour changed.
+
+  The suite cannot judge whether an attribution is *accurate*, nor whether a
+  fixture still means what its class name says; reviewers should.
+  `docs/api.md#authorization-tiers` publishes the tier assignment to operators,
+  so update it when a change moves a route between tiers — it is a tier table,
+  not this matrix, and most cells here have no counterpart there.
 
 ### Integration suites (build-tagged)
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -122,10 +122,14 @@ Conventions:
 - `internal/api/authbaseline_test.go` is the recorded authorisation baseline:
   every client class against every covered route, as the middleware behaves
   today. A change that deliberately alters who may reach an endpoint updates the
-  affected row, sets `changedBy`, and refreshes that row's `fingerprint`, all in
-  the same commit — the suite fails otherwise, which is the point. Update
-  `docs/api.md#authorization-tiers` in step, since it publishes the same matrix
-  to operators.
+  affected row, records the responsible change in `changedBy`, and refreshes
+  that row's `fingerprint` — all in the same commit. Leaving `changedBy` empty
+  fails permanently, because each row also carries a `baseline` digest of its
+  originally committed outcomes that is never updated. Do not edit `baseline`.
+  The suite cannot judge whether the attribution is *accurate*; reviewers
+  should.
+  Update `docs/api.md#authorization-tiers` in step, since it publishes the same
+  matrix to operators.
 
 ### Integration suites (build-tagged)
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -130,11 +130,16 @@ Conventions:
   outcomes. Do not edit it: no legitimate change to an existing row touches a
   `baseline:` line, which is what makes one in a diff worth stopping on.
 
-  Two cases the above does not cover. A *new* row needs a `baseline` of its own
-  first outcomes, written once with `changedBy` left empty — "do not edit
-  `baseline`" applies from its second commit onwards. Adding or removing a
-  *client class* re-digests every row, so that commit sets `changedBy` on all of
-  them naming the class, even though no authorisation behaviour changed.
+  Two cases the above does not cover. A *new* row needs **both** digests, not
+  just one: write `baseline: ""` and run the suite, which prints a paste-ready
+  `fingerprint`/`baseline` pair. Leave `changedBy` empty — a new row has not
+  moved, and attributing it from birth permanently retires its baseline check.
+  "Do not edit `baseline`" applies from its second commit onwards.
+
+  Adding or removing a *client class* re-digests every row, so that commit sets
+  `changedBy` on all of them naming the class, even though no authorisation
+  behaviour changed. Note what that costs: an attributed row's `baseline` is
+  never compared again, so a class change retires the baseline layer repo-wide.
 
   The suite cannot judge whether an attribution is *accurate*, nor whether a
   fixture still means what its class name says; reviewers should.
--- a/docs/api.md
+++ b/docs/api.md
@@ -179,7 +179,15 @@ When mTLS is enabled (both `--tls-cert` and `--tls-key` set), each endpoint requ
 | **Self or admin** | Cert CN matches path subject, OR cert is admin | `GET /certificate_request/{subject}` |
 | **Admin** | Cert is admin (see below) | `PUT /certificate_status/{subject}`, `DELETE /certificate_status/{subject}`, `DELETE /certificate_request/{subject}`, `GET /certificate_statuses/*`, `POST /sign`, `POST /sign/all`, `POST /generate/{subject}`, `PUT /clean`, `PUT /certificate_revocation_list/ca`, `PUT /certificate/{subject}` |
 
-In plain HTTP mode (no TLS), all endpoints are accessible without authentication.
+Above the public tier, a presented certificate must also be **currently valid and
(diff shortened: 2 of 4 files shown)


trevor-vaughan-ai:
## 🟢 Review Council: APPROVE

>
> Models used:
> - claude-sonnet-5

_Reviewed at commit [`7ecfbc4`](https://github.com/voxpupuli/openvox-ca/commit/7ecfbc44ae76651e58a1d73f58ad9a6d43d22c13) (deep effort)._

**TL;DR:** PR docs approved with 4 minor issues: overstated cert revocation guarantee, an undocumented CRL per-replica caveat, a spelling slip, and a missing test-convention carve-out.

**Findings:** 🔴 0 Critical, 🟠 0 High, 🟡 2 Medium, 🔵 2 Low

|---|---|---|
| 🛡️ Adversary (code) | ✅ Approve | 1 MEDIUM |
| 📚 Curator (code) | ✅ Approve | 1 LOW |
| 🧪 Tester (code) | ✅ Approve | 1 LOW |

- 🛡️ **DELETE /certificate_status doc overclaims revocation guarantee that Clean() does not enforce** ([`docs/api.md:205`](https://github.com/voxpupuli/openvox-ca/blob/7ecfbc44ae76651e58a1d73f58ad9a6d43d22c13/docs/api.md#L205))

  ```
  - `DELETE /certificate_status/{subject}` revokes *and* deletes it immediately,
    on shared storage, so it does not depend on any replica's cache.
  ```

  💡 **Recommendation:** Either have Clean() return (or aggregate) the revoke error so handleDeleteStatus can surface a non-204 response when revocation fails, or soften the docs claim to state that deletion is unconditional but revocation is best-effort and operators should confirm via GET /certificate_revocation_list/ca or server logs before treating a compromised agent as contained.
  **Constraint:** OWASP A09:2021 (Security Logging and Monitoring Failures) / CWE-390 (Detection of Error Condition Without Action) - security-relevant failures must not be silently downgraded to a log line at a control the docs describe as reliable.

  This new prose (part of the PR's "containing a compromised agent" guidance, docs/api.md:214-225, which tells operators to revoke a compromised agent's certificate as step 2 of the recommended containment procedure) states DELETE /certificate_status/{subject} 'revokes *and* deletes it immediately' as an unconditional guarantee. The implementation does not honor that: internal/api/handlers.go's handleDeleteStatus (lines 743-763) calls s.CA.Clean(ctx, subject), and internal/ca/signing.go's Clean() (lines 467-520) treats the revoke sub-step as best-effort - if c.revokeLocked fails, the code only logs `slog.Warn("Clean: revoke failed (proceeding with delete)", "subject", subject, "error", err)` (signing.go:499) and continues to `c.Storage.DeleteCert(ctx, subject)` (signing.go:501) regardless. The outer WithLock closure still returns nil in this case (signing.go:512), so Clean() returns nil, handleDeleteStatus never learns the revoke failed, and the client receives 204 No Content (handlers.go:764) - an apparent success. Because internal/api/auth.go's admission check (auth.go:119) consults only the in-memory CRL (myCA.IsRevokedSerial), not certificate-file presence in storage, a certificate whose revoke sub-step failed remains fully valid for authentication even though its on-disk copy was deleted and the API reported success. An operator following the newly-documented containment procedure has no signal that the compromised certificate is still live.
  </details>

- ⚙️ **Containment step 3 ("re-sign on every replica") names no mechanism for reaching every replica** ([`docs/api.md:219`](https://github.com/voxpupuli/openvox-ca/blob/7ecfbc44ae76651e58a1d73f58ad9a6d43d22c13/docs/api.md#L219))

  ```
  3. Force a CRL re-sign on every replica, so no peer keeps admitting the old
     certificate from a stale cache.
  ```

  💡 **Recommendation:** Cross-reference `openvox-ca-ctl reissue-crl` (or the PUT endpoint) from the containment steps and state explicitly that it must be invoked once per replica's own address - not through a shared load balancer/VIP - for step 3 to actually clear every replica's stale cache.
  **Constraint:** Operational documentation for a new/changed procedure must be actionable, not just descriptive (Operator persona: missing operational docs).

  I verified the mechanics against the implementation: `PUT /certificate_revocation_list/ca` (internal/api/handlers.go:377 handleReissueCRL -> internal/ca/crl.go ReissueCRL) updates `c.cachedCRL` only on the process instance that handles that specific request (internal/ca/crl.go:74, `c.cachedCRL = parsedCRL`), and the background periodic refresher (internal/ca/crl.go RefreshCRLIfDue, wired in cmd/openvox-ca/crl_refresh.go) skips re-signing - and therefore skips updating that replica's cache - whenever another replica has already pushed NextUpdate out far enough, exactly as the new callout at docs/api.md:234-236 states. So "force a re-sign on every replica" can only be satisfied by issuing the reissue call against each replica's own address individually; hitting a shared load-balancer/VIP (the normal way HA replicas in this doc's own audience - etcd/redis/postgres/mysql backends per docs/storage-backends.md - are reached) will refresh at most one replica per call. The only existing tool for this, `openvox-ca-ctl reissue-crl` (cmd/openvox-ca-ctl/main.go:320-344), issues a single PUT to whatever one server the client is configured against (grep confirms it is documented only in README.md:43 and docs/operator-cli.md:71, never in docs/api.md). Neither the new containment-ordering list (docs/api.md:214-225) nor the new callout (docs/api.md:227-238) tells the operator this, so an operator following step 3 literally with the one tool the repo ships risks believing containment is complete after refreshing a single replica during a live compromised-agent response.
  </details>

</details>

- 📚 **New prose uses British spelling amid the section's established American spelling** ([`docs/api.md:241`](https://github.com/voxpupuli/openvox-ca/blob/7ecfbc44ae76651e58a1d73f58ad9a6d43d22c13/docs/api.md#L241))

  ```
  the authorisation middleware is only installed when `--tls-cert`/`--tls-key`
  ```

  💡 **Recommendation:** Spell it "authorization" in the new sentence to match this file's existing usage of the term throughout (heading and body), or if the project intends to standardize on British spelling per CONTRIBUTING.md, do so as a separate, dedicated pass across the whole file rather than one word in one new sentence.
  **Constraint:** Internal spelling consistency within a single document/section - the same section is headed "## Authorization tiers" (line 171) and two lines below the new sentence reads "authorization extensions" (line 244), both pre-existing and unchanged by this diff

  This PR's new paragraph in the Authorization tiers section spells the word "authorisation" (British), while every other instance of the word in this file - the section heading itself ("## Authorization tiers"), "authorization_extensions" (JSON field, must stay American as a wire-format name), "authorization-arc OIDs", and the "authorization extensions" phrase immediately following the new text - uses American spelling. CONTRIBUTING.md mandates British English in prose generally, but this file's own established, pre-existing usage for this specific term is American; the new sentence introduces the only British-spelled instance of the word in the document, creating a visible inconsistency right next to unchanged American-spelled text in the same section. Cosmetic only, no functional impact.
  </details>

- 🧪 **BeforeEach convention text has no carve-out for authbaseline_test.go's BeforeAll usage** ([`AGENTS.md:112`](https://github.com/voxpupuli/openvox-ca/blob/7ecfbc44ae76651e58a1d73f58ad9a6d43d22c13/AGENTS.md#L112))

  ```
  - Per-spec setup/teardown belongs in `BeforeEach`/`AfterEach` (or `DeferCleanup`),
    never in package-level `var` initialisers — specs must be isolated.
  ```

  💡 **Recommendation:** Add a short parenthetical or sentence to the AGENTS.md Ginkgo conventions section noting that authbaseline_test.go's two Ordered/BeforeAll containers are a deliberate, narrowly-scoped exception for expensive shared fixture setup, cross-referencing the in-file comment.
  **Constraint:** Tester domain: test architecture / convention adherence and its documentation trail

  internal/api/authbaseline_test.go:255-262 explicitly says its two `Ordered`/`BeforeAll` containers run "against the BeforeEach convention in AGENTS.md, and are deliberate" (shared CA/RSA-key setup that would otherwise cost minutes per spec, plus `ContinueOnFailure` for full-row failure reporting). That justification is sound and specs remain functionally independent (every certificate is re-minted per route). But this same PR edits AGENTS.md's testing-conventions section extensively (lines 122-152, documenting the fingerprint/baseline protocol) without adding a corresponding one-line exception next to the "never in package-level var initialisers - specs must be isolated" / "Do not rely on tests running serially" rules (lines 112-113, 117). A contributor who reads AGENTS.md in isolation - the document that calls itself "authoritative for repository conventions" (line 6) - has no way to know this file is an intentional, reviewed exception rather than a convention violation that slipped through, and could either flag it incorrectly in a future review or copy the pattern into a new file without the same justification.
  </details>

</details>

---

bootc:
All four findings from the review at `7ecfbc4` are addressed, though only two of them ended up where the recommendation pointed. Five further council rounds ran on the fixes; those are the five commits after it, and each round found something wrong with the round before. Rebased onto `main` and force-pushed.

### The four findings

| Finding | Outcome |
| --- | --- |
| `DELETE /certificate_status` doc overclaims revocation | Documented, not fixed in code — see below |
| Containment step 3 names no mechanism | `openvox-ca-ctl reissue-crl` / the `PUT`, per replica, with probes and their conditions |
| British spelling amid American | Declined, see below |
| No AGENTS.md carve-out for `Ordered`/`BeforeAll` | Carve-out added, and the second container converted away so there is one exception rather than two |

### The overclaim, and what I did not change

`Clean` swallows **three** errors, not the one the finding named: a failed revoke, a failed certificate delete and a failed CSR delete are each logged and stepped over, and the handler answers `204` regardless. So a `204` records that the subject existed — not that anything was revoked or removed. Since admission reads the CRL and not storage, the dangerous combination is a delete whose revocation failed: the file is gone and the certificate still authenticates.

I corrected the documentation and left the behaviour alone. Making `Clean` surface those errors changes what `PUT /clean` buckets as an error and what the compose suite asserts about it, which is a separate change with its own blast radius. **It is worth queueing** — the three swallows now have specs recording today's behaviour, so whoever fixes it has a regression anchor and a spec that will tell them exactly what they are changing.

### The spelling finding, declined

`CONTRIBUTING.md` line 76 asks for British English in prose, so the new sentence follows the convention and the file's older instances are the drift. I did not sweep the file either: the remaining occurrences are the `authorization_extensions` JSON field, the authorization-arc OID naming, and a heading whose anchor both `README.md` and `AGENTS.md` link to.

### Worth a second look

- **The containment procedure is substantially rewritten**, and is the part most worth reading. It now distinguishes what reaches shared storage from what reaches a replica's admission cache, names the two probes that report the latter (`GET /certificate_status/{subject}`, and `/ocsp` with a nonce) and the conditions under which each holds, and says plainly which signals cannot answer the question. `docs/storage-backends.md` had been telling operators the opposite for the HA backends.
- **`puppetca_crl_update_failures_total` was documented wrongly in both directions** before this settled. The counted paths and the uncounted ones are enumerated once now, and the metric's own HELP string and the mixin comment — which contradicted the docs — agree with it.
- **`internal/api/authbaseline_test.go` has one `Ordered` container instead of two.** The configuration-axes block only replays cached PEMs, so it is a plain `BeforeEach` now; that also stops two of its specs reissuing a CRL through a shared CA.

### On the review rounds

Five rounds, and the same paragraph was wrong in five different ways. Twice I compressed a reviewer's hedge into a false absolute — "does not *necessarily* move" became "does not cover", "subject to its own response cache" became a probe with no caveat — and then, correcting that, wrote a false absolute the other way ("nothing reports a replica's in-process CRL from outside") that two endpoints falsify. The conditional claim was the true one each time. Round five was the first clause-by-clause pass it survived unchanged.

Round five's own fixes were applied without a sixth review, since that was the agreed limit. They are small — a table cell restructured into a note, path names disambiguated, two log-grep pointers, and three new specs — and each new spec was checked by mutation.

Note for whoever merges: #164 is stacked on this branch and will need rebasing after the force-push.

PR #164, opened 2026-07-29
Renew only our own certificates; make certificate_status admin-only

bootc:
> Stacked on #163, which records the authorisation baseline this PR is the first
> to move. Review that one first; the diff here is against it.

Two authorisation gaps, both on the certificate lifecycle, both closed by taking
Puppet Server's shipped `auth.conf` as the reference.

**`POST /certificate_renewal` would renew a certificate this CA did not issue.**
The middleware admits any certificate that chains to the configured trust anchor,
and the handler then checked only that the CSR's CN matched the presented CN.
Nothing checked the issuer, and nothing checked revocation at the point of
renewal — so a revoked certificate could be exchanged for a fresh valid one, and
under a topology with more than one trusted issuer a certificate from a sibling
CA could obtain one from this CA. `assertOwnValidCertificate` now requires the
presented certificate to verify under this CA's own certificate and not to be
revoked, on both renewal paths.

Upstream's `certificate_renewal` rule is `allow: "*"` plus a handler-level issuer
check, which is exactly this shape: the tier stays open, the handler decides.

**`GET /certificate_status/{subject}` was readable by any agent.** Upstream grants
`certificate_status` and `certificate_statuses` to `pp_cli_auth` only. It is now
`tierAdminOnly`, reachable by the CN allow list or the `pp_cli_auth` extension.

### What review added

`Renew` verified the presented certificate's *provenance* but never bound it to
the *subject* being renewed. A holder of any live certificate this CA issued
could therefore re-key another node — and revoke the incumbent's certificate in
the process. Nothing in `Renew` prevented it; the only thing that did was the
handler passing `subject = clientCN`, so the invariant `Renew`'s own doc comment
claimed to enforce "a second time as defence-in-depth" was in fact enforced once,
by its caller. It is now checked in `Renew`, with its own sentinel
(`ErrRenewalSubjectMismatch`) so that an operator debugging a refusal is not told
a certificate we issued "is not a live certificate issued by this CA", and its own
log line so a cross-subject re-key is distinguishable from a topology problem.

The gate also had to move ahead of `ValidateSubject`. A second issuer's leaf need
not respect this CA's lowercase certname grammar, and `ValidateSubject` returns an
unsentinelled error that the handler can only render as a 500 — pre-empting the
403 this branch documents. Both paths now answer provenance first, each pinned by
its own spec.

`ErrNotInitialized` on either renewal path fell through to a 500; it now answers
`503 CA not ready`, matching the import path.

### Compatibility

`GET /certificate_status/{subject}` is a **behaviour break** for anyone whose
tooling reads statuses with an ordinary agent certificate. This is deliberate and
was chosen over a compatibility shim because the project is pre-1.0 with no
releases, and diverging from upstream's authorisation model is worse than the
break. `allow_public_status` still opens the route to unauthenticated clients for
bootstrapping agents; the routes back for tooling are the CN allow list and
`pp_cli_auth`.

`docs/migrating-from-puppet-server.md` states the symptom, the exact log field to
grep for, and — before the remedies — what they cost: admin is a single boolean,
so both authenticated options grant the full admin tier, not status access alone.
The `pp_cli_auth` route in particular requires signing offline with the CA private
key, because authorisation-arc OIDs are stripped from submitted CSRs, and is
unavailable entirely when the key lives in OpenBao Transit.

The renewal gate is not a compatibility break in the default topology — every
certificate an agent holds was issued by this CA — but it will reject a renewal
attempt from a certificate issued by a *previous* CA whose material was replaced
without re-issuing. Such a certificate already failed the middleware's own chain
check, so it was locked out of every mTLS route before this change too.

### Testing

The gate is pinned at the CA boundary, where its two callers have independent
code paths and each rejection reason can be stated once per path: nil, foreign,
foreign-bearing-a-name-we-issued, revoked, and wrong-subject.

It is also reachable over HTTP, and `api_test.go` reaches it: that fixture points
the middleware at a different anchor than the CA's own certificate, which is the
topology the intermediate-CA work introduces. Under the shipped single-anchor
topology the middleware's accept set is a strict subset of the gate's, so nothing
gets past one to be refused by the other — which is why the gate looks unreachable
today, and why it must not be removed on that basis.

One pre-existing test fixture was wrong and is fixed here: the weak-key renewal
test used a self-signed certificate, so it only ever reached the key-policy check
by accident and would now be refused by the gate first.


bootc:
Pushed \`8a31ad4\` addressing the review.

**The CRITICAL was real and I confirmed the blast radius myself.** \`test/puppet/puppet-stack.sh\` reads \`certificate_status\` with an ordinary autosigned agent certificate; the stack's allow list holds only \`puppet-master\`, and \`compose-puppet.yml\` states outright that no service uses \`--no-tls-required\`, so the middleware applies and both assertions fail deterministically across three required legs.

The status reads now use the master's certificate. The agent-certificate case became the end-to-end regression test for the change itself — it asserts 403, so an ordinary agent certificate reaching that route again fails the stack.

I checked the other suites rather than assuming: \`integration-compose.sh\` and \`migration-test.sh\` run without TLS certs, and \`AuthConfig\` is only wired when both \`--tls-cert\` and \`--tls-key\` are set, so the middleware is absent entirely there. \`redis-stack.sh\` already uses the master's certificate. Only the puppet stack broke.

**The own-goal is fixed and worth naming.** Commit 1 introduced \`middlewareDenied\` so the oracle could tell a middleware rejection from a handler one; commit 2 then gave the two new handler rejections the body \`access denied\` — the exact string the discriminator keys on. They now say \`certificate not eligible for renewal\`, and the comment says why so it does not regress a third time.

Those two branches also had no test, being unreachable while the middleware's trust anchor and this CA's certificate are the same object. The new fixture makes them diverge deliberately — trusting a second issuer for authentication while the CA still refuses to renew its certificates — which is the topology this work exists to support.

Two test-coverage findings were precise and both correct: the \`Renew\` half of the gate never pinned the revocation check (a bare \`CheckSignatureFrom\` would have left every spec passing while a revoked certificate stayed a re-key credential), and the "no prior certificate" spec had stopped exercising \`hadOldCert=false\` because \`DeleteCert\` removes the cert blob while \`LatestSerialForSubject\` reads the inventory. Both fixed, the latter by minting directly with the CA key and asserting the inventory is empty first.

Also: \`AutoRenew\` dereferenced \`presentedCert\` without \`Renew\`'s nil guard; \`ErrForeignCertificate\` claimed "not issued by this CA" for revoked certificates this CA did issue; admin-tier denials were server-side silent and now log CN, path and reason at Warn; and the migration guide now states the symptom, how to find who is affected from the logs, and the two remedies that keep authentication rather than only the one that removes it.

trevor-vaughan-ai:
## 🟢 Review Council: APPROVE

>
> Models used:
> - claude-sonnet-5

_Reviewed at commit [`9e50faa`](https://github.com/voxpupuli/openvox-ca/commit/9e50faa7d44de9344e15ea97b46034e453dd15aa) (deep effort)._

**TL;DR:** A five-persona review approved a PR unconditionally, surfacing only 6 minor findings - 3 medium and 3 low - mostly gaps in test coverage and documentation, no security or correctness issues.

**Findings:** 🔴 0 Critical, 🟠 0 High, 🟡 3 Medium, 🔵 3 Low

|---|---|---|
| 📚 Curator (code) | ✅ Approve | 2 LOW |
| 🧪 Tester (code) | ✅ Approve | 2 MEDIUM, 1 LOW |

- 🧪 **New denyWithLog structured-logging helper has no test asserting its log output** ([`internal/api/auth.go:162`](https://github.com/voxpupuli/openvox-ca/blob/9e50faa7d44de9344e15ea97b46034e453dd15aa/internal/api/auth.go#L162))

  ```
  func denyWithLog(w http.ResponseWriter, r *http.Request, clientCN, reason string) {
  	slog.Warn("Request denied by authorisation middleware",
  		"method", r.Method, "path", r.URL.Path, "client_cn", clientCN, "reason", reason)
  	http.Error(w, "access denied", http.StatusForbidden)
  }
  ```

  💡 **Recommendation:** Add a spec (or extend one of the existing 403 specs) that follows the internal/ca/autosign_test.go pattern: install a slog.NewTextHandler over a bytes.Buffer as the default logger for the duration of the spec, trigger a denial through each of the three denyWithLog call sites, and assert the buffer contains the expected reason string and client_cn - at minimum for the new tierAdminOnly certificate_status denial this PR introduces.
  **Constraint:** Security-sensitive/observability behavior added in the changeset should have both positive and negative-path coverage; this repo has an established pattern for testing slog output tied to a behavioral change.

  denyWithLog is a new function, added in this changeset specifically to give structured denial logging (method, path, client_cn, reason) so operators can correlate a 403 back to a cause - per its own doc comment, 'an operator whose tooling broke against a tier change has nothing to correlate against' otherwise. All three call sites (tierSelfOrAdmin, tierAdminOnly, default) route through it with distinct reason strings ('not an admin and not the subject of the request', 'route requires admin access', 'unclassified route'), but grep across internal/api/auth_test.go and internal/api/authbaseline_test.go for 'slog', 'bytes.Buffer', or 'SetDefault' returns nothing - no test in either file captures the default slog output or asserts on any of the reason strings, the client_cn field, or the Warn level. The project already has an established convention for exactly this: internal/ca/autosign_test.go (lines 176-191) captures slog.Default() into a buffer for the duration of a spec and asserts on its content ('Expect(buf.String()).To(ContainSubstring("misconfigured"))') when a change adds warning-level logging tied to a behavioral change. That pattern was not applied to the new helper here, even though this PR's own description frames the logging as a first-class part of the change alongside the tier move. The HTTP-status behavior (403) that denyWithLog also produces is thoroughly tested via auth_test.go and authbaseline_test.go's pinned-baseline mechanism, so this is a coverage gap in the new observability feature specifically, not in access-control enforcement.
  </details>

- ⚙️ **Breaking-change guidance for existing operators lives only inside the one-time Puppet-Server cutover guide** ([`docs/migrating-from-puppet-server.md:1`](https://github.com/voxpupuli/openvox-ca/blob/9e50faa7d44de9344e15ea97b46034e453dd15aa/docs/migrating-from-puppet-server.md#L1))

  ```
  # Migrating from OpenVox Server (or Puppet Server) to openvox-ca
  ```

  💡 **Recommendation:** Add a CHANGELOG.md entry (or a "Breaking changes" callout at the top of README.md) for this release that states the certificate_status and renewal-eligibility changes and links to the "Authorisation parity" section, so operators upgrading an existing openvox-ca deployment encounter the warning before their monitoring/tooling starts failing in production, not after.
  **Constraint:** Operator persona, Upgrade and Migration Paths: "Breaking changes documented in release notes or changelogs?"

  This diff adds the entire operator-facing explanation of two breaking behavior changes (`certificate_status` becoming admin-only, and the new renewal issuer/revocation/subject gate) as new sections - "Authorisation parity" (line 220) and "Renewal eligibility" (line 272) - inside a document whose own title and the README's description of it (README.md line 122: "Replacing an OpenVox/Puppet Server built-in CA") scope it to a one-time cutover from Puppet Server, not to in-place upgrades of an already-running openvox-ca deployment. An operator who migrated from Puppet Server months ago and is now upgrading openvox-ca to a release containing this change has no README pointer that would send them back into this file. I checked for other channels that could surface the change instead: there is no CHANGELOG file anywhere in the repository (`find` for `CHANGELOG*` under the repo root and `docs/` returns nothing), no mention of "changelog" in any `*.md` file, and no documented semver/breaking-change policy in CONTRIBUTING.md or AGENTS.md. The only automated release-notes mechanism is `.github/workflows/release.yml`'s `gh release create --generate-notes`, which is GitHub's PR/commit-title digest, not a hand-authored explanation of symptom, log signature, and remediation cost - the detail an operator needs to diagnose a sudden `403 access denied` on `GET /certificate_status/{subject}` or `POST /certificate_renewal` without downtime. The `docs/api.md` note (line 250) does link to "Authorisation parity", which mitigates this somewhat for anyone re-reading the API reference, but nothing surfaces the change to an operator who just runs the new binary.
  </details>

- 🧪 **Documented denial-log rendering (text vs. JSON) has no test pinning it** ([`docs/migrating-from-puppet-server.md:232`](https://github.com/voxpupuli/openvox-ca/blob/9e50faa7d44de9344e15ea97b46034e453dd15aa/docs/migrating-from-puppet-server.md#L232))

  ```
  for a `reason` field of `route requires admin access` — rendered
  `reason="route requires admin access"` on stderr, and
  `"reason":"route requires admin access"` when `logfile` is set, since that
  selects the JSON handler. The message is `Request denied by authorisation
  middleware` and the CN is in `client_cn`.
  ```

  💡 **Recommendation:** Add a test (e.g. in internal/api/auth_test.go, or a small internal/api/log_test.go) that installs a slog handler over a buffer, triggers the tierAdminOnly 403 denial, and asserts the log record's message and the `reason="route requires admin access"` / `client_cn` attributes; optionally a cmd/openvox-ca test asserting setupLogger picks NewJSONHandler when LogFile is set. This turns the doc's grep recipe into a pinned regression.
  **Constraint:** Coverage Completeness / Regression anchors - documented behavior a reader is told to literally grep for should be locked down by a test

  This is a precise, actionable troubleshooting instruction: it tells an operator exactly which two string renderings (`reason="..."` on stderr vs `"reason":"..."` in JSON) to grep logs for, and attributes the format switch to `logfile` selecting `slog.NewJSONHandler` vs `slog.NewTextHandler`. I verified both halves of the claim are accurate against source: `internal/api/auth.go:145` calls `denyWithLog(w, r, clientCN, "route requires admin access")`, and `internal/api/auth.go:162-165`'s `denyWithLog` logs `slog.Warn("Request denied by authorisation middleware", "method", ..., "path", ..., "client_cn", ..., "reason", reason)`, matching field-for-field; `cmd/openvox-ca/main.go:67-76` selects `slog.NewJSONHandler` when `cfg.LogFile != ""` and `slog.NewTextHandler(os.Stderr, ...)` otherwise, matching the doc's handler-selection claim. However, no test anywhere in the suite asserts the rendered log line. I searched `internal/api/*_test.go` for `reason`, `slog`, `TextHandler`, `JSONHandler`, and `logfile` - the only hits are unrelated comments and `cmd/openvox-ca/config_test.go`'s config-parsing assertions on `cfg.LogFile` (never on log output format). No `_test.go` file anywhere in the repo (`grep -rln "TextHandler\|JSONHandler\|slog.SetDefault\|CaptureOutput\|logfile"` across `*_test.go`) captures `denyWithLog`'s actual slog output or verifies the text/JSON handler selection in `main.go`'s `setupLogger`. A future refactor that renames the `reason` key, changes the reason string, or swaps handler selection logic would silently break this exact grep recipe with no test to catch the drift - the 403 status code is well tested (`internal/api/auth_test.go:701-708`), but the log-format contract the doc explicitly promises is not.
  </details>

</details>

<details><summary>🔵 LOW (3)</summary>

- 📚 **newAuthMiddleware's tier-model doc comment omits certificate_status from its tierAdminOnly example list** ([`internal/api/auth.go:72`](https://github.com/voxpupuli/openvox-ca/blob/9e50faa7d44de9344e15ea97b46034e453dd15aa/internal/api/auth.go#L72))

  ```
  //   - tierAdminOnly: admin CN only (signing, revocation, generation)
  ```

  💡 **Recommendation:** Add certificate_status (or "status reads") to the tierAdminOnly example list, e.g. "tierAdminOnly: admin CN only (signing, revocation, generation, status reads)".
  **Constraint:** Inline documentation should stay accurate to the code it describes; this comment enumerates example routes per tier immediately above the switch statement this diff modifies.

  This GoDoc-style comment on newAuthMiddleware (lines 67-72) lists example operations covered by tierAdminOnly as "signing, revocation, generation." This diff moves GET /certificate_status/{subject} into tierAdminOnly by default (auth.go line 204, was tierAnyClient), which is the PR's headline behavior change and is documented thoroughly elsewhere in this same diff (the lookupTier case comment at auth.go lines 192-199, the main.go flag help text, and the auth_test.go/authbaseline_test.go comments). This one enumeration was not touched and now under-represents what tierAdminOnly gates by default, since a reader skimming only the four-tier summary above the switch would not learn that status reads are now included.
  </details>

- 📚 **"Renewal eligibility" uses a raw HTML anchor instead of the established heading convention** ([`docs/api.md:69`](https://github.com/voxpupuli/openvox-ca/blob/9e50faa7d44de9344e15ea97b46034e453dd15aa/docs/api.md#L69))

  ```
  <a id="renewal-eligibility"></a>
  ```

  💡 **Recommendation:** Promote the callout to a real subsection heading - e.g. `### Renewal eligibility` under `## Certificate renewal` - dropping the manual `<a id>` tag and letting the standard slug (`#renewal-eligibility`) generate itself, consistent with every other in-file and cross-file link target and with the heading already used for the same content in docs/migrating-from-puppet-server.md.
  **Constraint:** Every other cross-reference target in this docs tree (docs/api.md, docs/configuration.md, docs/migrating-from-puppet-server.md) is a markdown heading (`##`/`###`) relying on its auto-generated slug - e.g. `#certificate-renewal`, `#certificate-import`, `#authorization-tiers`, `#auth-arc-oid-stripping`, `#authorisation-parity`. `grep -rn '<a id=' docs/ README.md` returns exactly one hit in the whole tree: this line.

  The new "Renewal eligibility" callout in docs/api.md is introduced as a bolded paragraph lead-in with a hand-written `<a id="renewal-eligibility"></a>` anchor, rather than as a heading. This is the only manual HTML anchor anywhere in docs/ or README.md - every other link target in this same file (and in migrating-from-puppet-server.md, which links to it) is a `##`/`###` heading whose anchor is auto-generated. The inconsistency is sharpened by the fact that the parallel content in docs/migrating-from-puppet-server.md, describing the identical renewal-eligibility gate, *is* given a proper heading: `### Renewal eligibility` (migrating-from-puppet-server.md:272). So the same concept is a heading in one file and a bold-paragraph-plus-manual-anchor in the other. The link itself still resolves correctly (no broken navigation), but it breaks the single established mechanism this docs tree uses for anchor targets and will read as an outlier to anyone scanning the file's heading structure or building a TOC from headings.
  </details>

- 🧪 **ErrRenewalSubjectMismatch handler branch is untested (and self-documented as unreachable)** ([`internal/api/handlers.go:991`](https://github.com/voxpupuli/openvox-ca/blob/9e50faa7d44de9344e15ea97b46034e453dd15aa/internal/api/handlers.go#L991))

  ```
  if errors.Is(err, ca.ErrRenewalSubjectMismatch) {
  				// Same 403 body as below — the client learns no more either way
  				// — but logged apart, because this one is an authenticated
  				// caller reaching for another node's identity rather than a
  				// topology problem.
  				//
  				// Unreachable from this handler by construction: it passes cn
  				// and the certificate cn came from, so the two always agree.
  ```

  💡 **Recommendation:** Optional: no action required given the documented unreachability. If desired, a narrow unit test could call handlePostCertificateRenewal-adjacent logic indirectly by asserting ca.Renew's ErrRenewalSubjectMismatch maps through a small helper, or simply leave as-is and revisit when a caller that can trigger it is introduced.
  **Constraint:** Coverage completeness - security-relevant HTTP error mapping should be exercised by a test

  This branch maps ca.ErrRenewalSubjectMismatch to a 403 with its own log line, but no test in the changeset (or the pre-existing suite, confirmed via grep for the only production caller at handlers.go:982) can reach it: the handler always calls s.CA.Renew(ctx, cn, body, r.TLS.PeerCertificates[0]) where cn is derived from the same certificate, so presentedCert.Subject.CommonName == subject holds by construction and ca.ErrRenewalSubjectMismatch can never be returned to this call site. The CA-layer condition itself is well covered (renewgate_test.go's "refuses to renew a subject the presented certificate is not for"), but the HTTP-layer mapping - status code, log fields, and body string - has zero coverage. The code's own comment concedes this, so this is a documented, deliberate gap rather than an oversight, but it means a typo in this block (wrong status code, wrong errors.Is target) would not be caught by any test.
  </details>

</details>

---

bootc:
All six findings from the review at `9e50faa` are addressed, five as recommended and one deliberately not. Four further council rounds ran on the fixes; those are the commits after it. Rebased onto the rebased `test/authorisation-baseline` (which is itself rebased onto `main` now that #162 has landed) and force-pushed.

### The six findings

| Finding | Outcome |
| --- | --- |
| `denyWithLog` output untested | Specs assert the message, `reason`, `client_cn`, `path` and `method`, and that a self-or-admin denial carries its own distinct reason |
| Documented text-vs-JSON rendering untested | `setupLogger`'s handler selection and both destinations are pinned |
| Breaking changes only in the cutover guide | A README section with its own heading, linked from the migration blockquote — but no CHANGELOG, see below |
| Tier comment omits `certificate_status` | Says "status reads" now |
| Hand-written `<a id>` anchor | A real `### Renewal eligibility` heading, moved to the end of its parent section so it stops re-parenting the rest |
| `ErrRenewalSubjectMismatch` branch untested | Not taken — you marked it optional, and it is unreachable by construction |

### The CHANGELOG question is yours

The recommendation was a CHANGELOG entry or a README callout. I did the README half. I did not create a CHANGELOG: there is none in the repo, no tags or releases exist yet, and notes come from `gh release create --generate-notes` off PR titles — so introducing one is a release-policy decision rather than something this branch should invent. **Worth settling before 0.9.0**, since these two rules are exactly what such a file exists to carry.

### What the later rounds changed, and why it is worth a glance

The rounds kept finding the same species of defect in my own text: a claim slightly wider than the code supports.

- I wrote that a foreign certificate is "already refused at the TLS layer". It is not — `ClientAuth: tls.RequestClientCert` verifies nothing, as the comment beside it says; `clientCert.Verify` in the authorisation middleware is what answers 403. The README named the wrong enforcement point in the summary an operator reads first.
- I claimed both rules "refuse a request that used to succeed". Both linked sections say otherwise: the status tier matches Puppet Server's shipped `auth.conf`, and in a single-issuer topology nothing reaches the renewal gate that the middleware had not already refused. It also promised an opt-out for renewal, and there is none by design.
- I published two log strings as grep targets and pinned neither — one commit after establishing that publishing a string makes it a contract. Both are asserted now, through `ca.ErrForeignCertificate` rather than a copy of its wording.
- A spec comment said both `setupLogger` callers surface the error as a startup failure; `runSignerMode` logs it and falls back to stderr.

Every new spec was mutation-checked: renaming a message, redirecting stderr to stdout, transposing the Debug and Trace arms, and inverting the resolver each fail the spec that claims to catch them.

### Scope disclosure

The verbosity-to-level table in `cmd/openvox-ca/main_test.go` covers a pre-existing gap, not anything this branch published. It is twenty lines, it caught a transposition, and it sits in the `Describe` that already owns `setupLogger`'s other half — but it is scope creep in a branch about renewal gates, so it is called out rather than smuggled. Say the word and I will lift it into its own PR.

### Rebase notes

`main` moved twice under this stack today. Two adaptations were needed and are their own commits, labelled as such:

- `cmd/openvox-ca/main_test.go` needed its `bytes` import back: this branch dropped it with the buffer, and `main` meanwhile grew a `--version` spec in the same file that uses `bytes.Buffer`.
- Nothing else in this branch conflicted with #162; the CRL-chain collisions all landed in #160.

PR #165, opened 2026-07-29
Let the CA issue and renew its own serving certificate

bootc:
The CA cannot currently issue the certificate its own listener presents. That is
fine when cert-manager or an operator supplies one, and impossible when the CA
key is held at a provider: cert-manager cannot act as a CA issuer without the
key, and `openvox-ca-ctl generate` needs an admin certificate that does not exist
until the CA is already serving. This closes that loop.

```bash
openvox-ca --cadir /etc/puppetlabs/puppet/ssl \
  --hostname puppet.example.com \
  --tls-self-provision
```

The CA mints a `serverAuth`-only leaf for its own hostname, stores it under
dedicated keys, serves it through a holder consulted per handshake, and renews it
on a shared maintenance loop with no listener rebuild and no restart. Mutually
exclusive with `--tls-cert`/`--tls-key`. Off unless asked for: nothing new runs
and no storage key is written otherwise.

### Two things worth reviewing closely

**The certificate is `serverAuth` only, never `clientAuth`.** Its common name is
the CA's own hostname. Where that hostname also appears in `puppet_server`, a
`clientAuth` certificate sitting in the storage backend would be a usable admin
credential.

**A superseded certificate is revoked after 24h by default, not immediately.**
Replicas need time to notice the replacement, so revoking at once would break the
ones still serving it. The delay is durable and shared — the replica that minted
the replacement may die before it elapses — and the pending list is reconciled
rather than merely drained, so turning revocation off discards entries instead of
stranding them to fire much later.

### What a review of this branch changed

Two blocking defects, and they were one trap between them.

`tls_self_provision_encrypt_key` **stopped the server starting.** `crypto/tls`
accepts any PEM block whose type ends `" PRIVATE KEY"`, so an
`ENCRYPTED PRIVATE KEY` block passed the type check and then failed to parse —
fatally, since a serving-certificate failure at startup is. The encrypted blob is
persisted before the assembly runs, so clearing the flag did not recover it
either. The certificate is now built from the signer the CA already decrypted
while validating the stored material.

And the documentation said storing that key unencrypted was "one blob over, not a
new class of exposure" — true only when the CA key is itself a backend blob.
Under `ca_key_provider: openbao`, an external signer, or `ca_key_file`, the
backend holds **no private key at all** until this feature is enabled, and those
are exactly the deployments it exists for. `ca_key_file` does not pin the serving
key, so encryption is the only protection — the flag that would not boot.

Also fixed:

- `ReconcileSuperseded` and the mint path read-modify-write the same pending
  list, and the per-subject lock cannot serialise them: it is derived from each
  replica's hostname, and replicas are explicitly allowed to disagree about that,
  so two of them take *different* subject locks and exclude each other from
  nothing. A replica minting between the sweep's read and its write had its entry
  erased, leaving that certificate valid for its full life with nothing recording
  it. Both paths now take a fixed serving lock outside the subject one.
- A transient read error reported the pending list as empty, so the append path
  wrote a one-entry list over however many were pending.
- `tls_self_provision_renew_before_sec` was unbounded. A window at or beyond the
  certificate lifetime makes every fresh certificate immediately due, so the CA
  reissues on every pass forever. Rejected at startup **and** clamped at runtime,
  because `issueLeafLocked` caps certificates at the CA certificate's *remaining*
  life — a window that was inside the configured lifetime becomes larger than the
  real one as the CA certificate ages, which startup validation cannot see.
- `PuppetCAServingCertRenewalFailing` now ships in the mixin, which the docs told
  operators to alert on twice while no such alert existed. Added
  `puppetca_serving_cert_revocation_failures_total` alongside it: the 24h default
  is only safe if a persistently failing sweep is noticed.

### On the concurrency tests

The filesystem backend implements no `Locker`, so `WithLock` falls back to a
process-local mutex and a convergence assertion between two CA instances on one
service holds whether or not the lock coordinates anything. The specs use a
shared lock table instead, so replicas contend as they would on etcd or SQL.

Four replicas racing must converge on one serial and one mint. The sweep's
serialisation point is asserted **through the lock table rather than by racing**:
the window is a few hundred microseconds, and a timing-based spec passed 10 runs
out of 10 against the buggy code — worse than no test. The lock-table version
fails on the old code and passes on the fix.

Two specs that previously forced a reissue with a 100-year renewal window now
widen the name set: that idiom depended on the loop the clamp removes.

### What later review rounds changed

Nine rounds. The first six are above; rounds 7 to 9 found almost nothing wrong
with the feature and a great deal wrong with its *regression protection* — fixes
that were correct, shipped, and would have reverted silently.

Sixteen such arms are now closed, each verified by reverting the fix and
confirming the named spec fails. The ones worth calling out:

- **The serving lock had no spec.** The change round 6 is named for; deleting the
  wrapper left the whole suite green, because both multi-replica specs gave every
  replica the same subject and so never constructed the defect the lock exists
  for.
- **Nothing anywhere failed a `Put`**, so every write-failure branch was
  unexecuted — including a CWE-532 fix that keeps a SQL driver's DSN out of the
  logs, which reverts to `%w` with the suite green.
- **No spec made `ReconcileSuperseded` return an error**, so both counter
  increments that fire when it does were dead. One lived inside the serve
  command's `RunE`, which no unit spec can execute.
- **The authorisation gate.** Reverting it to the old `tls_cert`/`tls_key` test
  left `AuthConfig` nil under self-provisioning — every route unauthenticated
  over TLS — with the suite green. Its comment said that branch and
  `cfg.tlsEnabled()` "must never disagree", which was an invariant asserted in
  prose and enforced by nothing. They are now one expression.
- **The serving private key's `0600`** was pinned by nothing; flipping its blob
  class to public put a private key at `0644` in a shared cadir.

Behaviour changes from those rounds:

- an unparseable pending list was treated as merely empty. It is now counted as
  the unrecoverable loss it is, the entries that still decoded are swept rather
  than discarded, and the blob is actually overwritten — previously the
  "Discarding" it logged never happened, so the warning repeated every pass
  forever;
- `sql_max_open_conns` below **4** is now refused at startup on PostgreSQL and
  MySQL. Each held lock pins a pooled connection, and this feature adds a third
  nested one; below the floor the sweep blocks until the lock timeout expires on
  every pass, reporting a failure indistinguishable from a backend outage.

### A documentation limitation worth knowing before merge

There is **no by-serial revoke**. If a mint cannot record what it replaced, that
serial is orphaned: no sweep can find it, and `openvox-ca-ctl revoke --certname`
resolves to the certificate the listener is *currently* serving — so the obvious
remedy takes the working credential out of circulation and leaves the orphan
valid. The docs and the alert runbook now say so plainly instead of prescribing
a step that misfires. #177 tracks the fix.

Six issues were filed for work deliberately left out: #173, #174, #175, #176,
#177, #178.


bootc:
I've marked this one as a draft as I'm having second thoughts about it. I think the CA needs functionality like this, but this may not be quite the right approach. I'll have a think about it and come back later. I've also marked the other PRs that include or build on this one as draft as well.

trevor-vaughan-ai:

>
> Models used:
> - claude-sonnet-5

_Reviewed at commit [`9f15097`](https://github.com/voxpupuli/openvox-ca/commit/9f15097e596fc24fcb458e6a19a8d1f7c5eedc06) (deep effort)._

**TL;DR:** Review found 14 verified issues in the tls_self_provision CA feature -- most seriously, an incomplete security fix leaving one certificate-revocation path unguarded -- plus test and doc gaps; no false positives stripped.

**Findings:** 🔴 0 Critical, 🟠 2 High, 🟡 7 Medium, 🔵 5 Low

|---|---|---|
| 🛡️ Adversary (code) | ❌ Changes | 1 HIGH, 1 LOW |
| 📚 Curator (code) | ❌ Changes | 1 MEDIUM, 4 LOW |
| 🧪 Tester (code) | ❌ Changes | 1 HIGH, 5 MEDIUM |

- 🛡️ **AutoRenew lacks the serving-certificate revocation guard just added to Renew** ([`internal/ca/signing.go:898`](https://github.com/voxpupuli/openvox-ca/blob/9f15097e596fc24fcb458e6a19a8d1f7c5eedc06/internal/ca/signing.go#L898))

  ```
  		if !c.RevokeOnAutoRenew {
  			return nil
  		}
  		// Revoke the certificate just replaced, same as Renew's rekey path:
  		// only the newest serial should ever be valid for a subject. Best
  		// effort: a failure here shouldn't undo the renewal the agent is
  		// waiting on.
  		// Lock ordering: subject-lock (held) -> CRL-lock -> c.mu, matching Clean.
  		if err := c.Storage.WithLock(ctx, lockNameCRL, func() error {
  			c.mu.Lock()
  			defer c.mu.Unlock()
  			return c.revokeSerialLocked(ctx, oldSerial)
  		}); err != nil {
  ```

  💡 **Recommendation:** Apply the same `subject == c.Hostname && c.servingSerialMatches(ctx, oldSerial)` check (or factor it into a shared helper called by both Renew and AutoRenew) before the `revokeSerialLocked` call in AutoRenew (signing.go ~898-913), so the CA's live serving certificate cannot be revoked via the mTLS auto-renewal endpoint either.
  **Constraint:** CWE-284: Improper Access Control / incomplete security fix - a protection added to one code path but not an equivalent, reachable path guarding the same asset

  This diff adds a guard to Renew (signing.go:795-800) that skips revocation when `subject == c.Hostname && c.servingSerialMatches(ctx, oldSerial)` - the PR's own commentary explains why: 'if a node ever holds that same certname [as the CA's hostname]... this revoke would take the live serving certificate out of circulation immediately, with none of the delay tls_self_provision_revoke_after_sec exists to give, and every agent doing revocation checking would fail handshakes until the next maintenance pass re-mints.' AutoRenew (signing.go:837-918), the sibling revoke-on-replace path in the same file, hits the identical `revokeSerialLocked(ctx, oldSerial)` call (line 909) with no such check. AutoRenew is reachable directly over HTTP: internal/api/handlers.go:927 calls `s.CA.AutoRenew(r.Context(), r.TLS.PeerCertificates[0])` from the `/certificate_renewal` empty-body handler, gated only by 'chains to our CA and is not revoked' (handlers.go:924-926) - i.e. any client holding a valid, non-revoked certificate whose CN equals the CA's own hostname (obtainable through ordinary CSR signing or autosign, exactly the scenario the Renew comment anticipates) can trigger AutoRenew and, with RevokeOnAutoRenew true (the default per ca.go's New()), immediately revoke the CA's own live serving certificate - an availability failure for every mTLS-verifying client in the fleet until the next maintenance pass re-mints. Because the identical `revokeSerialLocked(ctx, oldSerial)` sink is reachable from both Renew and AutoRenew, and only the former was hardened, the fix in this changeset is incomplete for the vulnerability class it describes.
  </details>

- 🧪 **Mint-time passphrase-resolution failure for an encrypted serving key has no test** ([`internal/ca/serving.go:618`](https://github.com/voxpupuli/openvox-ca/blob/9f15097e596fc24fcb458e6a19a8d1f7c5eedc06/internal/ca/serving.go#L618))

  ```
  	passphrase, _, err := resolvePassphrase(c.KeyPassphrase, c.Storage.CADir())
  	if err != nil {
  		return nil, fmt.Errorf("resolving serving key passphrase: %w", err)
  	}
  ```

  💡 **Recommendation:** Add a spec analogous to the existing decrypt-time ones: set cfg.EncryptKey=true with myCA.KeyPassphrase pointing at a nonexistent or unreadable PassphraseFile, call EnsureServingCert on a CA with no serving certificate stored yet, and assert the call fails, that ServingCertIssued() is not incremented, and that the returned error does not contain the passphrase file path (mirroring the existing detail-must-stay-empty assertions used for the decrypt path).
  **Constraint:** Coverage Completeness - error/negative paths in core, security-relevant functionality must be exercised (Tester detection focus: missing error-path tests).

  marshalServingKey is the encrypt-at-rest half of the tls_self_provision_encrypt_key feature this PR introduces. When cfg.EncryptKey is true, it calls resolvePassphrase before ever writing anything, and resolvePassphrase fails whenever the configured passphrase file cannot be read (internal/ca/keyenc.go:174-179) - a realistic operator misconfiguration (wrong path, permission denied, file deleted after validateTLS ran at startup but before a later maintenance-cycle reissue). This is the mint-side counterpart of a code path the suite otherwise treats as important: the decrypt-side failure (parseServingKey's resolvePassphrase call, when the stored key can no longer be decrypted) is exercised three times in internal/ca/serving_test.go ("reissues rather than failing when the passphrase no longer decrypts" at line 320, "withholds the passphrase file path when the key cannot be decrypted" at line 1138, and the ServingReuseReasonForTest variant). No test in this changeset - or anywhere else in the repository (grep for "EnsureServingCert" across all *_test.go files turns up only cmd/openvox-ca/serving_test.go, internal/ca/serving_test.go, internal/ca/servingconcurrent_test.go and internal/metrics/collector_test.go, none of which combine cfg.EncryptKey=true with an unreadable/misconfigured passphrase source) - ever drives EnsureServingCert through this branch. A regression that dropped the error wrap, logged the raw passphrase-file error (which, per the CodeQL finding already fixed elsewhere in this same file, is exactly the class of leak this PR is careful about on the read side), or let issueServingCertLocked report Issued:true on a failed mint would ship undetected.
  </details>

</details>

<details><summary>🟡 MEDIUM (7)</summary>

- ⚙️ **maintenance_interval_sec has no floor, unlike every sibling knob it gates** ([`cmd/openvox-ca/config.go:247`](https://github.com/voxpupuli/openvox-ca/blob/9f15097e596fc24fcb458e6a19a8d1f7c5eedc06/cmd/openvox-ca/config.go#L247))

  ```
  func (c *serverConfig) maintenanceInterval() time.Duration {
  	if c.MaintenanceIntervalSec > 0 {
  		return time.Duration(c.MaintenanceIntervalSec) * time.Second
  	}
  	return defaultMaintenanceInterval
  }
  ```

  💡 **Recommendation:** Add a lower bound to maintenanceInterval() or to validateTLS() (e.g. reject or clamp MaintenanceIntervalSec below some minimum such as 30s-60s) so the same fail-fast discipline applied to revoke-after, renew-before, and sql_max_open_conns also covers the interval that drives their per-tick cost.
  **Constraint:** Operator-controlled polling frequency for an expensive, lock-holding sweep should be bounded, matching the floor/ceiling validation this same diff applies to every other duration it introduces (severity.md Operator MEDIUM example: 'unquantified perf requirement').

  MaintenanceIntervalSec accepts any positive integer of seconds with no lower bound, and directly sets the ticker interval passed to runMaintenance in main.go (`go runMaintenance(ctx, cfg.maintenanceInterval(), maintenanceTasks)`). When tls_self_provision is on, every tick unconditionally runs supersededRevocationTask (maintenance.go: 'Registered whenever self-provisioning is on, including when the delay is zero'), which calls myCA.ReconcileSuperseded - the operation config.go's own validateTLS documents as holding three nested SQL locks (each pinning a pooled connection) plus a fourth for its own work on session-locked backends, and elsewhere as involving 'a remote round trip under ca_key_provider: openbao'. The diff carefully floors or ceilings every other new duration against this same interval (tls_self_provision_revoke_after_sec must be >= 2x the interval, tls_self_provision_renew_before_sec must stay under leaf validity, sql_max_open_conns must clear a floor of 4) but never bounds the interval itself. An operator who sets maintenance_interval_sec to a small value (e.g. during load testing, or by mistakenly entering seconds where minutes were intended) turns the loop into a tight cycle of nested distributed locks and possible remote KMS round-trips against the storage backend, with no validation in validateTLS to catch it.
  </details>

- 🧪 **Malformed-value coverage not extended to the new TLS self-provision / maintenance env vars** ([`cmd/openvox-ca/config_test.go:724`](https://github.com/voxpupuli/openvox-ca/blob/9f15097e596fc24fcb458e6a19a8d1f7c5eedc06/cmd/openvox-ca/config_test.go#L724))

  ```
  	// Malformed values are silently ignored.
  	It("silently ignores malformed values", func() {
  		clearServerEnv()
  		setEnv("PUPPET_CA_PORT", "not-a-number")
  		setEnv("PUPPET_CA_VERBOSITY", "bad")
  		setEnv("PUPPET_CA_NO_TLS_REQUIRED", "maybe")
  		setEnv("PUPPET_CA_CA_VALIDITY_DAYS", "not-a-number")
  		setEnv("PUPPET_CA_LEAF_VALIDITY_DAYS", "bad")
  		setEnv("PUPPET_CA_CA_PATH_LENGTH", "not-a-number")

  		cfg := &serverConfig{Port: 8140, Verbosity: 0, CAPathLength: -1}
  		applyServerEnv(cfg)

  		Expect(cfg.Port).To(Equal(8140), "Port changed on bad input: got %d, want 8140", cfg.Port)
  		Expect(cfg.Verbosity).To(Equal(0), "Verbosity changed on bad input: got %d, want 0", cfg.Verbosity)
  		Expect(cfg.NoTLSRequired).To(BeFalse(), "NoTLSRequired changed on bad input: want false")
  		Expect(cfg.CAValidityDays).To(Equal(0), "CAValidityDays changed on bad input: got %d, want 0", cfg.CAValidityDays)
  		Expect(cfg.LeafValidityDays).To(Equal(0), "LeafValidityDays changed on bad input: got %d, want 0", cfg.LeafValidityDays)
  		Expect(cfg.CAPathLength).To(Equal(-1), "CAPathLength changed on bad input: got %d, want -1", cfg.CAPathLength)
  	})
  ```

  💡 **Recommendation:** Add PUPPET_CA_TLS_SELF_PROVISION, PUPPET_CA_TLS_SELF_PROVISION_RENEW_BEFORE_SEC, PUPPET_CA_TLS_SELF_PROVISION_ENCRYPT_KEY, PUPPET_CA_TLS_SELF_PROVISION_REVOKE_AFTER_SEC, and PUPPET_CA_MAINTENANCE_INTERVAL_SEC to the "silently ignores malformed values" test (or a parallel entry) with invalid values (e.g. "maybe", "not-a-number") and assert each field stays at its zero/default value, matching the treatment already given to NoTLSRequired.
  **Constraint:** Coverage completeness / established project test convention for TLS-relevant boolean and numeric env vars

  config.go's applyServerEnv (lines 520-547) parses PUPPET_CA_TLS_SELF_PROVISION, _RENEW_BEFORE_SEC, _ENCRYPT_KEY, _REVOKE_AFTER_SEC, and PUPPET_CA_MAINTENANCE_INTERVAL_SEC with the same `if _, err := strconv.ParseBool/Atoi(v); err == nil` pattern as the older fields this exact test already exercises (Port, Verbosity, NoTLSRequired, CAValidityDays, LeafValidityDays, CAPathLength) - a malformed value is silently discarded and the field keeps its prior/zero value. The project already treats this as worth testing specifically for TLS-relevant booleans: NoTLSRequired's malformed path is asserted here with the value "maybe". TLSSelfProvision is now comparably security-relevant - per config.go's tlsEnabled(), it is one of the two ways TLS gets enabled at all - yet no test sets a malformed value for it (or for the four sibling TLS self-provision / maintenance-interval vars) and asserts the field is left unchanged. The config_test.go authors were plainly aware of the typo risk for these six vars: the adjacent "applyServerEnv each variable" table entries added in this diff (lines 656-676) carry the comment "A typo in any of these would leave the feature silently off, or a duration silently at its default, with the only symptom a listener that comes up without TLS" - but that comment addresses typos in the env var *name*, not malformed *values*, which is exactly the gap this test covers for the older fields and leaves uncovered for the new ones.
  </details>

- 📚 **Filesystem backend section omits tls_self_provision's new files and warning** ([`docs/storage-backends.md:36`](https://github.com/voxpupuli/openvox-ca/blob/9f15097e596fc24fcb458e6a19a8d1f7c5eedc06/docs/storage-backends.md#L36))

  ````
  ## Filesystem backend (default)

  All CA state lives under `--cadir`. The on-disk layout matches the OpenVox/Puppet
  Server CA, so you can swap in `openvox-ca` without reorganising your SSL tree:

  ```text
  <cadir>/
  ├── ca_crt.pem              CA certificate
  ├── ca_pub.pem              CA public key
  ├── ca_crl.pem              Certificate Revocation List
  ├── inventory.txt           Issued/revoked certificate log
  ├── private/
  │   ├── ca_key.pem          CA private key                    0600
  │   └── <subject>_key.pem   server-generated private keys     0600
  ├── requests/
  │   └── <subject>.pem       pending CSRs
  └── signed/
      └── <subject>.pem       issued certificates
  ```
  ````

  💡 **Recommendation:** Update the Filesystem backend section's directory tree to include `serving_cert.pem`, `serving_superseded.json`, and `private/serving_key.pem`, and add the same `tls_self_provision` warning callout used in the etcd/Redis/SQL backend sections.
  **Constraint:** Established per-backend documentation pattern in this same file/PR: every other backend section (etcd, Redis/Valkey, SQLite, PostgreSQL, MySQL) received an identical `> **tls_self_provision adds a second private key.**` callout in this changeset (confirmed via grep, occurrences at docs/storage-backends.md:139, 256, 300, 339, 376). The Filesystem backend section is the only one that received no update.

  This PR adds `tls_self_provision`, which under `docs/development/storage-internals.md`'s updated 'Filesystem layout (full)' tree writes three new files for the filesystem backend: `serving_cert.pem`, `serving_superseded.json`, and `private/serving_key.pem` (0600). That internals doc explicitly disclaims itself as contributor-only reference and directs users to `docs/storage-backends.md` for deployment guidance ('Deploying `openvox-ca` needs none of this - see the user-facing storage backends guide instead'). But the Filesystem backend section of storage-backends.md (the doc it points to) still shows the pre-feature tree with only `ca_key.pem` and no serving files, and unlike every sibling backend section it received no note that `tls_self_provision` adds a second private key. A filesystem-backend user reading the doc this PR tells them to consult will not learn that enabling the flag adds `private/serving_key.pem` on disk, nor get the same 'this is a new class of exposure if your CA key lives elsewhere' guidance every other backend section now carries.
  </details>

- 🧪 **Gather-failure survival of the serving counters is untested** ([`internal/metrics/collector.go:234`](https://github.com/voxpupuli/openvox-ca/blob/9f15097e596fc24fcb458e6a19a8d1f7c5eedc06/internal/metrics/collector.go#L234))

  ```
  ch <- prometheus.MustNewConstMetric(c.servingCertIssued, prometheus.CounterValue,
  ```

  💡 **Recommendation:** Add a test that forces `Collector.gather` to fail (e.g. by removing read permission on the storage directory, or another available fault-injection point in `internal/storage`) and asserts `puppetca_collector_scrape_success` is 0 while all four in-process counters (`crlUpdateFailures` plus the three new serving counters) are still exported with their current values.
  **Constraint:** Coverage Completeness - untested code path in core functionality (reviewer-protocol.md); lang-go.md testing_conventions expects the behavior a design comment justifies to be locked down by a test

  The three new serving-cert counters are emitted at lines 234-239, before the `if err != nil { ...; return }` branch at line 241, so they survive a `c.gather` failure - the same placement and rationale ("must not blind operators") already used for `crlUpdateFailures` just above them. No test in collector_test.go ever forces `c.gather` to fail (e.g. by making `ListCerts`/`ListCSRs` return an error) and then asserts that `puppetca_serving_cert_issued_total`, `_renewal_failures_total`, and `_revocation_failures_total` (and `puppetca_crl_update_failures_total`) are still present in the gathered output alongside `puppetca_collector_scrape_success 0`. Both existing tests (`"reports CA certificate and CRL metrics"` and `"keeps the three serving counters distinct"`) only exercise the success path where `gather` returns a nil error, so the one invariant the code's placement is designed to guarantee is unverified for either the pre-existing or the three new counters. Note `StorageService` is a concrete filesystem-backed struct (internal/storage/storage.go:41), not an interface, so injecting a `ListCerts`/`ListCSRs` failure would likely require filesystem-level fault injection (e.g. permission removal) rather than a mock - this is a real testability constraint, not an oversight unique to this diff, which is why the severity is not higher.
  </details>

- 🧪 **No promtool unit test locks down the documented churn-threshold math** ([`mixin/config.libsonnet:75`](https://github.com/voxpupuli/openvox-ca/blob/9f15097e596fc24fcb458e6a19a8d1f7c5eedc06/mixin/config.libsonnet#L75))

  ```
      // servingChurnThreshold or it can never fire at all -- at a 2h interval it
  ```

  💡 **Recommendation:** Add a `promtool test rules` unit-test file exercising at least `PuppetCAServingCertChurning` at the shipped defaults and at a raised `maintenance_interval_sec`, to catch the exact silent-failure scenario the config.libsonnet comment describes.
  **Constraint:** Coverage Completeness - missing edge-case/boundary coverage and missing contract test for a documented fragile invariant (reviewer-protocol.md, severity.md Tester HIGH/MEDIUM examples)

  config.libsonnet documents, in detail, that `PuppetCAServingCertChurning`'s correctness depends on `servingChurnWindow / maintenance_interval_sec` exceeding `servingChurnThreshold`, and gives a concrete example (a 2h maintenance interval silently makes the rule permanently unable to fire at the shipped 6h/4 defaults). This is exactly the kind of boundary-sensitive, easy-to-regress relationship a `promtool test rules` unit-test file (Prometheus's standard alert-rule test format) would lock down by feeding synthetic `puppetca_serving_cert_issued_total` sample series at different `maintenance_interval_sec` scenarios and asserting `PuppetCAServingCertChurning` fires or doesn't. No such test file exists anywhere in the repo (`find ... -iname '*promtool*test*' -o -iname '*rules_test*' -o -iname '*mixin*test*'` returned nothing), and CI (.github/workflows/ci.yml, `mage dev:mixin` → magefile.go:822 `promtool check rules`) only runs `promtool check rules`, which validates YAML/PromQL syntax, not alert-firing behavior. This gap predates the diff (none of the pre-existing alerts have promtool tests either), but the diff adds the one rule in the mixin whose correctness is explicitly threshold-arithmetic-dependent and whose own comment warns it can go permanently silent under a plausible operator configuration change.
  </details>

- 🧪 **This changeset adds three new logical keys (KeyServingCert, KeyServingKey, KeyServingSuperseded) to backend.go's const block, and the author manually mirrors them into etcdLayout (etcd.go:116-118), redisLayout (redis.go:149-151), and fsLayout (filesystem.go:56-58), then manually adds matching DescribeTable Entry lines to etcd_unit_test.go:109-111 and redis_test.go:559-561 (comment at redis_test.go:558: "See the etcd table: an omission here is invisible until runtime.")** ([`internal/storage/etcd_unit_test.go:105`](https://github.com/voxpupuli/openvox-ca/blob/9f15097e596fc24fcb458e6a19a8d1f7c5eedc06/internal/storage/etcd_unit_test.go#L105))

  ```
  		// physicalKey fails closed on an unmapped key, so an omission from
  		// etcdLayout makes the serving certificate unstorable at runtime -- on
  		// exactly the backends this feature exists for -- while the unit suite
  		// stays green.
  ```

  💡 **Recommendation:** Add a structural completeness test (mirroring the migratableSingletons Describe in migrate_test.go) that iterates fsLayout and asserts each key has a corresponding entry in etcdLayout and in redisLayout, e.g. `for key := range fsLayout { Expect(etcdLayout).To(HaveKey(key)); Expect(redisLayout).To(HaveKey(key)) }`. This converts the comment's stated risk into an enforced regression anchor instead of relying on every future contributor remembering to hand-add three parallel Entry lines across three test files.
  **Constraint:** Coverage completeness - regression anchor for a documented silent-failure mode

  This changeset adds three new logical keys (KeyServingCert, KeyServingKey, KeyServingSuperseded) to backend.go's const block, and the author manually mirrors them into etcdLayout (etcd.go:116-118), redisLayout (redis.go:149-151), and fsLayout (filesystem.go:56-58), then manually adds matching DescribeTable Entry lines to etcd_unit_test.go:109-111 and redis_test.go:559-561 (comment at redis_test.go:558: "See the etcd table: an omission here is invisible until runtime."). The comment quoted above states the exact risk plainly: physicalKey fails closed on an unmapped key, so a future key added to backend.go/fsLayout but forgotten from etcdLayout or redisLayout compiles fine and the unit suite stays green, while the backend silently rejects that key at runtime (`unknown key %q` from etcd.go:174 / redis.go:276) - for a TLS-serving feature, that is a runtime failure with nothing caught in CI.
  The author demonstrably recognizes this exact failure class and closed it for one path: migrate_test.go:276-287 (`Describe("migratableSingletons", ...)`) adds a structural test that loops `for key := range fsLayout` and asserts every fsLayout key is present in migratableSingletons, so a forgotten Migrate mapping is now a real test failure, not a silent gap. I confirmed via `grep -n "range fsLayout\|range etcdLayout\|range redisLayout" internal/storage/*_test.go` that this loop-based completeness check exists nowhere else - etcd_unit_test.go and redis_test.go only enumerate physicalKey Entry rows by hand, per key, with no test asserting that the DescribeTable (or etcdLayout/redisLayout themselves) stay in sync with fsLayout or with the KeyServing* constants in backend.go. The same class of bug the author paid down for Migrate remains open for the two remote backends' key-mapping tables, and remains open for any key added after this PR.
  </details>

- 🧪 **toTLSCertificate's nil-Key and nil-Leaf error branches are never exercised** ([`cmd/openvox-ca/serving.go:88`](https://github.com/voxpupuli/openvox-ca/blob/9f15097e596fc24fcb458e6a19a8d1f7c5eedc06/cmd/openvox-ca/serving.go#L88))

  ```
  	if sc.Key == nil {
  		return nil, fmt.Errorf("assembling serving certificate for TLS: no private key resolved")
  	}
  	if sc.Leaf == nil {
  		return nil, fmt.Errorf("assembling serving certificate for TLS: no certificate resolved")
  	}
  ```

  💡 **Recommendation:** Add two small specs that call toTLSCertificate directly with &ca.ServingCertificate{Leaf: someLeaf} (nil Key) and &ca.ServingCertificate{Key: someKey} (nil Leaf), asserting the specific error text for each - the same style already used for servingCertHolder.GetCertificate's nil-holder branch a few lines below in the same file.
  **Constraint:** Coverage Completeness - every branch a function commits to (CLAUDE.md: "if a function exists it must honor its full contract - every error path, edge case, and branch") should be reachable by a test, not merely reasoned about in a comment.

  toTLSCertificate is unexported but lives in package main, the same package as cmd/openvox-ca/serving_test.go, so both guard clauses are directly callable from a test with a hand-built *ca.ServingCertificate{} - no fixture or CA setup required. Neither branch is called anywhere in the diff's test files (grep for "toTLSCertificate" in the repo finds only its declaration and its one call site inside ensureServingCert). Every existing test reaches toTLSCertificate exclusively through ensureServingCert, and ca.EnsureServingCert always returns a ServingCertificate with both Leaf and Key populated on a nil error (confirmed in internal/ca/serving.go's loadUsableServingCert and issueServingCertLocked, both of which construct the struct with Leaf and Key set in every success return), so these two branches are unreachable through the integration path the rest of the suite uses. That makes them cheap, direct-unit-test candidates that are currently untested by any path.
  </details>

</details>

<details><summary>🔵 LOW (5)</summary>

- 📚 **No blog/announcement issue filed for the tls-self-provision capability** ([`cmd/openvox-ca/main.go:829`](https://github.com/voxpupuli/openvox-ca/blob/9f15097e596fc24fcb458e6a19a8d1f7c5eedc06/cmd/openvox-ca/main.go#L829))

  ```
  	f.BoolVar(&tlsSelfProvision, "tls-self-provision", false,
  		"Let the CA issue and renew its own serving certificate (enables HTTPS; requires --hostname, excludes --tls-cert/--tls-key)")
  ```

  💡 **Recommendation:** Once a Docs repo is configured for this project, file a blog-labeled issue announcing tls_self_provision. Suggested command: gh issue create --repo <DOCS_REPO> --title "blog: announce CA self-provisioned TLS serving certificates" --label "blog" --body "cmd/openvox-ca/config.go and cmd/openvox-ca/main.go add tls_self_provision, letting the CA issue and renew its own serving certificate rather than loading tls_cert/tls_key from disk. This unblocks deployments where the CA key lives at a provider (e.g. OpenBao) and cert-manager cannot act as an issuer, and where openvox-ca-ctl generate needs an admin client certificate that does not exist until the server is already serving. Draft an announcement covering the use case, the mutually-exclusive relationship with tls_cert/tls_key, and the new maintenance loop (cmd/openvox-ca/maintenance.go) that handles renewal and superseded-certificate revocation. See docs/configuration.md's Self-provisioned serving certificate section for full behavioral detail."
  **Constraint:** Curator review criterion 4, Blog Opportunity Identification: a changeset introducing a 'new major capability' should have a matching blog-labeled issue filed.

  This changeset adds tls_self_provision, letting the CA issue and renew its own serving certificate from its own CA key and serve it from memory, instead of loading tls_cert/tls_key from disk (cmd/openvox-ca/config.go: TLSSelfProvision and four related fields; cmd/openvox-ca/main.go: the --tls-self-provision/--tls-self-provision-names flags and five PUPPET_CA_TLS_SELF_PROVISION_* env vars; cmd/openvox-ca/maintenance.go: a new shared maintenance loop handling renewal and superseded-certificate revocation). This is a new major capability targeting deployments where nothing else can issue the listener's certificate (e.g. CA key held at an external provider such as OpenBao, where cert-manager cannot act as an issuer) -- it crosses the significance threshold for a blog/announcement opportunity. Reference documentation is already thorough (README.md and docs/configuration.md's 'Self-provisioned serving certificate' section cover flags, config keys, and operational caveats in depth), so this is not a documentation gap -- only the announcement/blog tracking is missing. No Docs repo is configured in this project's context documents (AGENTS.md contains no 'Docs repo' entry), so per the graceful-degradation rule for an unconfigured Docs repo, no duplicate-issue search was run and no issue was filed on the reviewer's behalf.
  </details>

- 🛡️ **promtool release checksum is fetched from the same unauthenticated channel as the artifact** ([`.github/workflows/ci.yml:57`](https://github.com/voxpupuli/openvox-ca/blob/9f15097e596fc24fcb458e6a19a8d1f7c5eedc06/.github/workflows/ci.yml#L57))

  ```
            curl -sSfL --retry 3 -o "${RUNNER_TEMP}/sha256sums.txt" \
              "https://github.com/prometheus/prometheus/releases/download/v${PROMETHEUS_VERSION}/sha256sums.txt"
            (cd "${RUNNER_TEMP}" \
              && grep " prometheus-${PROMETHEUS_VERSION}.linux-amd64.tar.gz$" sha256sums.txt \
              | sed "s| prometheus-.*| prometheus.tar.gz|" \
              | sha256sum -c -)
  ```

  💡 **Recommendation:** If a stronger guarantee is wanted, pin the expected SHA256 digest for the specific `PROMETHEUS_VERSION` inline in the workflow (verified once out-of-band) rather than trusting the sums file fetched alongside the artifact, accepting the added Renovate upkeep the author's comment already weighs. Given this is a build-time tool (not a release artifact of this project) and the gap is already transparently documented, this is optional hardening rather than a blocking issue.
  **Constraint:** NIST SSDF PW.4 (verify acquired software) / CWE-345: Insufficient Verification of Data Authenticity - supply-chain integrity for CI-fetched binaries

  The promtool binary is downloaded over HTTPS and checked against `sha256sums.txt`, but that sums file is fetched from the same GitHub release as the tarball itself, so a party able to substitute the release artifact could substitute the accompanying sums file too - the check only catches truncation/corruption, not asset substitution. If GitHub release assets for prometheus/prometheus were compromised, the forged binary would pass verification, be installed to `/usr/local/bin/promtool` via `sudo install` (line 64), and run inside CI to gate `mage dev:mixin`. The author's own comment (lines 47-54) explicitly acknowledges this exact gap: 'not asset substitution -- anyone able to replace the tarball could replace the sums beside it... it is the one binary here that neither a commit SHA nor the Go checksum database covers.' By contrast, the GitHub Actions used elsewhere in this same job are pinned to commit SHAs (e.g. `actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1`) and `jsonnet` is installed via `go install ...@v0.21.0`, which is verified against the Go checksum database (GOSUMDB) - both stronger integrity guarantees than a same-source SHA256 file.
  </details>

- 📚 **Features list has no bullet for the new self-provisioned TLS capability** ([`README.md:30`](https://github.com/voxpupuli/openvox-ca/blob/9f15097e596fc24fcb458e6a19a8d1f7c5eedc06/README.md#L30))

  ```
  - **Full Puppet CA API compatibility:** all 13 endpoints used by agents and OpenVox Server. See the [HTTP API reference](docs/api.md)
  - **Pluggable storage:** filesystem (default; drop-in compatible with the OpenVox/Puppet Server CA layout), SQLite (single database file), or PostgreSQL / MySQL (MariaDB) / etcd / Redis (Valkey) for HA clusters; CA cert/key can be pinned to local files independently. See the [storage backends guide](docs/storage-backends.md)
  - **Pluggable CA key custody:** keep the CA private key as a local file (default) or delegate it entirely to an OpenBao Transit secrets engine key, which never leaves OpenBao — works identically on a VM (AppRole/token) or in Kubernetes (native ServiceAccount auth, no sidecar). See [OpenBao Transit-engine CA key](docs/openbao-transit.md)
  - **Autosigning:** `true`, glob-pattern file, or executable plugin modes
  - **mTLS support:** optional HTTPS with per-endpoint tier-based client certificate authorization
  ```

  💡 **Recommendation:** Add a Features bullet, e.g. under the mTLS entry: '**Self-provisioned TLS:** the CA can issue and renew its own serving certificate in the background, for deployments an external issuer cannot reach. See [self-provisioned serving certificate](docs/configuration.md#self-provisioned-serving-certificate)'.
  **Constraint:** README `## Features` established convention: every major capability (mTLS, OCSP, Kubernetes export, Prometheus exporter, OpenBao key custody, health probes, etc.) gets its own bullet with a 'See X' link into the relevant doc.

  `tls_self_provision` is documented as a full feature elsewhere in this same PR - its own ~230-line section in `docs/configuration.md`, a new section in `docs/ca-key-security.md`, three new storage keys, and a background maintenance loop with dedicated metrics - comparable in scope to 'Kubernetes export' or 'Prometheus exporter', both of which have their own Features bullets. The README instead only surfaces it inline under the 'HTTPS with mTLS' Quick start example (lines 106-127), not in the Features list a reader scans first.
  </details>

- 📚 **New serving-certificate bullet breaks the list's single-topic convention and names 'Revocation failing' twice** ([`mixin/README.md:16`](https://github.com/voxpupuli/openvox-ca/blob/9f15097e596fc24fcb458e6a19a8d1f7c5eedc06/mixin/README.md#L16))

  ```
  is in use; *Revocation failing* is live wherever `hostname` is set, because the
    startup sweep runs unconditionally, and that is the case with no retry. *Renewal failing*: the CA cannot renew the certificate its own
    listener presents, which is silent until it expires, at which point every
    agent handshake fails at once. *Revocation failing*: a superseded certificate was
  ```

  💡 **Recommendation:** Split the exception clause from the enumeration, e.g.: state once that 'Revocation failing' is the exception to the tls_self_provision requirement, then list all three alerts (Renewal failing / Revocation failing / Churning) once each without repeating a name before its definition. Matching docs/metrics.md's structure (a short lead-in paragraph, then one line per alert) would resolve the duplication and restore the sibling bullets' one-topic-per-item scannability.
  **Constraint:** Established list convention in this file: every sibling top-level bullet (exporter down/CA not ready, CA certificate expiry, CRL expiry, leaf certificates, CRL update failures) is a concise, single-pass description of one alert category, 2-4 lines each.

  The new 'Serving-certificate failures' bullet (mixin/README.md lines 14-27) is a single run-on paragraph covering three alerts, and it names *Revocation failing* twice: first as an informal aside ('*Revocation failing* is live wherever `hostname` is set...') before the three-way enumeration has even started, then again with its actual colon-delimited definition after *Renewal failing*. A reader scanning the bullet for 'three rules' encounters the term twice in different roles, which is disorienting in a way none of the sibling bullets are - they each state one idea once. The underlying facts are consistent with the cleaner treatment in docs/metrics.md (paragraph + table, no duplication), so this is specific to how the content was compressed into a single README bullet, not an error in the underlying explanation.
  </details>

- 📚 **Significant new capability has no blog issue filed, and Docs repo is unconfigured** ([`docs/configuration.md:216`](https://github.com/voxpupuli/openvox-ca/blob/9f15097e596fc24fcb458e6a19a8d1f7c5eedc06/docs/configuration.md#L216))

  ```
  ## Self-provisioned serving certificate

  By default `openvox-ca` serves TLS from `tls_cert` and `tls_key` on disk. With
  `tls_self_provision: true` it instead issues that certificate to itself, from
  its own CA key, and renews it in the background.
  ```

  💡 **Recommendation:** Once a Docs repo is configured, file a `blog`-labeled issue describing this capability (self-provisioned CA serving certificates for deployments where cert-manager/OpenBao can't otherwise issue one) using: `gh issue create --repo <DOCS_REPO> --title "blog: openvox-ca can now issue its own serving certificate" --label blog --body "..."`. Until then, note the opportunity for the maintainers to file manually.
  **Constraint:** Blog opportunity threshold: 'New major capability' (Curator review criterion 4).

  `tls_self_provision` is a new major capability: the CA becomes its own serving-cert issuer with a background renewal loop, three new storage keys, three new metrics, and a dedicated CI validation job - meeting the significance bar for a blog post. No 'Docs repo' is configured in this project's context documents (checked AGENTS.md, CONTRIBUTING.md, README.md, and both CLAUDE.md files in the review context; none define a 'Review Council Configuration' → 'Docs repo' entry), so per the graceful-degradation rule this is reported as a finding rather than a filed or auto-checked issue, and no `gh issue list`/`gh issue create` command is included since there is no repo to target.
  </details>

</details>

---

bootc:
**Superseded. To be replaced rather than merged — left open until the
replacement exists.**

The serving certificate becomes one instance of a general managed-certificate
mechanism (#242), and **its private key never goes near the backing store**.

`serving_cert` and `serving_key` here are singular, non-per-subject blobs — one
keypair for the whole cluster, whatever each replica calls itself. That single
choice is what forces most of this branch: `unionNames`/`carryNames` to converge
a shared blob across replicas that may disagree about hostname,
`withServingLocks` because the hostname-derived subject lock then serialises
nothing, and `tls_self_provision_encrypt_key` to protect a shared key in shared
storage.

The replacement keeps two stores and neither is the backend — local files on a
VM, a Kubernetes Secret in a cluster — so all three premises go with it.
Replicas behind a Service share its DNS name, so there is nothing to converge
and the ordinary subject lock suffices.

**What survives:** the renewal decision (`loadUsableServingCert` /
`servingReason`, including the runtime clamp for a renew-before window that
outgrows the CA certificate's remaining life), extracted as a pure function in
#242; `serverAuth`-only issuance, for the reason argued here; and the
delayed-supersession design, which moves to #174 and is generalised there to
span subjects.

**On the VM case:** largely covered by #189 plus what ships today — `openvox-ca
generate --key-out`, point `tls_cert`/`tls_key` at the result, renew from a
timer, `SIGHUP`. What remains is the Kubernetes case.

Ordering: #189 → #174 (with #176) → this branch's replacement → #167's.

bootc:
Superseded by #242, which covers the CA's own serving certificate as one of the two instances of its managed-certificate mechanism.

Closing rather than rebasing: this draft is 353 commits behind `main` with 13 conflicts, and its design has been carried forward. What it settled that #242's body did not already record — the `tls_self_provision_revoke_after_sec` config surface, the three-state `CSRRateLimit` convention, and the 86400 default that #267 later matched by maintainer ruling — is now noted on #242 so it is not re-derived: https://github.com/voxpupuli/openvox-ca/issues/242#issuecomment-5458491810

This PR stays readable; its review history is the fuller record.

PR #167, opened 2026-07-29
Publish the serving certificate to Kubernetes, and wire it through the chart

bootc:
Publishes the self-provisioned serving certificate into a Kubernetes Secret, and
teaches the chart that the CA can supply its own TLS.

> [!IMPORTANT]
> **This PR is not independently reviewable, and must not merge before its
> dependencies.** It is based on #157 and carries #165 wholesale — see
> [What this includes](#what-this-includes) for exactly which commits are which.

## What this includes

Sixteen commits ahead of `feature/helm-chart`, in three groups:

| Commits | Origin | Review there, not here |
| --- | --- | --- |
| 28 (in the base) | **#157** — the Helm chart | `feature/helm-chart` is the base branch, so these are ancestry rather than duplicates |
| 9 | **#165** — self-provisioned serving certificate | Cherry-picked. #165 is cut from `main` and `charts/` does not exist there, so there is no branch that has both |
| 7 | **This PR** | `6e33895`..`d3f93a5` |

The nine carried from #165 are byte-identical to that branch's tip as of
`26b236b`. They will disappear from the diff once #157 and #165 have landed and
this is rebased onto `main` — which is the intended end state, not a workaround.

**Merge order:** #157, then #165, then this. Merging it earlier drags both in at
whatever revision this branch happens to carry.

## What is actually new here

**Export targets gain `serving_cert` and `serving_key`**, so an Ingress or
Gateway terminating in front of the CA can read a `kubernetes.io/tls` Secret the
CA maintains. Data keys default to `tls.crt`/`tls.key` rather than the
trust-bundle convention, because that is what such a consumer looks for.

Both materials come from a **single source returning the pair together**. Reading
them separately meant two storage round trips with a rotation possible in
between, which publishes a `tls.crt` and `tls.key` from different keypairs — a
Secret that is well-formed and fails every handshake made against it. The source
is the same holder the listener presents from, swapped atomically, so there is no
window inside a process. Two replicas can still order their applies against each
other; the next notification-driven cycle corrects that, and the comment says so
rather than claiming more.

**Two validation rules on `serving_key`,** both about blast radius rather than
correctness:

- rejected on a ConfigMap — those are not encrypted at rest and are readable by
  anything that can `get` them;
- cannot share a target with `cert` or `crl` — a Secret holding `ca.crt` is
  public trust material and gets mounted across the cluster, and letting it
  quietly acquire a `tls.key` entry would extend the serving key's reach to
  every workload that reads it.

The serving *certificate* is public and unrestricted.

**The exported key is plaintext**, even with `tls_self_provision_encrypt_key`
set, because a TLS consumer cannot use an encrypted PEM. The server warns at
startup whenever a `serving_key` target is configured.

**Rotation gets its own wake-up.** The exporter reconciles on `CRLUpdated()`,
which a serving-certificate rotation does not fire.

**A failed read now fails only the targets that asked for the missing material.**
It previously aborted the cycle before the per-target loop, so `recordApply`
never ran and no per-target series was written — and both arms of the shipped
alert put `last_error` on the left-hand side. One transient blip left every
target stale with nothing firing. Failed cycles are also retried on a fixed
two-minute interval, inside the alert's fifteen-minute debounce.

**Chart:** `openvox-ca.tlsConfigured` now recognises `tls_self_provision`, so a
self-provisioning install renders HTTPS probes instead of failing the chart's own
no-TLS precondition. Two new preconditions reproduce server-side refusals at
install time — self-provision without `hostname`, and self-provision alongside
`tls.existingSecret` or `tls_cert`/`tls_key`. `startupProbe` goes from 30×2 to
90×2: the old 60s budget was exactly one lock timeout with nothing left for the
second one self-provisioning adds.

No new `tls.selfProvision.*` values. `AGENTS.md` is explicit that the chart does
not enumerate server settings, and this needs no Kubernetes wiring — so operators
set `config.tls_self_provision` directly and the chart reads it back out of the
merged config.

## Corrections to claims this branch made earlier

Both were mine, and both were wrong:

- The design said a rotated certificate "would sit unexported until the next CRL
  change — potentially months". The commit beneath it in the same stack falsifies
  that: revoking superseded certificates after 24h means the rotation itself
  produces a CRL update a day later. The real bound is ~24 hours, or ~20 days
  with revocation disabled.
- The `ci/` fixture credited the Secret-separation rule to "the chart's own
  validation" when the server enforces it at startup, so it applies however you
  deploy.

## Verification

`mage chart:test` 47/47, `chart:validate` against Kubernetes 1.26 and 1.31
including a new `ci/self-provision-values.yaml` fixture, full Go suite,
`golangci-lint` clean, mixin renders.


bootc:
This PR includes or depends on #165 which I've marked as draft, so I'm marking this one as draft too.

trevor-vaughan-ai:

>
> Models used:
> - claude-sonnet-5

_Reviewed at commit [`d0d749c`](https://github.com/voxpupuli/openvox-ca/commit/d0d749c687808b854abdcf5b806be6f86d9c6d7f) (deep effort)._

**TL;DR:** Review blocks merge: two operational-safety bugs (a mistimed startup probe and a self-resolving alert) plus doc and test coverage gaps in an otherwise solid TLS self-provisioning feature.

**Findings:** 🔴 0 Critical, 🟠 2 High, 🟡 16 Medium, 🔵 9 Low

|---|---|---|
| 🛡️ Adversary (code) | ✅ Approve | 1 MEDIUM, 1 LOW |
| 📚 Curator (code) | ✅ Approve | 3 MEDIUM, 5 LOW |
| 🧭 Guard (code) | ✅ Approve | 1 LOW |
| ⚙️ Operator (code) | ❌ Changes | 2 HIGH, 5 MEDIUM, 2 LOW |
| 🧪 Tester (code) | ✅ Approve | 7 MEDIUM |

- ⚙️ **startupProbe budget comment undercounts sequential lock-bound startup phases** ([`charts/openvox-ca/values.yaml:369`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/charts/openvox-ca/values.yaml#L369))

  ```
  # The old 30 x 2 = 60s was exactly one lock timeout and left nothing for a
  # second. Startup can take two: the bootstrap lock during CA init, and the
  # per-subject lock when the CA issues its own serving certificate. Add a
  # signature that may be a Transit round trip rather than local CPU, and 60s is
  # not enough. 90 x 2 leaves room for both plus margin.
  ```

  💡 **Recommendation:** Recompute the budget to cover three sequential lockTimeout windows (>= 180s) plus real margin -- e.g. failureThreshold: 110-120 at periodSeconds: 2 -- or, if reconcileAtStartup's lock acquisition is believed to always resolve near-instantly in the self-provisioning path (no entries to reconcile on a fresh install, so the lock is held only briefly even under contention), say so explicitly in the comment and cite why, rather than omitting the phase from the count.
  **Constraint:** Operational safety budget must reflect the actual worst-case code path it is sized against (severity.md CRITICAL/HIGH boundary: concrete, evidenced operational risk, not theoretical).

  The comment claims a self-provisioning startup crosses exactly two lockTimeout(60s)-bound critical sections (bootstrap, then the serving-cert mint) plus margin for the Transit signature. Tracing cmd/openvox-ca/main.go shows a third, unaccounted phase runs between them: myCA.Init(ctx) (main.go:590) is followed unconditionally by reconcileAtStartup(ctx, myCA, cfg) (main.go:596) whenever cfg.Hostname != "" -- which config.tls_self_provision requires (see charts/openvox-ca/templates/_helpers.tpl:399's own precondition) -- and only then does ensureServingCert(ctx, myCA, cfg, servingCerts) run at main.go:701. reconcileAtStartup (cmd/openvox-ca/serving.go:220-231) calls myCA.ReconcileSuperseded, which independently does `ctx, cancel := context.WithTimeout(ctx, lockTimeout)` and acquires the same serving+subject lock pair via withServingLocks (internal/ca/serving.go:741-751) -- exactly the same shape EnsureServingCert uses (internal/ca/serving.go:143-151). lockTimeout is 60*time.Second (internal/ca/init.go:62). So a self-provisioning install runs three sequential 60s-capped phases before the listener opens -- bootstrap lock, ReconcileSuperseded's serving/subject lock, EnsureServingCert's serving/subject lock -- not the two the comment counts, and the 'Transit round trip' signature work the comment treats as needing extra room beyond the two locks in fact executes *inside* one of those same 60s-bounded windows (EnsureServingCert's ctx wraps both the lock wait and the mint). Under real lock contention -- which is the deployment this PR ships as its own reference fixture, charts/openvox-ca/ci/self-provision-values.yaml sets replicaCount: 2 -- the true worst case is 3 x 60s = 180s, exactly consuming the new 90 x 2 = 180s budget with none of the claimed margin left for kubelet scheduling jitter, the probe request itself, or a slow OpenBao Transit round trip. A pod that is genuinely still making progress at that point gets killed and restarted by the kubelet, which is precisely the failure mode this change was meant to prevent.
  </details>

- ⚙️ **PuppetCAServingCertRevocationFailing self-resolves for the one-shot startup-sweep failure it exists to catch** ([`mixin/alerts.libsonnet:244`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/mixin/alerts.libsonnet#L244))

  ```
  This alert fires only while the counter is still rising, so the retries are not clearing it: check the CA logs and its storage backend.
  ```

  💡 **Recommendation:** For the non-retrying startup-sweep failure mode, alert on state rather than a rate window -- e.g. a gauge/timestamp for 'last unresolved startup-sweep failure' compared the way PuppetCAKubernetesExportFailing compares last-error/last-success -- so the alert stays firing until an operator acts or the CA restarts, instead of aging out of a fixed window.
  **Constraint:** Alert expr/for design must match the actual retry semantics of the counter it watches, or a real incident silently reads as resolved.

  The alert's expr is `increase(puppetca_serving_cert_revocation_failures_total{...}[servingRevocationWindow]) > 0` with `for: servingRevocationFor` (1h / 15m by default). This design -- and the quoted annotation text -- assumes the counter keeps getting refreshed by ongoing retries for as long as the underlying fault persists, so a firing alert always means 'retries are not clearing it.' That assumption is false for one of the two failure modes this same counter records. mixin/README.md (added in this diff, line 17) states the revocation-sweep counter is live on any CA with `hostname` set: 'the startup sweep runs unconditionally, and that is the case with no retry.' A backend hiccup during that one-shot startup sweep increments the counter exactly once and nothing ever increments it again until the process restarts. `increase()` over the 1h window will show that single increment as an elevated rate for roughly one hour after boot -- long enough to satisfy `for: 15m` and page -- and then, once the increment ages out of the window, the expression returns to `0` and the alert clears, even though the fault was never fixed and the orphaned serving certificate is still a valid, unrevoked credential exactly as before. The same PR demonstrates awareness of this exact class of bug: PuppetCAKubernetesExportFailing deliberately avoids a rate window in favor of comparing last-error/last-success state, with the comment 'a rate window, which would silently resolve between attempts' (mixin/alerts.libsonnet line 283) -- but that reasoning was not applied to the no-retry sub-case of this alert, which has the identical shape (a state that persists until manual intervention, not a recurring rate).
  </details>

</details>

<details><summary>🟡 MEDIUM (16)</summary>

- 📚 **README Kubernetes-export summary is stale after docs/kubernetes-export.md rewrite** ([`README.md:48`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/README.md#L48))

  ```
  - **Kubernetes export (opt-in):** publish the CA certificate and/or CRL into any number of Kubernetes Secrets and ConfigMaps via in-cluster server-side apply, with configurable names, namespaces, data keys, labels, annotations, and Secret `type`; CRL-bearing objects are refreshed whenever the CRL changes. See [Kubernetes export](docs/kubernetes-export.md)
  ```

  💡 **Recommendation:** Update the README bullet and the Documentation table row to mention the serving certificate/key material, mirroring the docs/kubernetes-export.md intro. No Docs repo is configured for this review, so file this manually rather than via an automated issue.
  **Constraint:** README Features summary must stay consistent with the linked reference doc it summarizes; per this PR's own docs/kubernetes-export.md diff, the capability now covers `serving_cert`/`serving_key` and rotation-triggered re-export, neither reflected here.

  The README's Kubernetes export feature bullet still describes only CA-cert/CRL export and CRL-triggered re-export, but this PR rewrote docs/kubernetes-export.md to add a third material (serving_cert/serving_key), serving-certificate-rotation as a second re-export trigger, and per-material failure isolation. The matching Documentation table row (README.md:159, "Publishing the CA cert/CRL into Secrets and ConfigMaps") is equally stale.
  </details>

- 📚 **New mixin/ contributor gate has no matching Prerequisites entry** ([`CONTRIBUTING.md:93`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/CONTRIBUTING.md#L93))

  ```
  - If you touched `mixin/`, run `mage dev:mixin` too — it renders the Jsonnet and
  ```

  💡 **Recommendation:** Add a Prerequisites bullet parallel to the Helm/kubeconform one, e.g. "Only if you are changing the monitoring mixin under mixin/: go-jsonnet (`go install github.com/google/go-jsonnet/cmd/jsonnet@v0.21.0`) and promtool (download from a Prometheus release -- see .github/workflows/ci.yml for the exact steps, since promtool cannot be `go install`ed)." No Docs repo is configured for this review, so file this manually rather than via an automated issue.
  **Constraint:** Established CONTRIBUTING.md convention (Prerequisites section, lines 19-22) is to give a conditional "only if you are changing X" bullet naming the exact tools and install command for area-specific gates -- the Helm/kubeconform entry is the precedent this PR should have mirrored for jsonnet/promtool.

  CONTRIBUTING.md now tells contributors who touched mixin/ to run `mage dev:mixin`, but the Prerequisites section (which already has a parallel conditional entry for the Helm chart's tooling) was not given a matching entry for jsonnet/promtool, so a contributor following the new instruction has no install guidance and the target will just silently print SKIP.
  </details>

- ⚙️ **New promtool install step downloads the full Prometheus release tarball on every CI run with no caching** ([`.github/workflows/ci.yml:55`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/.github/workflows/ci.yml#L55))

  ```
  curl -sSfL --retry 3 -o "${RUNNER_TEMP}/prometheus.tar.gz" \
  ```

  💡 **Recommendation:** Cache the extracted `promtool` binary (or the downloaded tarball) keyed on `PROMETHEUS_VERSION`, mirroring the `actions/cache` pattern already used for the `dist` job's Go build cache, so the download/verify/extract sequence only runs when the pinned version changes.
  **Constraint:** Operator efficiency criterion: redundant downloads/computations that could be cached (delegation prompt explicitly calls out checking for missing caching)

  The new "Install jsonnet and promtool" step in the `check` job (lines 42-65) fetches the entire linux-amd64 Prometheus release tarball from GitHub -- which bundles the `prometheus` server binary, console templates, and `promtool` (confirmed by the extraction path at line 64, `prometheus-${PROMETHEUS_VERSION}.linux-amd64/promtool`, i.e. only one binary out of a full release is used) -- plus a second `sha256sums.txt` fetch for verification, on every single push and pull_request run. Unlike the `dist` job in this same file, which the diff's own surrounding code shows uses `actions/cache@55cc8345...` with a deliberate per-variant keying scheme (lines 216-229, with comments explaining why setup-go's built-in cache is insufficient) specifically to avoid repeating expensive downloads, this new step has no caching at all: two full network round-trips plus a tar extraction happen unconditionally on every run of the `check` job, even though `PROMETHEUS_VERSION` is pinned and changes only when Renovate bumps it. The `check` job is one of the jobs the `ci` aggregate (line 473) waits on, so this adds wall-clock time to every CI run for a binary that is unchanged between runs.
  </details>

- 📚 **kubernetesExport section comment not updated for self-provisioned cert/key export** ([`charts/openvox-ca/values.yaml:219`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/charts/openvox-ca/values.yaml#L219))

  ```
  # Publish the CA certificate and/or CRL into Secrets and ConfigMaps via
  # in-cluster server-side apply. See docs/kubernetes-export.md.
  ```

  💡 **Recommendation:** Update the comment at values.yaml lines 219-220 to mention the self-provisioned serving certificate/key, mirroring the wording already used in the updated README.md row and in the `targets` comment further down in this same file.
  **Constraint:** Cross-reference consistency: documentation describing the same feature across files in the same changeset must stay consistent

  This diff updates charts/openvox-ca/README.md's `kubernetesExport.enabled` row to read "Publish the CA cert, CRL and/or the self-provisioned serving certificate and key into Secrets and ConfigMaps" (README.md line 121), and updates the inline `targets` comment in this same values.yaml file (lines 229-231) to describe `serving_cert`/`serving_key`. But the section-level intro comment directly above `kubernetesExport:` (lines 219-220) was left as the pre-existing text, which still describes the block as publishing only "the CA certificate and/or CRL" -- omitting the new self-provisioned serving certificate/key capability that the same PR adds one paragraph below it. A reader skimming only this top comment (the first thing seen when scrolling to the section) gets an incomplete picture of what the block now does, while the parallel README description was kept in sync.
  </details>

- ⚙️ **Empty extraEnv value for PUPPET_CA_TLS_SELF_PROVISION rejects an install the server accepts** ([`charts/openvox-ca/templates/_helpers.tpl:317`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/charts/openvox-ca/templates/_helpers.tpl#L317))

  ```
  {{- if eq .name "PUPPET_CA_TLS_SELF_PROVISION" -}}
  {{- if hasKey . "value" -}}
  {{- $on = .value -}}
  ```

  💡 **Recommendation:** Mirror effectiveConfig's presence check: only set $on from the extraEnv entry when hasKey is true AND the value is non-empty; treat a present-but-empty value the same as absent (skip it, keep the config-file value) rather than passing it into the boolean parser.
  **Constraint:** The chart's own stated rule for these preconditions, at _helpers.tpl:285-286: "a chart that refuses a configuration the server accepts ... is worse than no precondition."

  selfProvisionEnabled's extraEnv loop (_helpers.tpl:316-324) only checks `hasKey . "value"` before treating an entry as authoritative -- it does not check whether that value is empty. An extraEnv entry `{name: PUPPET_CA_TLS_SELF_PROVISION, value: ""}` therefore sets $on to the empty string, which falls through the boolean-parse arm at line 325-330 and fails the render with "PUPPET_CA_TLS_SELF_PROVISION is \"\", which the chart cannot read as a boolean". But the server (cmd/openvox-ca/config.go:520, `if v := os.Getenv("PUPPET_CA_TLS_SELF_PROVISION"); v != "" { ... }`) skips an empty env value entirely and keeps config.tls_self_provision -- so an install with config.tls_self_provision: true and this extraEnv entry would run fine on the server but is rejected at render time. The sibling helper effectiveConfig (_helpers.tpl:297-299), used for hostname/tls_cert/tls_key, gets this right: its extraEnv condition is `(or (not (hasKey . "value")) (ne (toString .value) ""))`, which treats an empty-string value the same as absent. The mage test suite exercises the equivalent empty-override case for hostname via .Values.env (magefile.go's "an empty environment override does not mask the config value" case) but has no matching case for extraEnv value="" on the self-provision flag, and no code path here handles it the same way.
  </details>

- 🛡️ **Serving-material fetch has no bound, unlike the K8s apply it feeds** ([`internal/k8sexport/exporter.go:284`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/internal/k8sexport/exporter.go#L284))

  ```
  certPEM, keyPEM, err := e.serving.ServingMaterial(ctx)
  ```

  💡 **Recommendation:** Derive a bounded context for fetchMaterials the same way applyTarget does for the apply call (e.g. `ctx, cancel := context.WithTimeout(ctx, applyTimeout); defer cancel()` at the top of fetchMaterials, or a dedicated constant), so a hung storage backend surfaces as a counted, logged per-cycle failure instead of stalling the exporter goroutine for the life of the process.
  **Constraint:** Error Handling and Resilience (I/O failure handling at a security-relevant boundary) -- CWE-400-adjacent: unbounded wait on external I/O in a single-goroutine reconcile loop.

  fetchMaterials (exporter.go:253) is called from ExportAll (exporter.go:173, `m, matErrs := e.fetchMaterials(ctx)`) with the raw context runK8sExporter was started with -- the process-lifetime context, cancelled only at shutdown (cmd/openvox-ca/k8sexport.go:62-88). That call chain now includes e.serving.ServingMaterial(ctx), which in the production wiring is `currentOnly.ServingMaterial` (cmd/openvox-ca/k8sexport.go:226-252) and does `storedPEM, err := c.store.GetServingCert(ctx)` (line 235) -- a call into internal/storage/storage.go:546-550, which itself just forwards ctx to `s.backend.Get(ctx, KeyServingCert)` with no additional bound. A backend that hangs (the surrounding comment at storage.go:538 and k8sexport.go:237-241 both acknowledge SQL-backed storage exists) blocks fetchMaterials, and therefore the single exporter goroutine, indefinitely -- exactly the failure mode `applyTimeout` was introduced to prevent for the K8s API call: "without this a black-holed API-server connection could block the single exporter goroutine (and thus re-export of every target) for as long as the OS transport takes to give up" (exporter.go:31-38). That bound is applied only inside applyTarget (`ctx, cancel := context.WithTimeout(ctx, applyTimeout)`, exporter.go:341) -- after fetchMaterials has already returned. The same unbounded pattern pre-existed for GetCACert/GetCRL, but this diff extends it to the path that reads the serving private key, and a stall here silently freezes every export target (trust bundle included), not just the serving ones, until a restart.
  </details>

- 🧪 **materialErrors.forTarget's cert-before-CRL precedence is untested when a target wants both and both fail** ([`internal/k8sexport/exporter.go:233`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/internal/k8sexport/exporter.go#L233))

  ```
  func (me materialErrors) forTarget(t *Target) error {
  	switch {
  	case t.Cert && me.cert != nil:
  		return me.cert
  	case t.CRL && me.crl != nil:
  		return me.crl
  ```

  💡 **Recommendation:** Add a case to 'Material read failures' in exporter_test.go with a target that sets both Cert and CRL, stubSource returning both certErr and crlErr, and assert which error surfaces (and, if only one is meant to be reported, document that as the contract in a comment on forTarget).
  **Constraint:** Coverage completeness: edge cases where multiple state combinations are dangerous should be exercised, not just each flag in isolation.

  forTarget() returns only the first matching failure for a target (cert checked before CRL, before serving cert, before serving key), so a target configured with both Cert and CRL that fails to read both will report only the certificate error and silently drop the CRL failure from ExportAll's returned error and from the Warn log line. Every existing test exercises a Cert-only failure (exporter_test.go, e.g. 'returns an error and applies nothing when the cert cannot be read') or a CRL-only failure ('returns an error and applies nothing when the CRL cannot be read'), and the one target that sets both Cert and CRL ('applies a Secret with both materials, keys, type and managed-by label', exporter_test.go line ~113) is only exercised on the success path. No test drives a Cert+CRL target through fetchMaterials with both stubSource.certErr and stubSource.crlErr set, so the precedence/information-loss behavior of forTarget is unverified -- a change to the case order, or a future material combined with Cert/CRL, could silently swap or drop which failure is surfaced without any test catching it.
  </details>

- 🧪 **Case-insensitive kubernetes.io/tls Type match is untested, and the differently-cased value is never canonicalised before being applied** ([`internal/k8sexport/config.go:278`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/internal/k8sexport/config.go#L278))

  ```
  if strings.EqualFold(t.Type, secretTypeTLS) && (!t.ServingCert || !t.ServingKey) {
  ```

  💡 **Recommendation:** Add a config_test.go case asserting Validate() accepts a differently-cased Type (e.g. "Kubernetes.Io/TLS") the same way it accepts the canonical string, and either assert the stored Target.Type is canonicalised to secretTypeTLS, or add an exporter_test.go case showing the applied Secret's .Type still equals corev1.SecretTypeTLS for a non-canonical input -- whichever matches the intended contract.
  **Constraint:** Coverage completeness: boundary-sensitive normalisation logic (case handling) needs a positive/negative pair proving the normalisation actually happens, not just that validation activates.

  Validate() deliberately uses strings.EqualFold so a differently-cased Type (e.g. "Kubernetes.IO/TLS") is still recognised and forced to require both serving_cert and serving_key -- unlike Kind, which validate() canonicalises in place (t.Kind = KindSecret / KindConfigMap at lines 223-226). Type is never rewritten to the canonical secretTypeTLS string; build.go's buildSecretApply (line ~101) applies t.Type verbatim via corev1.SecretType(t.Type). Every test in config_test.go and exporter_test.go that exercises the kubernetes.io/tls path uses the exact lowercase literal 'kubernetes.io/tls' (config_test.go lines 119-120, 126, 180; exporter_test.go lines 443, 620), so there is no test proving what happens end-to-end for a differently-cased Type: whether the applied Secret ends up with a non-canonical .type the Kubernetes API server will not recognise as kubernetes.io/tls (and therefore will not enforce the tls.crt/tls.key requirement on, silently defeating the validation this rule exists for).
  </details>

- ⚙️ **Shipped-default 1h windows flap on a continuously persisting failure** ([`mixin/config.libsonnet:73`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/mixin/config.libsonnet#L73))

  ```
  persistent failure resolves and re-fires between passes
  ```

  💡 **Recommendation:** Set the default window comfortably larger than the default maintenance_interval_sec (e.g. 2h window against a 1h interval) so a continuously recurring failure stays inside the window across at least two consecutive maintenance passes and the alert does not resolve between them, or document explicitly in mixin/README.md that operators should widen the window before relying on this alert to stay latched for a persistent fault.
  **Constraint:** Alert thresholds and `for` durations should not produce flapping/repeated re-paging for a continuously persisting fault under the shipped default configuration.

  config.libsonnet documents, in its own comment, that at the default `maintenance_interval_sec` (1h) the `servingRenewalWindow` and `servingRevocationWindow` (both 1h) are 'knife-edge': a failure that keeps recurring on every maintenance pass still causes the alert to 'resolve and re-fire between passes' rather than staying continuously firing. This is not a misconfiguration scenario -- it is the out-of-the-box default shipped by the mixin (servingRenewalWindow: '1h', servingRevocationWindow: '1h' in the same file, paired with the CA's own default maintenance_interval_sec of 1h documented in cmd/openvox-ca/config.go). The remediation note that follows only tells operators to raise the windows in step if they raise the interval; it does not address the flapping that occurs at the default values with zero user misconfiguration. Flapping means on-call gets repeated resolve/re-fire notifications for a single ongoing incident, which is the alert-fatigue pattern this review was asked to check for.
  </details>

- 🧪 **Churn-alert threshold/window coupling has no test verifying it fires** ([`mixin/config.libsonnet:78`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/mixin/config.libsonnet#L78))

  ```
      servingChurnWindow: '6h',
      servingChurnThreshold: 4,
      servingChurnFor: '15m',
  ```

  💡 **Recommendation:** Add a promtool `test_rules` unit-test file (e.g. mixin/tests.yaml) that renders the churn rule and asserts it fires under a synthetic increase() sequence at the shipped defaults, and does not silently pass at the documented breaking point (maintenance_interval_sec raised without a matching servingChurnWindow increase). This locks down the exact regression the config.libsonnet comment warns about.
  **Constraint:** reviewer-protocol.md Coverage Completeness -- boundary-sensitive code needs edge case coverage; severity.md HIGH boundary note: risk requiring an absent misconfiguration is MEDIUM, not HIGH

  The `PuppetCAServingCertChurning` rule (mixin/alerts.libsonnet, `increase(puppetca_serving_cert_issued_total{...}[%(window)s]) > %(threshold)d`) and its config (`servingChurnWindow`/`servingChurnThreshold`/`maintenance_interval_sec` coupling) are new, non-trivial logic: the same comment block documents that `servingChurnWindow / maintenance_interval_sec` must exceed `servingChurnThreshold` or the rule 'can never fire at all' and 'becomes permanently invisible' -- the author's own words for the failure mode. No test exercises this arithmetic or confirms the rule actually fires for a synthetic increasing counter. I searched for alert-rule test coverage: `grep -rn "promtool\|unit_test\|tests.yaml\|test_group" mixin/ magefile.go` finds only `magefile.go:1986: return sh.RunV("promtool", "check", "rules", out)`, which is `promtool check rules` (syntax/lint only), never `promtool test rules` (behavioral unit tests with synthetic series) anywhere in the repo -- confirmed by `find . -iname "*mixin*test*" -o -iname "*rules_test*"` returning nothing. This gap is consistent with the file's pre-existing convention (no alert in alerts.libsonnet, old or new, has a promtool unit test), which is why this sits at MEDIUM rather than HIGH: the failure mode requires an operator misconfiguration (raising `maintenance_interval_sec` without raising `servingChurnWindow` to match) to manifest, and it is already flagged prominently to operators via comments and the mixin/README.md callout box.
  </details>

- 🧪 **New fsLayout entries lack a pinned-path unit test, unlike etcd/redis siblings** ([`internal/storage/filesystem.go:56`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/internal/storage/filesystem.go#L56))

  ```
  	KeyServingCert:       "serving_cert.pem",
  	KeyServingKey:        "private/serving_key.pem",
  	KeyServingSuperseded: "serving_superseded.json",
  ```

  💡 **Recommendation:** Add a DescribeTable for FilesystemBackend.pathFor (or Path) mirroring EtcdPhysicalKey/RedisPhysicalKey, asserting the literal relative path for each fsLayout key including the three new serving_* entries.
  **Constraint:** testing_conventions (lang-go.md) / equivalent coverage across the three backends the diff modifies

  etcd_unit_test.go's EtcdPhysicalKey DescribeTable and redis_test.go's RedisPhysicalKey DescribeTable were both extended in this diff with exact-string Entry rows asserting `physicalKey(KeyServingCert) == "/puppet-ca/serving/cert"` etc. (etcd_unit_test.go:109-111, redis_test.go:559-561). FilesystemBackend has no equivalent test at all: there is no filesystem_test.go, and I grepped the repo for `pathFor`, `FilesystemPath`, and the literal strings `serving_cert.pem` / `private/serving_key.pem` / `serving_superseded.json` outside filesystem.go itself and found no assertion pinning fsLayout's new values. The closest coverage is migrate_test.go's seedCA->Migrate round trip (migrate_test.go:45-47) and MigratePreservesVisibility (migrate_test.go:107-109), but both operate through `dst.Path(c.key)` dynamically and only check that Put/Get round-trip and that permissions are 0644/0600 -- neither checks that the underlying relative path string is `serving_cert.pem`/`private/serving_key.pem`/`serving_superseded.json` specifically. MigrateRoundTripsLocalFileLayout (migrate_test.go:264) spot-checks literal on-disk paths but only for `ca_crt.pem`, `private/ca_key.pem`, and `requests/web01.pem` -- none of the three new keys. A typo in one of these three fsLayout values (e.g. a missing underscore or wrong subdirectory) would pass every existing test.
  </details>

- 🧪 **migratableSingletons completeness guard checks fsLayout only, not etcdLayout/redisLayout** ([`internal/storage/migrate_test.go:282`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/internal/storage/migrate_test.go#L282))

  ```
  		for key := range fsLayout {
  			Expect(listed).To(HaveKey(key),
  				fmt.Sprintf("%q is in fsLayout but not migratableSingletons, so migrate would drop it", key))
  		}
  ```

  💡 **Recommendation:** Extend the migratableSingletons DescribeTable (or add sibling It blocks) to also assert HaveKey for every key in etcdLayout and redisLayout, not just fsLayout.
  **Constraint:** Coverage completeness -- regression anchor for the exact failure mode the test's own comment describes

  The new migratableSingletons test (migrate_test.go:276-287) exists specifically to catch a key present in a backend's layout but missing from migratableSingletons, which the docstring above it (migrate_test.go:271-275) says is 'silently dropped by openvox-ca-ctl migrate' with 'nothing reported.' But Migrate is explicitly backend-agnostic (migrate.go:121-124: 'any pair of Backend implementations may be combined, enabling filesystem->database, database->database and database->filesystem migrations'), and the guard only iterates fsLayout. I grepped for etcdLayout/redisLayout in *_test.go and the only hit is a comment (etcd_unit_test.go:106); there is no assertion that every key in etcdLayout or redisLayout also appears in migratableSingletons. Today all three layouts hold the same 11 keys, so nothing is currently broken, but if a future key were added to etcdLayout or redisLayout without also updating fsLayout and migratableSingletons, a migration sourced from etcd or Redis would silently drop that blob with the existing test suite staying green -- the exact scenario this test was written to prevent, just not for two of the three backends it's meant to cover.
  </details>

- ⚙️ **Startup now blocks on a storage round trip regardless of tls_self_provision** ([`cmd/openvox-ca/main.go:596`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/cmd/openvox-ca/main.go#L596))

  ```
  reconcileAtStartup(ctx, myCA, cfg)
  ```

  💡 **Recommendation:** Either gate the added latency behind a much shorter, dedicated timeout for the startup sweep (distinct from the general 60s lock timeout used for in-band operations), or document the added worst-case boot latency for existing hostname-only deployments so operators can size liveness-probe initialDelaySeconds/failureThreshold accordingly.
  **Constraint:** Startup/liveness implication of the serving-mode startup path (Operator focus area); no bound tighter than the general-purpose lockTimeout is applied to a call that gates listener startup.

  reconcileAtStartup runs synchronously in RunE before the TLS listener is constructed (main.go:590-598), and it is called whenever cfg.Hostname is non-empty -- independent of whether tls_self_provision is configured (serving.go:220-232, doc comment: "deliberately without regard to tls_self_provision"). Internally it calls CA.ReconcileSuperseded, which wraps the whole operation in context.WithTimeout(ctx, lockTimeout) with lockTimeout = 60 * time.Second (internal/ca/init.go:62). hostname is a long-standing, commonly-set flag (docs/container-images.md and docs/configuration.md examples both set --hostname) that predates this feature and is unrelated to serving-certificate self-provisioning. On upgrade, any deployment with hostname set now performs an additional storage lock-and-read (under the fixed "serving" lock plus the per-subject lock) on every boot before the API listener opens, even when tls_self_provision is never used. If the storage backend is briefly degraded at exactly the moment the CA restarts (a plausible coincidence during a coordinated infra rollout or backend blip), startup latency grows by up to 60 seconds with no listener bound to fail liveness/readiness probes against, which can trigger probe-driven restart loops in orchestrated environments. The failure path is non-fatal (logged and continues), so this is a latency/liveness risk rather than an outright startup failure.
  </details>

- ⚙️ **maintenance_interval_sec has no minimum floor, unlike its sibling TLS self-provision knobs** ([`cmd/openvox-ca/config.go:247`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/cmd/openvox-ca/config.go#L247))

  ```
  func (c *serverConfig) maintenanceInterval() time.Duration {
  	if c.MaintenanceIntervalSec > 0 {
  		return time.Duration(c.MaintenanceIntervalSec) * time.Second
  	}
  	return defaultMaintenanceInterval
  }
  ```

  💡 **Recommendation:** Add a minimum floor for maintenance_interval_sec in validateTLS (or unconditionally, since the field is also intended for other future maintenance tenants), consistent with the floor already enforced on tls_self_provision_revoke_after_sec.
  **Constraint:** New configuration surface's operational defaults/bounds (Operator focus area); every other new tls_self_provision_* duration in this changeset (revoke_after_sec, renew_before_sec) is validated against a floor or ceiling in validateTLS, but the interval that drives how often those checks run is not.

  maintenance_interval_sec accepts any positive integer of seconds with no lower bound, and validateTLS (config.go) never checks it directly -- it only uses it to derive servingRevokeFloor() for validating tls_self_provision_revoke_after_sec. An operator (or a bad default pushed via a Helm values override / env var) who sets PUPPET_CA_MAINTENANCE_INTERVAL_SEC to a very small value (e.g. 1) causes runMaintenance's ticker to fire the serving-renewal and superseded-revocation tasks every second. Each pass takes the fixed "serving" lock plus the per-subject lock (and, when a revocation is due, the CRL lock too), and on ca_key_provider: openbao each pass may include a remote KMS round trip for passphrase resolution -- the codebase's own comments elsewhere describe this as "a remote round trip under ca_key_provider: openbao" that should be minimized. There is no validation rejecting an unreasonably small interval the way there is for the revoke-after and renew-before windows.
  </details>

- 🧪 **New concurrency tests run without the race detector** ([`magefile.go:1467`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/magefile.go#L1467))

  ```
  testArgs := append([]string{"test", "-json", "-cover", "-coverprofile=coverage.out"}, pkgs...)
  ```

  💡 **Recommendation:** Add -race to the go test args in Test.Unit (magefile.go), at least for internal/ca and cmd/openvox-ca, given the new concurrency-sensitive serving-certificate rotation code.
  **Constraint:** lang-go.md testing_conventions: "Race detection: go test -race (flag missing race detection in concurrent code tests)"

  internal/ca/servingconcurrent_test.go (new in this changeset) launches goroutines that hit shared mutable state through a custom sharedLockBackend (type sharedLockBackend struct { storage.Backend; mu sync.Mutex; locks map[string]*sync.Mutex; acquired []string }, line 44) and a WaitGroup-coordinated multi-replica race in "issues exactly one certificate when replicas start simultaneously" (const replicas = 4, line 143; go func(r *ca.CA) {...}, line 159). cmd/openvox-ca/serving_test.go and maintenance.go's runMaintenance tests similarly exercise servingCertHolder's atomic.Pointer under concurrent Set/GetCertificate calls. I confirmed via .github/workflows/ci.yml (`run: mage test:unit`) that CI's only unit-test entrypoint is Test.Unit in magefile.go, and that function's go test invocation (quoted above) has no -race flag anywhere in the args. A real data race introduced in the new lock-ordering or atomic-holder code (e.g. a future edit that reads servingCertHolder.current without going through the atomic.Pointer) would pass this suite silently in CI.
  </details>

- 🧪 **mysql/mariadb branches of isSessionLockedSQL are never exercised** ([`cmd/openvox-ca/config.go:426`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/cmd/openvox-ca/config.go#L426))

  ```
  	return backend == "postgres" || backend == "mysql" || backend == "mariadb"
  ```

  💡 **Recommendation:** Add two more validateTLS cases (or a DescribeTable) in tlsmode_test.go asserting the floor is enforced for "mysql" and "mariadb", mirroring the existing "postgres" case.
  **Constraint:** Tester coverage-completeness: equivalence-class branches of a boundary-sensitive validation should each be exercised

  validateTLS's SQL connection-pool floor check (config.go:407, `if isSessionLockedSQL(c.StorageBackend) && c.SQLMaxOpenConns > 0 && c.SQLMaxOpenConns < servingSQLConnFloor`) only fires for backends isSessionLockedSQL treats as session-locked. cmd/openvox-ca/tlsmode_test.go's "rejects a SQL pool too small for the sweep's nested locks" and "accepts an unlimited SQL pool, and one at the floor" specs set cfg.StorageBackend only to "postgres" (positive case) and "sqlite" (negative case, in "does not impose the floor on backends whose locks pin no connection"). I grepped every test file in this changeset (config_test.go, tlsmode_test.go, serving_test.go in both cmd and internal/ca, servingconcurrent_test.go) for "mysql" and "mariadb" and found zero matches: neither positive branch is exercised directly (isSessionLockedSQL is unexported but tlsmode_test.go is `package main`, so it could call it directly) or indirectly through validateTLS. A future edit that drops or typos one of these two backend literals would pass the whole suite silently, leaving the connection-pool floor unenforced for that backend while validateTLS's own doc comment and error message claim it is enforced.
  </details>

</details>

<details><summary>🔵 LOW (9)</summary>

- 🛡️ **promtool download integrity check trusts the same origin as the artifact** ([`.github/workflows/ci.yml:55`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/.github/workflows/ci.yml#L55))

  ```
            curl -sSfL --retry 3 -o "${RUNNER_TEMP}/prometheus.tar.gz" \
              "https://github.com/prometheus/prometheus/releases/download/v${PROMETHEUS_VERSION}/prometheus-${PROMETHEUS_VERSION}.linux-amd64.tar.gz"
            curl -sSfL --retry 3 -o "${RUNNER_TEMP}/sha256sums.txt" \
              "https://github.com/prometheus/prometheus/releases/download/v${PROMETHEUS_VERSION}/sha256sums.txt"
            (cd "${RUNNER_TEMP}" \
              && grep " prometheus-${PROMETHEUS_VERSION}.linux-amd64.tar.gz$" sha256sums.txt \
              | sed "s| prometheus-.*| prometheus.tar.gz|" \
              | sha256sum -c -)
  ```

  💡 **Recommendation:** If stronger integrity is wanted, pin an expected sha256 digest for the specific `PROMETHEUS_VERSION` inline in the workflow (updated by the same Renovate custom manager that already tracks `PROMETHEUS_VERSION`), independent of the sums file fetched at run time. Given the low blast radius (no secrets, non-shipping tool), this is optional hardening rather than a blocking requirement.
  **Constraint:** NIST SSDF PW.4 (verify acquired software) / CWE-494 (Download of Code Without Integrity Check) -- supply-chain integrity for build tooling

  The `promtool` binary used by the new `mage dev:mixin` CI gate is fetched as a release tarball and checksummed against `sha256sums.txt`, but both files are pulled from the same GitHub release over the same channel. An attacker capable of substituting the tarball (e.g. a compromised release/CDN edge) can trivially substitute the accompanying checksums file too, so the check only guards against truncated/corrupted downloads, not deliberate asset substitution. The author's own inline comment (lines 47-49, adjacent to this evidence) already acknowledges this exact gap. Blast radius is limited: the `check` job runs under the workflow's top-level `permissions: contents: read` with no job-level override and no secrets in scope, and `promtool` is used only to lint the Prometheus alerting rules, not shipped in any release artifact -- so a compromised binary here cannot exfiltrate credentials or affect the released product, only potentially corrupt CI output for this one advisory lint step.
  </details>

- 📚 **Self-provisioned serving certificate is a major capability with no blog/tutorial issue** ([`docs/configuration.md:217`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/docs/configuration.md#L217))

  ```
  ## Self-provisioned serving certificate
  ```

  💡 **Recommendation:** No Docs repo is configured for this review (no "Review Council Configuration" Docs repo entry found), so duplicate-checking and automated issue filing were skipped. Recommend the maintainers manually track two content items once a Docs repo is configured: (1) a blog/release-notes post announcing self-provisioned serving certificates and the OpenBao-key-custody use case it unblocks; (2) a tutorial walking through enabling it, rotating a compromised serving key, and turning it back off, distinct from the existing configuration.md reference material.
  **Constraint:** Curator's Blog/Tutorial Opportunity Identification criteria: a new major capability (blog threshold) and a new multi-step operational workflow -- enable, renew, replace a compromised key, turn it off, encryption at rest (tutorial threshold) -- should have a tracked content issue.

  This PR adds `tls_self_provision`, a new major capability (self-issued/renewed serving certificates, new CLI flags, new metrics, new storage keys, new Kubernetes-export material, new Helm-chart preconditions) documented across seven files, but there is no Docs repo configured in this review's context to check for or file a blog/tutorial tracking issue.
  </details>

- 📚 **TLS self-provisioning is a new major capability with no blog issue filed** ([`charts/openvox-ca/README.md:24`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/charts/openvox-ca/README.md#L24))

  ```
  There are two routes to a serving certificate. The one above mounts a Secret you
  supply. The other lets the CA issue and renew its own, which is the route to use
  when nothing else can issue it — with the CA key in OpenBao, cert-manager cannot
  act as a CA issuer:
  ```

  💡 **Recommendation:** If a Docs repo is configured for this project elsewhere, consider filing a `blog` issue describing the new TLS self-provisioning route and the OpenBao Transit deployment topology it unlocks. No action needed against the code/docs in this changeset.
  **Constraint:** Blog opportunity identification: new major capability introduced by the changeset

  This changeset adds a new deployment route to the chart -- the CA issuing and renewing its own serving certificate -- which is what makes an OpenBao-Transit-backed CA key deployable at all (cert-manager cannot act as an issuer for a key it cannot read). That meets the "new major capability" significance threshold for a blog/announcement opportunity. No Docs repo is configured for this review (no "Review Council Configuration" entry found in the invoking project or session tracking), so per the Curator's graceful-degradation rule this is reported as an advisory finding rather than a forge issue-create command. In-repo narrative and reference documentation for the feature (this README section and docs/helm-chart.md's "Letting the CA issue its own certificate") are already thorough, so this is purely a content-marketing opportunity, not a documentation gap.
  </details>

- 🧭 **Self-provision precondition re-checks a guard its enclosing block already established** ([`charts/openvox-ca/templates/_helpers.tpl:397`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/charts/openvox-ca/templates/_helpers.tpl#L397))

  ```
  {{- if and (eq (include "openvox-ca.selfProvisionEnabled" .) "true") (eq (include "openvox-ca.configFullyKnown" .) "true") -}}
  ```

  💡 **Recommendation:** Drop the `(eq (include "openvox-ca.configFullyKnown" .) "true")` clause from line 397 and gate on `openvox-ca.selfProvisionEnabled` alone, matching how the TLS-required check at line 385 relies solely on the outer `configFullyKnown` guard.
  **Constraint:** Zero-Waste / Structural Coherence

  This `if` sits inside the outer `{{- if eq (include "openvox-ca.configFullyKnown" .) "true" -}}` opened at line 382, which is still open here (its matching `{{- end -}}` is at line 413) and is not re-entered or reassigned in between. `openvox-ca.configFullyKnown` is a pure function of `.Values.existingConfigMap`, `.Values.args`, and `.Values.envFrom`, none of which the intervening TLS-required check (lines 382-388) mutates, so re-testing `eq (include "openvox-ca.configFullyKnown" .) "true"` at line 397 always evaluates true and is dead weight. The sibling precondition at line 385 (`ne (include "openvox-ca.tlsConfigured" .) "true"`), which is nested at the same level under the same outer guard, does not re-check `configFullyKnown` -- it relies on the single outer guard, which is the pattern this new block should have followed instead of introducing its own redundant copy.
  </details>

- 📚 **Package doc comment omits the new serving-certificate/key capability** ([`internal/k8sexport/config.go:18`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/internal/k8sexport/config.go#L18))

  ```
  // Package k8sexport publishes the CA certificate and/or CRL into Kubernetes
  // Secrets and ConfigMaps. It is an optional feature: when no targets are
  // configured nothing in this package runs. Objects are reconciled with
  // server-side apply so each export is an idempotent create-or-update.
  ```

  💡 **Recommendation:** Extend the package doc comment to mention that the package can also publish the self-provisioned serving certificate and private key (gated on tls_self_provision), matching the scope described in docs/kubernetes-export.md.
  **Constraint:** Inline documentation format: godoc package comments should accurately summarize package scope (reviewer-protocol.md Review Criteria #2, Inline documentation format)

  This is the package-level doc comment surfaced by `go doc k8sexport` and by any IDE hover on the package -- the first thing a maintainer reads to learn the package's scope. The diff adds a second, security-sensitive material kind (the serving certificate and its private key, via the new `ServingSource` interface, `ErrServingStale`, and the `ServingCert`/`ServingKey` target fields), but this comment still describes the package as publishing only "the CA certificate and/or CRL." A maintainer skimming the package doc would not learn that the package also ever publishes a private key.
  </details>

- ⚙️ **Serving-material staleness skip has no metric, only a Debug log** ([`internal/k8sexport/exporter.go:183`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/internal/k8sexport/exporter.go#L183))

  ```
  slog.Debug("Skipping serving export: this replica is behind",
  ```

  💡 **Recommendation:** Add a lightweight counter or gauge (e.g. puppetca_k8s_export_serving_stale_skips_total or a last-success timestamp per target) recorded alongside the Debug log, so a replica that never clears its stale state becomes visible to monitoring without being treated as a hard error.
  **Constraint:** Runtime observability: long-running/edge-triggered conditions should be visible to operators, not only at Debug verbosity.

  When a replica's serving certificate is behind the copy in storage (ErrServingStale), ExportAll skips the target entirely (line 178-187) without calling e.metrics.recordApply and without incrementing any counter -- the only trace is a slog.Debug line, which is typically disabled in production. The design rationale (documented in the surrounding comments and exercised by the exporter_test.go case "skips serving targets when the source says it is behind, without failing them") is that this is normal, transient, and self-correcting on the replica's next maintenance pass. That reasoning holds for the common case, but it also means there is no operational signal to distinguish a replica that is transiently one rotation behind from one that is permanently stuck never catching up (e.g. because its own maintenance/refresh path stopped running) -- both look identical from the monitoring side: a quiet Secret and Debug-only logs. An operator relying on the shipped last_error alert would have no way to detect a replica stuck in this state indefinitely.
  </details>

- 📚 **Serving-certificate cross-reference targets the general section, not the specific subsection** ([`mixin/README.md:26`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/mixin/README.md#L26))

  ```
  [the metric's notes](../docs/metrics.md#self-provisioned-serving-certificate)
  ```

  💡 **Recommendation:** Point the mixin/README.md link at the more specific anchor: ../docs/metrics.md#superseded-serving-certificates, matching the runbook_url in mixin/config.libsonnet.
  **Constraint:** Cross-reference consistency (documentation references should point to the most relevant existing target, and parallel references to the same content should be consistent)

  This sentence explains that a mint which could not read/write down what it replaced 'never scheduled that serial at all... cannot be retired,' and points readers to 'the metric's notes' for that distinction. The link targets docs/metrics.md's general '### Self-provisioned serving certificate' heading (docs/metrics.md:93), not the '#### Superseded serving certificates' subsection (docs/metrics.md:113) that actually contains the retried/not-retried/cannot-be-retired breakdown being described. The parallel reference for the same underlying content, mixin/config.libsonnet:63 (servingRevocationRunbook), correctly targets the specific 'https://github.com/voxpupuli/openvox-ca/blob/main/docs/metrics.md#superseded-serving-certificates' anchor. Both anchors exist and resolve (verified via grep against docs/metrics.md headings), so this is not a broken link -- just an inconsistency in precision between two references to the same explanation within the same changeset.
  </details>

- ⚙️ **migrate --help text omits the newly-migrated serving certificate material** ([`cmd/openvox-ca-ctl/migrate.go:75`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/cmd/openvox-ca-ctl/migrate.go#L75))

  ```
  Long: `Copy every stored CA asset from a source storage backend to a destination
  backend: the CA certificate and key, public key, CRL, serial, inventory (with
  its integrity HMAC), all pending CSRs and all signed certificates.
  ```

  💡 **Recommendation:** Extend the Long description in cmd/openvox-ca-ctl/migrate.go (or add a follow-up sentence) to mention that the self-provisioned TLS serving certificate, its private key, and pending-supersession state are copied too, matching the migratableSingletons list.
  **Constraint:** Operational documentation should describe the current, full scope of a data-migration operation (severity.md Operator LOW example: "optional doc gap"; criterion 7, Operational Documentation)

  This changeset adds three new fixed blobs to migratableSingletons in internal/storage/migrate.go -- KeyServingCert, KeyServingKey, and KeyServingSuperseded (migrate.go lines 50-52) -- so `openvox-ca-ctl migrate` now also copies the TLS self-provisioning serving certificate, its private key, and the pending-supersession list. The migration mechanism itself correctly picks these up (they are plain entries in the same exhaustively-tested migratableSingletons slice, and migrate_test.go's new `migratableSingletons` Describe block guards against future omissions from the filesystem layout). However, the `migrate` subcommand's own --help/Long description in cmd/openvox-ca-ctl/migrate.go -- the primary place an operator learns what the command does before running it against a production CA -- was not updated and still enumerates only "the CA certificate and key, public key, CRL, serial, inventory ... CSRs and ... signed certificates," with no mention of the serving certificate/key/supersession state. An operator reading only this text could wrongly conclude that serving material needs a separate manual copy step after migration, or be surprised when it appears in the destination store. This file is outside the storage-backends changeset's file list but is the direct downstream consumer of the new migratableSingletons entries and the natural place this coverage should be documented.
  </details>

- 📚 **No blog issue tracked for the self-provisioning capability** ([`cmd/openvox-ca/main.go:865`](https://github.com/voxpupuli/openvox-ca/blob/d0d749c687808b854abdcf5b806be6f86d9c6d7f/cmd/openvox-ca/main.go#L865))

  ```
  f.BoolVar(&tlsSelfProvision, "tls-self-provision", false,
  		"Let the CA issue and renew its own serving certificate (enables HTTPS; requires --hostname, excludes --tls-cert/--tls-key)")
  ```

  💡 **Recommendation:** If a documentation/blog repo exists for this project, configure it under a "Review Council Configuration" section in AGENTS.md so this and future significant-capability changes get a tracked blog issue automatically. In the meantime, manually open a `blog`-labeled issue summarizing the self-provisioning feature (the motivation paragraph already in docs/configuration.md's "Self-provisioned serving certificate" section is a ready-made basis).
  **Constraint:** Curator persona, Review Criteria 4 (Blog Opportunity Identification): a new major capability crossing the significance threshold should have a tracked blog/announcement issue.

  `tls_self_provision` is a new major capability (the CA can now bootstrap and renew its own serving TLS certificate from its own CA key, solving a real chicken-and-egg problem for `ca_key_provider: openbao` and `openvox-ca-ctl generate` deployments), introduced by this diff's new flags (`--tls-self-provision`, `--tls-self-provision-names`), config keys, and CA-layer machinery (`EnsureServingCert`, three new Prometheus counters). It clears the blog-opportunity significance threshold. No `Docs repo` entry is configured in this repository's AGENTS.md/CLAUDE.md (checked via `grep -n -i "docs repo" AGENTS.md`, no match), so I could not search for or file a tracking issue per the persona's Issue Filing Template.
  </details>

</details>

---

bootc:
**Superseded. To be replaced rather than merged — left open until the
replacement exists.**

**A bug in here is carved out as #240 — please do not let it disappear with the
branch.** On `main`, `ExportAll` returns as soon as `fetchMaterials` fails,
before the per-target loop, so `recordApply` never runs and no per-target series
is written. `PuppetCAKubernetesExportFailing` carries
`puppetca_k8s_export_last_error_timestamp_seconds` on the left-hand side of both
arms, so with no series written neither arm can fire. This branch already fixes
it; #240 exists so the fix ships against `main` on its own.

**Retired:** the `serving_cert` / `serving_key` export targets and the paired
`ServingSource.ServingMaterial` behind them. Under the reshape the Secret *is*
the store, so an Ingress reads it directly — one mechanism instead of two. That
also retires the reason the pairing existed: with one write of both materials
there is no rotation window in which `tls.crt` and `tls.key` can come from
different keypairs.

**Carried into the replacement:** `openvox-ca.tlsConfigured` recognising
self-provision, the install-time preconditions plus a new one (in Kubernetes,
replicas must agree on `hostname` — it is the Service DNS name), and the
`startupProbe` budget.

**Also worth keeping:** the rule that a Secret holding `ca.crt` must not also
carry a serving private key stays for export targets, but is deliberately not
extended to component certificates (#243), whose Secret is mounted by the one
component it belongs to.

Ordering: #189 → #174 (with #176) → #165's replacement → this branch's. #240 is
independent and can ship today.

bootc:
Superseded by #243, which covers issuing and renewing component certificates into Kubernetes Secrets, including the chart-rendered RBAC and `ci/` fixtures this draft prototyped.

Closing rather than rebasing: it is 367 commits behind `main` with 37 conflicts, is based on `feature/helm-chart` (#157, since merged), and carries #165 — now also closed — wholesale. Its own note that it was "not independently reviewable, and must not merge before its dependencies" was the honest framing throughout and is why it never landed.

What it established is noted on #243: https://github.com/voxpupuli/openvox-ca/issues/243#issuecomment-5458492002

This PR stays readable; its review history is the fuller record.

PR #168, opened 2026-07-29
Publish a refreshable upstream CRL chain, and scope what export targets publish

bootc:
Publishes ancestor CRLs from a file that can be refreshed, so a sub-CA's chain
stays current, and lets Kubernetes export targets say how much of a chain they
want.

> [!IMPORTANT]
> **This PR is not independently reviewable, and must not merge before its
> dependencies.** It is based on #165 and carries #162 wholesale — see
> [What this includes](#what-this-includes) for exactly which commits are which.

## What this includes

Twenty commits ahead of `main`, in three groups:

| Commits | Origin | Review there, not here |
| --- | --- | --- |
| 9 (in the base) | **#165** — self-provisioned serving certificate | `feature/tls-self-provision` is the base branch, so these are ancestry rather than duplicates |
| 3 | **#162** — CRL chain preservation | Cherry-picked. The maintenance loop this builds on lives here, and #162 is cut from `main`, so no single branch has both |
| 8 | **This PR** | `7959a67`..`dab9826` |

The three carried from #162 are byte-identical to that branch's tip as of
`758707b` — two match by `patch-id`, the third differs only in hunk offsets and
its content diff is empty. They disappear once #162 lands and this is rebased,
which is the intended end state rather than a workaround.

**Merge order:** #165, then #162, then this. Merging it earlier drags both in at
whatever revision this branch happens to carry.

There is also a **release-order constraint that is not a merge-order one**: #162
must be *released* — running on every replica — before `crl_chain_file` is
configured anywhere. A replica on an older build re-signs the CRL as a single
block and silently drops the chain, so one old replica handling one revocation
undoes the feature for the whole fleet. Chain preservation is a no-op on a
single-CRL deployment, so that ordering costs nothing. `docs/configuration.md`
says so where an operator will meet it.

## What is actually new here

**`crl_chain_file`** is a PEM bundle of upstream CRLs, re-read every maintenance
cycle and published alongside this CA's own at
`GET /certificate_revocation_list/ca`, so agents running Puppet's default
`certificate_revocation = chain` can check the whole chain. It is declarative:
whatever the file holds is what gets published. Refresh it however you already
refresh files — a mounted Secret, a sidecar, a CronJob.

Until now an ancestor CRL was only ever as current as what was imported, and
re-import was the only remedy. Three shipped documents said so; all three are
corrected here.

**Every CRL in the file is signature-verified** against a certificate in the
stored CA bundle before it is served, and discarded with a warning and a counter
[...]

PR #182, opened 2026-08-01
Make a revocation take effect on every replica, not just the one that revoked

### LINKED ISSUE #171: Revocation is not enforced across replicas until each re-signs the CRL
On the HA storage backends, revoking a certificate does not stop the other
replicas accepting it. Each process answers revocation from an in-memory CRL
copy that it refreshes only when *it* re-signs the CRL, so a peer keeps
admitting a revoked client until it happens to re-sign or is restarted.

## Why it happens

The mTLS middleware checks revocation through `IsRevokedSerial`
([`internal/api/auth.go`](https://github.com/voxpupuli/openvox-ca/blob/main/internal/api/auth.go)),
which reads `c.cachedCRL` and never touches storage. That field is written in
exactly three places:

- `signCRLLocked` (`internal/ca/crl.go`) — on revoke, reissue, or a refresh that
  actually re-signs
- `bootstrapCA` and `loadCRLCache` (`internal/ca/init.go`) — at startup

`RefreshCRLIfDue` reads the stored CRL and returns early when it is not near
expiry, *without* updating the cache — so the periodic job is not a refresh path
for a replica that has nothing to re-sign.

The CRL artefact itself is fine: `GET /certificate_revocation_list/ca` serves it
from storage, so agents fetching the CRL see the revocation immediately. It is
only the CA's own admission decision that goes stale.

## Impact

On `postgres`, `mysql`, `etcd` and `redis` — the backends documented as
active/active — a revoked agent certificate keeps working against every replica
that did not perform the revocation. With `crl_validity_days` at its default of
30 and the refresh window at a third of that, the window before a peer re-signs
on its own is on the order of weeks. The affected endpoints are everything above
the public tier, including `POST /certificate_renewal`, so a compromised agent
can also renew itself on a stale replica.

`filesystem` and `sqlite` are single-node, so they are unaffected in practice.

## Current mitigation

Documented in `docs/api.md` (added on #162's branch): restart the fleet, or force
a re-sign on each replica, when locking out a compromised agent promptly matters.

## Possible directions

- Have the maintenance loop reload `cachedCRL` when the stored CRL's number has
  advanced, rather than only when this replica re-signs. Cheapest option, bounded
  by the existing tick.
- Use the `crlNotify` signal that already wakes the Kubernetes exporter, on the
  backends that can carry a cross-process notification.
- Drop the cache on the auth path and read through to storage, with a short TTL.
  Simplest to reason about, but puts a storage read on every authenticated
  request.

## How this was found

A documentation review on #162 flagged that the branch claimed revocation
"immediately" cuts off API access. The tier claim was right; the temporal one was
not, and tracing it turned up the cache behaviour. The wording has been corrected
on that branch — this issue is for the underlying behaviour.

bootc:
On the HA backends, revoking a certificate did not stop the other replicas
accepting it. Each process answers revocation from an in-memory copy of the CRL
that it rewrites only when *it* re-signs, so a peer kept admitting a revoked
client until it happened to re-sign on its own — weeks, at the default 30-day
CRL validity — or until someone restarted it.

Fixes #171.

## The fix

`CA.SyncCRLCache` re-reads the stored CRL and installs it when its number has
advanced past the cached copy. A background job (`crl_sync_interval_sec`, 60s by
default) runs it on every replica, so that interval is now the worst-case window
in which a revoked certificate still works somewhere in the fleet.

Polling rather than a notification, which the issue left open as an option.
`crlNotify` is in-process only, and the four HA backends each want a different
mechanism — etcd watches, Redis pub/sub, Postgres `LISTEN`/`NOTIFY`, and for
MySQL nothing at all short of polling. A signal would have to be built and
tested per backend and would still need this loop underneath it for the one that
cannot carry one. A single small blob read on a timer costs the same everywhere.

Three details that are the difference between this working and looking like it
works:

- **No cluster lock.** The swap is monotonic in the CRL number, so a poll that
  races a local re-sign and loses declines to install what it read, and a lower
  number — a read served by a lagging database replica — can never undo an
  applied revocation. The storage read happens outside `c.mu`, so it never
  blocks the auth path.
- **Signature verified before installing.** The number comparison alone would
  prefer any CRL numbered high enough, and installing another authority's list
  would silently unrevoke everything. The same check runs on the startup loader,
  because a restart is the first thing an operator reaches for when the sync is
  refusing, and a startup that skipped it would install exactly what the sync
  declined.
- **Stale OCSP responses dropped.** They are pre-signed for four hours, so
  without invalidating newly-revoked serials the responder would keep asserting
  `good` against the CRL that just contradicted it.

Not gated on `disable_crl_refresh` — that switch is about re-signing, not
propagation.

## Renewal is the one path the window would not have bounded

Every other decision made from a stale CRL is self-limiting: it admits a revoked
certificate for at most one interval, after which the same certificate is
rejected. `POST /certificate_renewal` is not — it mints a *new* certificate with
a fresh serial that no CRL will ever list, because the serial it revokes is the
one it replaces. A revocation racing a renewal would have left the agent a
credential outliving its lockout by years, which is not a window anyone would
call 60 seconds.

Both branches of that route now re-read the CRL from storage and refuse a
revoked certificate with a 403. The CSR branch matters more than the empty-body
one: it also re-keys, so an agent getting through would hold a credential this
CA has never seen the private key of.

## Observability

Every CRL series the exporter published came from storage, so they read
identically on every replica and said nothing about which list a given process
was actually enforcing. A replica stuck on an old CRL looked healthy on all of
them.

- `puppetca_crl_cached_number` — the CRL number this replica is deciding from.
- `puppetca_crl_sync_failures_total` — reloads that failed. Kept distinct from
  `crl_update_failures_total`: that one means "this CA could not amend the CRL",
  this one means "this replica could not catch up with an amendment someone else
  made", and they want different remedies.

The mixin gains a warning on rising sync failures and a **critical** on a lag
that outlives `crlLagFor` (10m). The lag rule has a second arm for a stored CRL
that cannot be read at all — the exporter drops `puppetca_crl_number` in that
case, so the subtraction alone would go quiet in exactly the case worth paging
on — qualified on a successful scrape so a storage outage does not raise this
alongside `PuppetCAScrapeFailing`.

## Behaviour changes to know about on upgrade

- **A CA whose stored CRL was not signed by its own certificate now refuses to
  start**, rather than answering revocation from a list it cannot vouch for.
  Reachable only if the CA certificate was replaced out of band, or an
  ancestor-leading `--crl-chain` was imported. `docs/configuration.md` covers
  both remedies, including the offline one for when no replica is left running
  to accept a `reissue-crl`.
- **`openvox-ca-ctl import` rejects a `--crl-chain` bundle that does not lead
  with a CRL signed by the certificate being imported**, before writing
  anything. Block 0 is this CA's own CRL everywhere in the server, so this
  catches at import what would otherwise surface as a CA that will not boot.
- **Renewal returns 403 for a revoked certificate** where it previously
  succeeded on a lagging replica.

Note the PR deliberately does *not* teach the readers to search the stored blob
for this CA's CRL. An earlier revision did, and it was withdrawn: the re-sign
paths and the metrics exporter still take block 0, so a searching reader
disagreed with them — the cache would hold one list while every re-sign amended
another and the exporter reported a third. Ordering and preserving a real CRL
chain is #162's subject, and it settles the same question the same way.

## What the propagation window does not cover

Documented in `docs/configuration.md`, because an operator sizing an incident
response needs both numbers:

- OCSP responses already handed to a verifier stay valid for up to four hours
  and cannot be recalled. Applies only with `--ocsp-url` enabled.
- Revoking one serial does not revoke another the same subject already holds.
  For a compromised node, `openvox-ca-ctl clean` is the right instrument.

## Relationship to the open PRs

Cut from `main` and independently mergeable; it does not depend on #162, #165 or
#168 and none of them depend on it.

- **#162** rewrites `readStoredCRL` to take block 0 of a preserved CRL chain and
  to refuse a foreign CRL centrally. It agrees with this PR on the convention
  (ours leads), so the two compose; on rebase, this PR's `verifyOwnCRLLocked`
  should collapse into #162's `ownsCRL`.
- **#165** introduces a shared `runMaintenance` loop. `runCRLSync` is a natural
  tenant, but it wants its own interval — a minute, against that loop's much
  longer one — so folding it in is a decision for whoever rebases.

## Not fixed here

The OCSP responder answers `unknown` for a certificate signed on another
replica, because `serialIndex` is also built once at `Init` and never refreshed.
Same staleness shape, but `unknown` is not fail-open and the mTLS admission path
does not consult that index at all. Filed as #183.


trevor-vaughan-ai:

>
> Models used:
> - claude-opus-5
> - claude-opus-5[1m]
> - claude-sonnet-5

_Reviewed at commit [`269254d`](https://github.com/voxpupuli/openvox-ca/commit/269254d61a76cbde43d6d932693551c99aa9408d) (deep effort)._

**TL;DR:** Council requests changes: unsigned CRLs could be re-signed as authoritative, renewal-revocation tests never exercise the code they claim to guard, coverage gaps remain

**Findings:** 🔴 0 Critical, 🟠 3 High, 🟡 9 Medium, 🔵 9 Low

|---|---|---|
| 🛡️ Adversary (code) | ❌ Changes | 1 HIGH, 1 MEDIUM |
| 📚 Curator (code) | ✅ Approve | 6 LOW |
| 🧪 Tester (code) | ❌ Changes | 2 HIGH, 7 MEDIUM, 3 LOW |

- 🛡️ **Re-sign paths trust the stored CRL's Number and revoked-entry list without the ownership check this PR adds elsewhere** ([`internal/ca/crl.go:109`](https://github.com/voxpupuli/openvox-ca/blob/269254d61a76cbde43d6d932693551c99aa9408d/internal/ca/crl.go#L109))

  ```
  func (c *CA) reissueCRLLocked(ctx context.Context) error {
  	crl, err := c.readStoredCRL(ctx)
  	if err != nil {
  		return err
  	}
  	return c.signCRLLocked(ctx, crl.Number, crl.RevokedCertificateEntries)
  }
  ```

  💡 **Recommendation:** Route readStoredCRL through the same ownership check before its result is trusted for re-signing - e.g. have it call c.verifyOwnCRLLocked(crl) (or reuse ownStoredCRLLocked's parse+verify) before returning, so reissueCRLLocked and RefreshCRLIfDue fail closed on a foreign or unsigned stored CRL exactly as SyncCRLCache and loadCRLCache now do, instead of re-signing and persisting whatever entries it contains.
  **Constraint:** CWE-347 (Improper Verification of Cryptographic Signature) / CWE-20 (Improper Input Validation) at a trust boundary this same PR establishes: storage holding the CRL blob is untrusted content that must be verified as signed by this CA before any of its fields are relied upon.

  This PR adds ownStoredCRLLocked (internal/ca/crlsync.go:135-144), which verifies a stored CRL's signature against c.CACert before SyncCRLCache or loadCRLCache will trust it - the SECURITY-tagged comment at crlsync.go:66 states the point explicitly ("install only a CRL this CA signed"), and ownStoredCRLLocked's own doc comment (crlsync.go:120-127) argues the check must be universal: "gating only the sync would be theatre... a bundle whose first block is an ancestor's ... could silently drop [a revocation]". Two other readers of the same storage-backed CRL blob do not go through that check. readStoredCRL (internal/ca/crl.go:122-136) parses the blob with no signature verification, and both of its callers pass its Number and RevokedCertificateEntries straight into signCRLLocked, which re-signs them with this CA's real key and writes the result back to storage: reissueCRLLocked (crl.go:109-115, quoted above) and RefreshCRLIfDue (crl.go:144-166, invoked automatically on a timer via DefaultCRLRefreshBefore). If a CRL not signed by this CA ends up in storage - the exact scenario this PR's own tests exercise for the sync/load paths (crlsync_test.go's "refuses a stored CRL this CA did not sign" and "refuses a chain that leads with an ancestor's CRL") - SyncCRLCache and loadCRLCache correctly refuse it, but the next scheduled RefreshCRLIfDue tick (or a manual ReissueCRL) reads that same foreign blob via readStoredCRL, adopts its RevokedCertificateEntries wholesale (which could be an empty list), and produces a freshly and validly CA-signed CRL built from it - persisting the result to storage. That silently launders unauthenticated revocation data (including the absence of real revocations) into an authoritative, CA-signed CRL, undoing every previous revocation. The precondition (something already able to write directly to the shared storage backend, bypassing the normal revoke/reissue API) is the same precondition the new ownStoredCRLLocked check treats as real and worth two dedicated tests and an ImportCA-time gate; the fix is applied to the read-into-cache paths but not to the read-and-re-sign paths in the same file. CORRECTION (independent validator): readStoredCRL has four callers, not two. Besides reissueCRLLocked (crl.go:110) and RefreshCRLIfDue (crl.go:152), revokeSerialLocked (internal/ca/revoke.go:83) appends the new revocation to the unverified crl.RevokedCertificateEntries and re-signs (revoke.go:106-109), and dropCRLEntriesLocked (internal/ca/cleanup.go:115) filters and re-signs the same unverified list (cleanup.go:132). A plain Revoke on a poisoned store therefore also persists a CA-signed CRL built from foreign entries, dropping every prior revocation while appearing to add one.
  </details>

- 🧪 **New revoked-cert renewal tests never reach the code they claim to guard** ([`internal/api/api_test.go:1345`](https://github.com/voxpupuli/openvox-ca/blob/269254d61a76cbde43d6d932693551c99aa9408d/internal/api/api_test.go#L1345))

  ```
  		It("should return 403 when renewing a revoked certificate (empty body)", func() {
  			Expect(myCA.Revoke(context.Background(), subject)).To(Succeed())

  			req := httptest.NewRequest("POST", "/certificate_renewal", bytes.NewReader(nil))
  			req.TLS = &tls.ConnectionState{PeerCertificates: []*x509.Certificate{clientCert}}
  			rr := httptest.NewRecorder()
  			mux.ServeHTTP(rr, req)
  			Expect(rr.Code).To(Equal(http.StatusForbidden))
  		})
  ```

  💡 **Recommendation:** Give the renewal-specific 403 test its own replica pair, mirroring the auth_test.go cross-replica test but posting to /certificate_renewal instead of GETting a resource: sign+revoke through a `signer` CA, then send the renewal request through a `replica` mux that has NOT yet called SyncCRLCache. That forces the middleware to admit the stale cert while the handler's `refuseIfRevoked` (which syncs from storage itself) catches it - the only way to actually exercise handlers.go's new error-mapping branches.
  **Constraint:** Tester coverage-completeness criterion: a regression test must actually exercise the code path it is written to protect.

  Both new tests (this one and its CSR-body sibling at line 1355) revoke the subject through `myCA` - the exact `*ca.CA` instance that also backs the request's auth middleware (`server = api.New(myCA)` in this file's BeforeEach; `Routes()` passes `s.CA` to `newAuthMiddleware`, confirmed at internal/api/handlers.go:133). `ca.CA.Revoke` updates `c.cachedCRL` synchronously via `signCRLLocked` (internal/ca/crl.go:74 `c.cachedCRL = parsedCRL`), so by the time the test's HTTP request arrives, `myCA.cachedCRL` already reflects the revocation. The auth middleware unconditionally checks `myCA.IsRevokedSerial` for every non-public tier - including `tierAnyClient`, which is what `POST /certificate_renewal` maps to (internal/api/auth.go: `case method == "POST" && p == "/certificate_renewal": return tierAnyClient`) - and returns 403 "access denied" at internal/api/auth.go:119-127 *before* `next.ServeHTTP` is ever called. So the request never reaches `handlePostCertificateRenewal`, and the new `errors.Is(err, ca.ErrCertRevoked)` branches this diff adds (internal/api/handlers.go:937-941 and :979-983) never execute. The test's own comment (lines 1339-1344) claims it guards "the errors.Is(err, ca.ErrCertRevoked) mappings at the HTTP layer, which the CA-layer specs cannot exercise" and defers the cross-replica case to internal/ca/crlsync_test.go - but that file only calls `replica.AutoRenew`/`replica.Renew` directly on a CA object, never through the HTTP mux, so it doesn't exercise the handler's error mapping either. I searched (`grep -rn "certificate_renewal" --include="*_test.go" .`) and the string appears only in this file; no test anywhere sends an HTTP renewal request through a replica whose middleware cache is stale (so the auth check passes it through) while the handler's own `refuseIfRevoked` catches it via a fresh storage read. Consequently, if the `errors.Is(err, ca.ErrCertRevoked)` branch in either handler arm were deleted or broken, both of these tests would still pass (the middleware would already have returned 403), and no test in the changeset would notice.
  </details>

- 🧪 **New "survive a gather failure" contract for crlSyncFailures/crlCachedNumber is untested** ([`internal/metrics/collector.go:218`](https://github.com/voxpupuli/openvox-ca/blob/269254d61a76cbde43d6d932693551c99aa9408d/internal/metrics/collector.go#L218))

  ```
  	// These are in-process tallies of the CA's own state, independent of the
  	// storage gather, so emit them even when the gather below fails — an
  	// unreadable backend must not blind operators to CRL-maintenance failures,
  	// nor to a replica enforcing an out-of-date revocation list. The cached CRL
  	// number in particular is most worth having when storage reads are the thing
  	// going wrong, since that is what makes a replica fall behind.
  	ch <- prometheus.MustNewConstMetric(c.crlUpdateFailures, prometheus.CounterValue,
  		float64(c.ca.CRLUpdateFailures()))
  	ch <- prometheus.MustNewConstMetric(c.crlSyncFailures, prometheus.CounterValue,
  		float64(c.ca.CRLSyncFailures()))
  	if cached, ok := c.ca.CachedCRLNumber(); ok {
  		num, _ := new(big.Float).SetInt(cached).Float64()
  		ch <- prometheus.MustNewConstMetric(c.crlCachedNumber, prometheus.GaugeValue, num)
  	}
  ```

  💡 **Recommendation:** Add a test that forces gather() to error (e.g. replace the "signed" subdirectory with a plain file so os.ReadDir returns a non-ErrNotExist error from FilesystemBackend.List, or inject a storage.Backend stub) and assert puppetca_collector_scrape_success is 0 while puppetca_crl_sync_failures_total and puppetca_crl_cached_number are still present with their pre-failure values.
  **Constraint:** reviewer-protocol.md Engineering Discipline / severity.md HIGH ("missing regression test", "untested code paths in core functionality")

  This diff's central claim - stated twice in comments (here and in the new alert's rationale) - is that crlSyncFailures and crlCachedNumber must keep reporting even when the storage gather fails, because that is precisely when an operator most needs them. None of the four new/modified Its in collector_test.go exercise the `err != nil` branch of Collect() at all: gather() only errors when Storage.ListCerts or Storage.ListCSRs return a non-ErrNotExist error (confirmed in internal/storage/filesystem.go:196-201, which maps a missing directory to an empty list rather than an error), and the "uninitialised" CA test at collector_test.go:199-208 hits that empty-list path, not a genuine gather error - so `puppetca_ca_ready` is still emitted there and the err!=nil early-return path (collector.go:233-236, which skips caReady/caInfo/crlNumber/etc.) is never taken by any test in the file. Searched the whole file (grep -n "scrape_success\|ScrapeSuccess" internal/metrics/*.go) and found no test that forces gather() to fail and then asserts crlSyncFailures/crlCachedNumber are still present with scrape_success=0. The exact behavior this PR is built around - these two series surviving a storage outage - is unverified.
  </details>

</details>

<details><summary>🟡 MEDIUM (9)</summary>

- 🧪 **Pre-write CRL gate's PEM-decode and parse-failure branches have zero test coverage** ([`internal/ca/caImport.go:90`](https://github.com/voxpupuli/openvox-ca/blob/269254d61a76cbde43d6d932693551c99aa9408d/internal/ca/caImport.go#L90))

  ```
  		crlBlock, _ := pem.Decode(crlPEM)
  		if crlBlock == nil {
  			return fmt.Errorf("crl-chain does not contain a valid PEM block")
  		}
  		if _, err := x509.ParseRevocationList(crlBlock.Bytes); err != nil {
  			return fmt.Errorf("failed to parse CRL: %w", err)
  		}
  ```

  💡 **Recommendation:** Add two ImportCA specs alongside the existing ordering-gate tests (in caImport_test.go or crlsync_test.go): one passing a crlPEM with no PEM block (e.g. random non-PEM bytes) and asserting the 'crl-chain does not contain a valid PEM block' error plus that store.GetCACert/GetCRL still return not-found afterward; one passing a PEM block of the wrong type/garbage DER and asserting the 'failed to parse CRL' error with the same no-write check.
  **Constraint:** Coverage completeness / regression protection for the validation gate this PR relocated ahead of all writes

  This PR's entire premise (per the caImport.go comment at lines 81-88 and the commit message for 269254d) is that the --crl-chain validation must run before any write so a rejection leaves storage untouched and the import repeatable. That gate has three failure branches: invalid PEM block, unparseable CRL bytes, and a CRL not led by the imported CA's own CRL. The third (ordering) branch is exhaustively tested in internal/ca/crlsync_test.go:394-495 -- including a dedicated 'writes nothing when it rejects' spec (crlsync_test.go:463-474) and a cross-check that the rejection matches what the server would refuse to boot on (crlsync_test.go:478-495). The other two branches have no counterpart anywhere: `grep -rn "crl-chain does not contain a valid PEM block\|failed to parse CRL" --include="*_test.go" .` from the repo root returns no matches, and internal/ca/caImport_test.go (the file that otherwise covers ImportCA's rejection paths -- cert/key mismatch, non-CA cert) never passes a malformed-PEM or malformed-CRL crlPEM. A future edit to this block (e.g. accidentally validating after a write, or changing the early-return) would not be caught by any existing test, even though the ordering-only branch right next to it is precisely the one this PR hardened against that exact failure mode.
  </details>

- 🧪 **No test verifies ancestor CRLs following the leading CRL survive the import write intact** ([`internal/ca/caImport.go:135`](https://github.com/voxpupuli/openvox-ca/blob/269254d61a76cbde43d6d932693551c99aa9408d/internal/ca/caImport.go#L135))

  ```
  	if crlPEM != nil {
  		// Import-time write: runs before any CRL consumer exists, so it
  		// deliberately skips the crlNotify signal (see signCRLLocked).
  		if err := store.UpdateCRL(ctx, crlPEM); err != nil {
  			return fmt.Errorf("failed to write CRL: %w", err)
  		}
  ```

  💡 **Recommendation:** Extend the 'accepts one that leads with this CA's own CRL' import spec (or add a sibling) to read the CRL back via store.GetCRL after a successful multi-block import and assert it equals the original chain bytes exactly, proving the ancestor CRL was not dropped or re-encoded.
  **Constraint:** Missing contract test for a documented public behavior

  docs/operator-cli.md (this changeset) and the new error message in caImport.go both document that a --crl-chain bundle may carry the CA's own CRL followed by ancestors' CRLs ('ancestors' CRLs may follow it'). ImportCA writes crlPEM through to store.UpdateCRL unchanged, so the contract depends on the full multi-block byte sequence being preserved, not just the leading block that was parsed for validation. No test checks this: caImport_test.go's happy-path spec (lines 46-66) only imports a single-CRL cachedCrlPEM and asserts `Expect(crl).NotTo(BeEmpty())` on read-back (caImport_test.go:53-55), never byte-equality. crlsync_test.go's 'accepts one that leads with this CA's own CRL' spec (crlsync_test.go:446-450) builds a two-CRL chain (`append(cachedCrlPEM, foreignCRLPEM(7)...)`) but only asserts `ImportCA(...).To(Succeed())` -- it never reads the stored CRL back to confirm the ancestor block wasn't dropped or truncated. `grep -n "GetCRL" internal/ca/caImport_test.go internal/ca/crlsync_test.go` shows GetCRL is called in these files but never compared for equality against a multi-block input. A regression that re-encoded or truncated the write to just the parsed leading block would pass every existing test while silently breaking the documented ancestor-preservation contract.
  </details>

- 🧪 **crlSupersedes' unnumbered-CRL ordering branch has zero test coverage** ([`internal/ca/crlsync.go:190`](https://github.com/voxpupuli/openvox-ca/blob/269254d61a76cbde43d6d932693551c99aa9408d/internal/ca/crlsync.go#L190))

  ```
  	if stored.Number == nil || cached.Number == nil {
  		// A CRL without a number cannot be ordered against one that has it.
  		// Prefer the numbered copy; if neither is numbered, keep what we hold.
  		return stored.Number != nil
  	}
  ```

  💡 **Recommendation:** Add a spec (or a direct unit test of crlSupersedes, since it is a pure function) that stores/parses a CRL missing the cRLNumber extension and asserts SyncCRLCache declines to install it over a numbered cache, and a second case asserting a numbered CRL does replace an unnumbered cached one.
  **Constraint:** Missing edge case coverage for boundary-sensitive code (Tester severity calibration: HIGH); AGENTS.md's own convention states 'Keep negative and edge cases first-class: every security-relevant branch ... needs an explicit It.'

  crlSupersedes is the sole gate deciding whether a freshly-read stored CRL replaces the cached one, and its comment explicitly carves out the case where either CRL lacks a CRL Number (RFC 5280 makes the extension optional, so an operator-imported or externally re-signed CRL that passed the signature check in ownStoredCRLLocked can legitimately reach this branch with stored.Number == nil or cached.Number == nil). I searched internal/ca/crlsync_test.go for a direct or indirect exercise of this branch (grep -rn 'crlSupersedes' internal/ca/ shows the only two references are the definition and its single call site in SyncCRLCache) and for any CRL constructed in the test file without a Number field (grep -n 'Number:' internal/ca/crlsync_test.go shows big.NewInt(number) is always supplied in foreignCRLPEM, and Init's bootstrap path also always sets Number via x509.CreateRevocationList, which itself errors on a nil Number per the crypto/x509 source). Every SyncCRLCache/Init test in this changeset exercises only the numbered comparison (equal, greater, lesser). The anti-rollback guarantee this function implements - the one 'never goes backwards to an older CRL' is written to protect - is untested for the one input shape the code's own comment says it must handle differently.
  </details>

- 🧪 **parseFirstCRL's ParseRevocationList error branch is never reached by a test** ([`internal/ca/crlsync.go:153`](https://github.com/voxpupuli/openvox-ca/blob/269254d61a76cbde43d6d932693551c99aa9408d/internal/ca/crlsync.go#L153))

  ```
  	crl, err := x509.ParseRevocationList(block.Bytes)
  	if err != nil {
  		return nil, fmt.Errorf("parsing CRL: %w", err)
  	}
  ```

  💡 **Recommendation:** Add a case that writes a PEM block with Type 'X509 CRL' wrapping garbage bytes (e.g. pem.EncodeToMemory with Bytes: []byte("garbage")) and assert SyncCRLCache/loadCRLCache returns the 'parsing CRL' error and increments the failure counter.
  **Constraint:** Missing edge case coverage for boundary-sensitive code

  parseFirstCRL has two distinct failure branches: pem.Decode returning a nil block (line 150-152) and x509.ParseRevocationList failing on a syntactically-valid PEM block whose DER is not a CRL (line 153-156). Both of the negative crlsync_test.go cases that target this path - 'counts a CRL it cannot read' (UpdateCRL with []byte("not a PEM block")) and the 'not a CRL' blob in 'still enforces a revocation it already knows about when the refresh fails' - are plain non-PEM strings, so both only ever exercise the pem.Decode nil branch (grep -n 'UpdateCRL(ctx, \[\]byte' internal/ca/crlsync_test.go confirms these are the only two malformed-CRL fixtures in the file, and neither is wrapped in a PEM block). The ParseRevocationList error branch - a corrupted or truncated CRL DER inside an otherwise well-formed PEM block, a realistic storage-corruption scenario - has no test reaching it.
  </details>

- 🛡️ **Renewal-path revocation guarantee documented as absolute, omitting the fail-open case on a storage read failure** ([`docs/api.md:74`](https://github.com/voxpupuli/openvox-ca/blob/269254d61a76cbde43d6d932693551c99aa9408d/docs/api.md#L74))

  ```
  That check reads the CRL from storage rather than the in-memory copy the mTLS layer uses, so it holds even on a replica that has not yet picked up a revocation performed elsewhere — renewal cannot be used to escape a lockout during the propagation window.
  ```

  💡 **Recommendation:** Qualify the claim in both docs/api.md and the 'Revocation across replicas' section of docs/configuration.md: note that the storage re-read is best-effort and that a failed refresh falls back to the cached CRL for that one renewal (matching the general propagation-window behaviour, not a stronger guarantee), and reference puppetca_crl_sync_failures_total alongside the existing puppetca_crl_cached_number / puppetca_crl_number guidance so operators can detect the degraded case.
  **Constraint:** Documentation of a security control's guarantee must reflect the actual error-handling behaviour of the code implementing it (OWASP A09 - false assurance from an inaccurate security claim).

  This sentence, and the parallel claim in docs/configuration.md ('Renewal cannot be used to escape a revocation - POST /certificate_renewal re-reads the CRL from storage rather than trusting the cached copy, so a revoked certificate is refused there even on a replica that has not synced'), states unconditionally that the renewal path reads from storage and therefore always catches a revocation performed elsewhere. The implementation (internal/ca/signing.go, refuseIfRevoked, lines 102-122) does call SyncCRLCache first, but on a sync failure it does not fail closed: 'if _, err := c.SyncCRLCache(ctx); err != nil { slog.Warn("Renewal: could not refresh the CRL first; checking against the CRL in memory", ...) }' - then proceeds to check IsRevokedSerial, which (internal/ca/revoke.go:149-161) always reads the stale in-memory c.cachedCRL, never storage directly. The code's own comment on refuseIfRevoked acknowledges this is a real trade-off ('A failed refresh is not itself a refusal. The check below still runs against the CRL already held... a storage blip costs freshness rather than every renewal in the fleet'), i.e. exactly the escape the documentation says cannot happen. Neither docs/api.md nor the new 'Revocation across replicas' section in docs/configuration.md discloses this fallback, and neither points operators at the one signal that would surface it - the puppetca_crl_sync_failures_total counter (internal/ca/ca.go:152-199, internal/metrics/collector.go) - even though the adjacent metrics guidance in configuration.md already covers puppetca_crl_cached_number and puppetca_crl_number for the same purpose. An operator relying on the doc's literal wording would believe a storage blip coinciding with a compromised agent's renewal attempt is impossible, when the code itself documents it as a known, accepted trade-off.
  </details>

- 🧪 **AutoRenew's key-policy-before-revocation ordering is untested for a revoked+weak-key certificate** ([`internal/ca/signing.go:875`](https://github.com/voxpupuli/openvox-ca/blob/269254d61a76cbde43d6d932693551c99aa9408d/internal/ca/signing.go#L875))

  ```
  	if err := validatePublicKey(presentedCert.PublicKey); err != nil {
  		return nil, fmt.Errorf("rejecting auto-renewal for %s: %w", subject, err)
  	}

  	if err := c.refuseIfRevoked(ctx, presentedCert, subject); err != nil {
  		return nil, err
  	}
  ```

  💡 **Recommendation:** Add a case (e.g. alongside "rejects a sub-policy key with ErrKeyPolicy" in internal/ca/renew_test.go's AutoRenew block) that mints a certificate with both a sub-policy key and a revoked serial, and asserts which error AutoRenew actually returns - locking in the intended precedence rather than leaving it as an accidental consequence of statement order.
  **Constraint:** Tester coverage-completeness criterion: edge cases at the interaction of two independently-tested branches need their own case, especially when they change the observable error/status code.

  AutoRenew checks `validatePublicKey` before `refuseIfRevoked`, so a certificate that is both revoked and carries a sub-policy key returns `ErrKeyPolicy` (HTTP 422, "certificate key does not meet policy; renew with a new CSR") rather than `ErrCertRevoked` (HTTP 403, "access denied"). Renew (the CSR/re-key path) orders the opposite way: `refuseIfRevoked` runs first, before any CSR or key validation (internal/ca/signing.go:764-768, checked earlier in this review). Neither ordering is wrong on its own, but the specific combination - a revoked identity whose key also predates policy, hitting the empty-body AutoRenew path - is untested, and the resulting 422 invites that revoked identity toward the CSR-based re-key path with no mention that it is in fact revoked. I searched (`grep -rn "ErrCertRevoked\|ErrKeyPolicy" --include="*_test.go" .`) across internal/ca/renew_test.go, internal/ca/crlsync_test.go, and internal/api/api_test.go: the existing `ErrKeyPolicy` tests (`renew_test.go`'s "rejects a sub-policy key with ErrKeyPolicy", `api_test.go`'s two 422 tests) all use a non-revoked certificate, and the existing `ErrCertRevoked` tests (`crlsync_test.go` lines 256-257, 274-275, 294-295) all use a cert with a policy-compliant key. No test pins down which sentinel wins when both conditions hold, so a future change to either function's check order would go unnoticed even though it changes the status code and message a revoked agent receives.
  </details>

- ⚙️ **New PromQL alert rules have no CI syntax/render validation** ([`mixin/README.md:54`](https://github.com/voxpupuli/openvox-ca/blob/269254d61a76cbde43d6d932693551c99aa9408d/mixin/README.md#L54))

  ```
  jsonnet -S -e "std.manifestYamlDoc((import 'mixin.libsonnet').prometheusAlerts)" \
    > puppet_ca_alerts.yaml
  promtool check rules puppet_ca_alerts.yaml
  ```

  💡 **Recommendation:** Add a CI job (or a step in an existing one, e.g. ci.yml) that renders the mixin with jsonnet and runs `promtool check rules` against the output, so a bad PromQL expression fails the build instead of only being caught by whoever happens to run the README's manual command.
  **Constraint:** Release pipeline integrity: rules and alerting logic that ship without a corresponding automated check are a latent build/runtime gap.

  This PR adds two new alert rules to mixin/alerts.libsonnet, including PuppetCACRLStale's multi-operator PromQL expression (`or` / `unless` / `and on(instance)`), the most structurally complex rule in the file. Rendering and validating the mixin is documented here as a manual, human-run step (`jsonnet` + `promtool check rules`), and I confirmed with `grep -rn "mixin|jsonnet|promtool" .github/workflows/` that none of the four workflow files (ci.yml, codeql.yml, container-images.yml, release.yml) invoke jsonnet or promtool anywhere; only `mixin/README.md` and the Markdown-lint job touch the mixin directory, and that job only lints prose, not the .libsonnet rule files. A malformed selector substitution or an unbalanced `|||` block in alerts.libsonnet or config.libsonnet would compile as valid Jsonnet/YAML in some cases yet fail `promtool check rules` (or silently produce a rule that never fires), and nothing in CI would catch it before merge or release. This gap predates this PR (the ten pre-existing alert rules share it), but this changeset is the one adding the most intricate expression to that unvalidated surface, so it's the right point to flag it.
  </details>

- 🧪 **PuppetCACRLStale's two-armed absent/present logic has no automated test** ([`mixin/alerts.libsonnet:248`](https://github.com/voxpupuli/openvox-ca/blob/269254d61a76cbde43d6d932693551c99aa9408d/mixin/alerts.libsonnet#L248))

  ```
              expr: |||
                puppetca_crl_number{%(selector)s} - puppetca_crl_cached_number{%(selector)s} > 0
                or
                (puppetca_crl_cached_number{%(selector)s}
                   unless puppetca_crl_number{%(selector)s})
                  and on(instance) puppetca_collector_scrape_success{%(selector)s} == 1
              ||| % {
                selector: $._config.puppetCASelector,
              },
  ```

  💡 **Recommendation:** Add a promtool rule test file (`promtool test rules`) covering at minimum: (1) the subtraction arm firing when cached lags stored, (2) the unless arm firing when stored is absent but cached and a successful scrape are present, and (3) the unless arm staying silent when the scrape itself failed (scrape_success != 1) to confirm the qualifier suppresses double-paging with PuppetCAScrapeFailing.
  **Constraint:** severity.md MEDIUM ("Missing property/fuzz/contract tests for untrusted input or public API contract")

  This expression's own comments (alerts.libsonnet:236-247) walk through a subtle absent-metric interaction: the second arm exists specifically to catch the case where puppetca_crl_number goes missing (unreadable stored CRL) while puppetca_crl_cached_number is still published, gated by `and on(instance) ... == 1` to avoid double-paging with PuppetCAScrapeFailing. That is exactly the kind of `absent`/`unless`/`and on()` composition that is easy to get subtly wrong (label-matching edge cases, vector-matching semantics) and that promtool's rule-test format exists to pin down. `find . -iname "*test*.yml" -o -iname "*test*.yaml"` and `grep -rn "promtool test\|rule_files\|test_group"` across the repo turned up nothing - there are no promtool unit tests for any alert in mixin/alerts.libsonnet, so this is consistent with existing project practice rather than a regression this diff introduces, but it is the most logically intricate rule added so far and ships with zero regression protection for either arm.
  </details>

- 🧪 **The core invariant of this PR -- sync runs unconditionally, not gated by disable_crl_refresh -- has no test exercising the actual wiring** ([`cmd/openvox-ca/main.go:691`](https://github.com/voxpupuli/openvox-ca/blob/269254d61a76cbde43d6d932693551c99aa9408d/cmd/openvox-ca/main.go#L691))

  ```
  go runCRLSync(ctx, myCA, cfg.crlSyncInterval())
  ```

  💡 **Recommendation:** Extract the background-job startup decisions (which jobs start, and under what config) into a small testable function -- e.g. a `startBackgroundJobs(ctx, cfg, myCA)` or a table of (condition, starter) pairs returned by a pure function of *serverConfig -- so a unit test can assert that CRL sync is included even when DisableCRLRefresh is true and CRL refresh is excluded, without needing to run the full HTTP server. Absent that refactor, at minimum add an integration-level test that runs newRootCmd with --single-process against a real listener and confirms (e.g. via CachedCRLNumber before/after a peer reissue) that sync happens with disable_crl_refresh set.
  **Constraint:** Tester HIGH calibration: missing regression test for behavior the change explicitly commits to (severity.md Tester/HIGH: "missing regression test"); reviewer-protocol.md Engineering Discipline requires the full contract of introduced logic to be honored and verified, not merely commented.

  This statement is the entire fix for issue #171's core requirement: CRL sync must run on every replica regardless of disable_crl_refresh, because that flag governs re-signing, not propagation. The only thing distinguishing this from the CRLRefresher goroutine three lines above it (which IS gated, inside `if !cfg.DisableCRLRefresh { ... }`) is indentation -- the test file's own comment admits this in as many words (cmd/openvox-ca/config_test.go:295-305: "the serve command keeps [the promise] only by where a line sits ... two adjacent near-identical goroutine starts distinguished by nothing but indentation ... It cannot pin the placement, since the goroutine is started inline in the cobra RunE and nothing in the package can observe it; that half rests on the comment at the call site."). The new test `crlSyncInterval is unaffected by disable_crl_refresh` (config_test.go:306-315) only pins that `cfg.crlSyncInterval()` -- the interval resolver -- doesn't change when DisableCRLRefresh is true; it says nothing about whether the goroutine call itself is reachable when that flag is set. No test in the changeset (or in main_test.go, which only checks cobra.NoArgs) exercises newRootCmd's RunE far enough to observe that runCRLSync actually starts when DisableCRLRefresh=true. A future refactor moving this line inside the adjacent `if` block -- exactly the accident the comment worries about -- would pass every existing test while silently reintroducing the stale-revocation window issue #171 exists to close.
  </details>

</details>

<details><summary>🔵 LOW (9)</summary>

- 📚 **New --crl-chain rejection not reflected in the flag's own --help text** ([`internal/ca/caImport.go:107`](https://github.com/voxpupuli/openvox-ca/blob/269254d61a76cbde43d6d932693551c99aa9408d/internal/ca/caImport.go#L107))

  ```
  return fmt.Errorf("the first CRL in --crl-chain was not signed by the CA certificate being imported; " +
  ```

  💡 **Recommendation:** Update the flag's help string in cmd/openvox-ca-ctl/main.go (e.g. `"Path to CRL PEM (optional; one will be generated if absent). Must lead with the CRL this CA signed."`) so `--help` surfaces the ordering requirement without a docs lookup. No Docs repo is configured for this invoking project (checked /workspace/AGENTS.md and /workspace/CLAUDE.md for a "Review Council Configuration" section with a `Docs repo` entry; none is present -- only README.md's own extension-point documentation shows the placeholder example `myorg/docs`), so this is reported as a finding rather than filed as an issue.
  **Constraint:** Documentation gap: a changed file introduces new user-facing CLI behavior (an import-time rejection) that is documented in prose (docs/operator-cli.md, docs/migrating-from-puppet-server.md) but not in the flag's own --help description, the surface an operator sees without leaving the terminal.

  This changed line adds a new rejection path for `--crl-chain`: ImportCA now fails when the bundle's first CRL was not signed by the CA certificate being imported. docs/operator-cli.md and docs/migrating-from-puppet-server.md were both updated to describe it in prose. However, `cmd/openvox-ca-ctl/main.go:557` still registers the flag as `cmd.Flags().StringVar(&crlChain, "crl-chain", "", "Path to CRL PEM (optional; one will be generated if absent)")` (confirmed by reading that line; unchanged by this diff -- grepped `crl-chain` across the repo, the only hits outside the diff's own files are this StringVar call and its two read/pass-through lines). An operator running `openvox-ca-ctl import --help` -- the fastest path to learn a CLI flag's contract, and the one that needs no doc lookup -- gets no hint that ordering matters, and only discovers the new rejection by triggering it. Since the behavior is otherwise well documented in both markdown files, this is a secondary surface gap rather than an undocumented feature.
  </details>

- 📚 **ImportCA's doc comment omits the new CRL-ordering contract it now enforces** ([`internal/ca/caImport.go:40`](https://github.com/voxpupuli/openvox-ca/blob/269254d61a76cbde43d6d932693551c99aa9408d/internal/ca/caImport.go#L40))

  ```
  // crlPEM may be nil; when nil a fresh empty CRL is generated and written.
  ```

  💡 **Recommendation:** Add a sentence to ImportCA's doc comment alongside the existing crlPEM note, e.g. "When non-nil, crlPEM must parse as a valid CRL whose first entry is signed by the certificate being imported; otherwise ImportCA returns an error before writing anything." No Docs repo is configured for this invoking project (see prior finding), so this is reported as a finding rather than filed as an issue.
  **Constraint:** Inline documentation convention: this same doc comment block already documents crlPEM's nil-case behavior and the supported key formats as part of ImportCA's contract, establishing that parameter behavior belongs in the doc comment -- the new non-nil validation/rejection path breaks that self-consistency by being left out.

  ImportCA's doc comment (lines 33-42) describes what the function does -- "validates the cert/key pair, writes the files, and initialises the serial and inventory files" -- and separately calls out crlPEM's nil-case behavior on line 40. This diff adds a substantial new contract for the non-nil case: the CRL must parse as a valid PEM/X.509 revocation list and its first entry must be signed by the certificate being imported, or ImportCA now returns an error before any write occurs. None of that is reflected in the doc comment above the function, even though the in-body comments added by this same diff (lines 82-88, 97-105) explain the reasoning in detail. A caller reading only the exported doc comment -- the normal way to learn a function's contract in Go -- will not learn that a non-nil crlPEM can fail validation, or why.
  </details>

- 🧪 **Own-CRL rejection assertions are shallower on the sync path than the startup path** ([`internal/ca/crlsync_test.go:187`](https://github.com/voxpupuli/openvox-ca/blob/269254d61a76cbde43d6d932693551c99aa9408d/internal/ca/crlsync_test.go#L187))

  ```
  		updated, err := replica.SyncCRLCache(ctx)
  		Expect(err).To(HaveOccurred())
  		Expect(updated).To(BeFalse())
  ```

  💡 **Recommendation:** Add Expect(err.Error()).To(ContainSubstring("not signed by the CA certificate this process is using")) to the 'refuses a stored CRL this CA did not sign' case, matching the depth already used for the Init path.
  **Constraint:** Assertion Quality - compare assertion depth across tests in the same file to find inconsistencies

  'refuses a stored CRL this CA did not sign' (lines 174-195) and 'counts a CRL it cannot read' (lines 197-206) both check only Expect(err).To(HaveOccurred()) for the SECURITY-tagged refusal path (crlsync.go line 66: '// SECURITY: install only a CRL this CA signed'), never asserting on err.Error(). The sibling Describe block 'Init against the stored CRL' testing the exact same underlying check (ownStoredCRLLocked, shared by loadCRLCache and SyncCRLCache) does the deeper check at lines 396-397 and 409-410: Expect(err.Error()).To(ContainSubstring("not signed by the CA certificate this process is using")). The SyncCRLCache tests verify the same refusal only by side effect (failure counter, cache unchanged) and never confirm the caller actually receives the documented, operator-facing error text on this path - so a change that silently swapped in a generic error here would pass every SyncCRLCache spec in the file.
  </details>

- 🧪 **CA-seeding sequence duplicated instead of reusing the seedCA helper** ([`internal/ca/crlsync_test.go:485`](https://github.com/voxpupuli/openvox-ca/blob/269254d61a76cbde43d6d932693551c99aa9408d/internal/ca/crlsync_test.go#L485))

  ```
  		store := storage.New(dir)
  		Expect(store.EnsureDirs(ctx)).To(Succeed())
  		Expect(store.SaveCAKey(ctx, cachedKeyPEM)).To(Succeed())
  		Expect(store.SaveCACert(ctx, cachedCrtPEM)).To(Succeed())
  		Expect(store.UpdateCRL(ctx, chain)).To(Succeed())
  		Expect(store.WriteSerial(ctx, "0001")).To(Succeed())
  		Expect(store.TouchInventory(ctx)).To(Succeed())
  ```

  💡 **Recommendation:** Hoist seedCA (or the seeding sequence it wraps) to a file-scope helper function so both Describe blocks share one definition.
  **Constraint:** Test Architecture - self-contained fixtures without unnecessary duplication

  'rejects exactly what the server would refuse to start on' (lines 478-495) re-implements, line for line, the same seven-call storage-seeding sequence already factored out as the seedCA closure at lines 375-392 in the sibling 'Init against the stored CRL' Describe block. seedCA is a local closure scoped to that Describe block, so it can't be called from 'ImportCA with a --crl-chain bundle' without hoisting it to file scope, which is why the duplicate exists.
  </details>

- 📚 **Cross-replica revocation propagation fix has no tracked blog issue** ([`README.md:44`](https://github.com/voxpupuli/openvox-ca/blob/269254d61a76cbde43d6d932693551c99aa9408d/README.md#L44))

  ```
  - **Revocation that propagates:** on the HA backends every replica reloads the stored CRL on a short timer (`crl_sync_interval_sec`, 60s by default), so a certificate revoked on one replica stops being accepted by the others within that window rather than whenever they next happen to re-sign. Renewal re-reads the CRL from storage, so a revoked certificate cannot renew itself into a fresh one on a replica that has yet to catch up. See [Revocation across replicas](docs/configuration.md#revocation-across-replicas)
  ```

  💡 **Recommendation:** Once a Docs repo is configured for this project, check for an existing `blog`-labeled issue covering the cross-replica revocation propagation fix (e.g. `gh issue list --repo <DOCS_REPO> --label blog --search "revocation propagation" --state open`) and, if none exists, file one, e.g.: `gh issue create --repo <DOCS_REPO> --title "blog: cross-replica revocation propagation fix" --label blog --body "openvox-ca previously left revoked certificates valid against HA replicas (etcd/redis/postgres/mysql) that had not performed the revocation, for up to the CRL refresh window (issue #171). This PR adds crl_sync_interval_sec (60s default) to close that gap; see README.md#Features and docs/configuration.md#revocation-across-replicas. Worth a blog post walking operators through the exposure and the fix."` Until a Docs repo is configured, this is a recommendation only, not an actionable issue-filing command.
  **Constraint:** Blog Opportunity Identification (curator persona contract, criterion 4): a new major capability warrants a blog issue when none exists.

  This changeset documents a new major capability: README.md elevates it to a headline Feature-list bullet (line 44), and docs/configuration.md adds a ~70-line "Revocation across replicas" section, and docs/api.md and docs/storage-backends.md each add a paragraph cross-referencing it. The underlying behavior closes a real operational/security gap tracked in linked issue #171 - on the `etcd`/`redis`/`postgres`/`mysql` backends a revoked certificate previously kept working against replicas that did not perform the revocation for as long as the CRL refresh window (documented elsewhere as on the order of weeks at default settings), including for the renewal endpoint. A fix this significant, shipped with this much operator-facing documentation, is a natural candidate for a blog post explaining the gap and the fix so operators running HA backends know to check their `crl_sync_interval_sec` exposure. No Docs repo was configured for this review run (the delegation brief's Review Council Configuration contains no "Docs repo" entry), so per the Bash Access Restriction and Graceful Degradation rules I did not run any forge issue-list query and cannot confirm whether a blog issue already exists anywhere.
  </details>

- 📚 **Stale signing.go line reference after refuseIfRevoked insertion** ([`internal/ca/renew_test.go:121`](https://github.com/voxpupuli/openvox-ca/blob/269254d61a76cbde43d6d932693551c99aa9408d/internal/ca/renew_test.go#L121))

  ```
  		// CN == subject as defence-in-depth (signing.go:647) and must reject.
  ```

  💡 **Recommendation:** Update the reference to `signing.go:782` (or drop the line number and describe the check by name, which would not require maintenance on every unrelated edit above it).
  **Constraint:** Cross-reference consistency (documentation references must point to correct, existing targets)

  This comment, unchanged by the diff, cites `signing.go:647` for the CN-match defence-in-depth check inside `Renew`. The same diff inserts the ~53-line `ErrCertRevoked`/`refuseIfRevoked` block earlier in signing.go (new lines 70-121) plus a 4-line `refuseIfRevoked` call inside `Renew` itself, shifting every later line down. I confirmed by reading internal/ca/signing.go directly: the CN check inside `Renew` (`// SECURITY: CN must match subject - defence-in-depth; handler also checks.`) is now at line 782, not 647. The comment in this test file was not updated to track the shift its own diff caused.
  </details>

- 📚 **Stale signing.go line reference for proof-of-possession check** ([`internal/ca/renew_test.go:145`](https://github.com/voxpupuli/openvox-ca/blob/269254d61a76cbde43d6d932693551c99aa9408d/internal/ca/renew_test.go#L145))

  ```
  		// Renew verifies the CSR proof-of-possession signature (signing.go:642)
  ```

  💡 **Recommendation:** Update the reference to `signing.go:777`.
  **Constraint:** Cross-reference consistency (documentation references must point to correct, existing targets)

  Same root cause as the CN-check reference at line 121: this comment cites `signing.go:642` for the proof-of-possession signature check inside `Renew`, but the diff's insertion of the `ErrCertRevoked`/`refuseIfRevoked` block and the new `refuseIfRevoked` call inside `Renew` pushed that check down. Reading internal/ca/signing.go directly, the `// SECURITY: Verify the CSR's proof-of-possession signature.` line inside `Renew` is now at line 777, not 642.
  </details>

- 📚 **Comment points to a "SECURITY note" that refuseIfRevoked's godoc never labels as such** ([`internal/api/handlers.go:929`](https://github.com/voxpupuli/openvox-ca/blob/269254d61a76cbde43d6d932693551c99aa9408d/internal/api/handlers.go#L929))

  ```
  		// see the SECURITY note on ca.refuseIfRevoked, which both renewal paths
  		// go through.
  ```

  💡 **Recommendation:** Either add a `// SECURITY:` line to refuseIfRevoked's godoc in internal/ca/signing.go (matching the convention used everywhere else, e.g. signing.go:777 `// SECURITY: Verify the CSR's proof-of-possession signature.`), or reword the handlers.go comment to stop promising a labeled section that doesn't exist.
  **Constraint:** Inline documentation format convention: security rationale is consistently marked with a `// SECURITY:`-prefixed line (established pattern, 37 occurrences across cmd/ and internal/, e.g. internal/api/auth.go:67, internal/ca/signing.go:299/308/332/662/671/777/782)

  This new comment tells the reader to go read "the SECURITY note on ca.refuseIfRevoked" for the rationale behind the read-through revocation check. I read refuseIfRevoked's doc comment in internal/ca/signing.go (lines 76-101, added by this same diff): it is a substantial, well-written rationale, but no line in it is prefixed `SECURITY:` - it ends instead with a bare `NIST 800-53: IA-5(2)...` citation line, which every other security-relevant function in the codebase pairs with a `SECURITY:` line, not in place of one. A reader following this pointer to scan for the labeled note this codebase's convention leads them to expect will not find one under that name on this newly introduced, security-critical function.
  </details>

- 🧪 **Documented "no pass before first tick" behavior of runCRLSync has no test that can distinguish it from an immediate-pass implementation** ([`cmd/openvox-ca/crl_sync.go:52`](https://github.com/voxpupuli/openvox-ca/blob/269254d61a76cbde43d6d932693551c99aa9408d/cmd/openvox-ca/crl_sync.go#L52))

  ```
  // Unlike the other background jobs this one does not run a pass before its
  // first tick: CA.Init has just loaded the CRL from the same storage, so an
  // immediate pass would re-read a blob that cannot yet be stale.
  ```

  💡 **Recommendation:** Add a test that revokes/reissues on the primary CA *before* calling runCRLSync, then starts runCRLSync with an interval long enough to observe a window (e.g. 500ms) during which the peer's cached CRL number must still equal the pre-sync value, and only advances after that window elapses -- proving the update came from a tick, not an immediate pass.
  **Constraint:** Tester coverage-completeness criterion (regression anchors): a behavior the spec/comment calls out as deliberate and distinguishing from a sibling job should be locked down by a test that fails if the behavior regresses.

  The comment states runCRLSync deliberately omits the startup pass that its sibling runCRLRefresher performs (crl_refresh_test.go:87-109 explicitly titles and tests that job's startup pass: "must perform an immediate check at startup"). No equivalent test exists for the negative claim about runCRLSync. The one test that exercises runCRLSync (crl_sync_test.go:236-264, "picks up a CRL written after it started, and returns on cancellation") calls c.ReissueCRL only after runCRLSync's goroutine has already started, then polls for up to 2s with a 10ms ticker interval -- so it cannot tell an implementation that ticks-only apart from one that also runs an immediate pass at start: both would show the peer's cache updated well within the 2s window. If a future change added an immediate first pass (e.g. by copying the pattern from runCRLRefresher), this test suite would not notice, even though the code comment calls that specific difference out as intentional. CORRECTION (independent validator): The runCRLSync spec is at cmd/openvox-ca/crl_sync_test.go:86-113 (It at line 87), not 236-264; that file is only 114 lines long.
  </details>

</details>

---

bootc:
Addressed in 13adaf9 and 924110b. All 3 HIGH and all 9 MEDIUM are resolved; 8 of 9 LOW are fixed and one is declined. One HIGH I don't think is correct — evidence below.

## The re-sign finding was the important one, and it was right

Verifying only the paths that read the CRL *into cache* was worse than not verifying at all. `revoke`, `reissue`, `refresh` and `cleanup` don't merely read the stored list: they take its number and entries, sign them with the CA's real key, and write the result back. So a foreign CRL that `SyncCRLCache` had correctly refused would be laundered by the next revocation into one every subsequent check accepts — and for an empty entry list, that means quietly unrevoking the fleet while appearing to add a revocation.

`readStoredCRL` now goes through the same `ownStoredCRLLocked` check, so all four callers fail closed. Two specs assert a poisoned store is left byte-identical rather than re-signed over.

The validator's correction to that finding (four callers, not two) was also right and is what made the fix worth doing centrally rather than at the two sites originally named.

## One finding I dispute

**"New revoked-cert renewal tests never reach the code they claim to guard."** The analysis assumes the auth middleware runs in that suite and 403s before the handler. It doesn't: `internal/api/api_test.go` never sets `server.AuthConfig`, and `newAuthMiddleware` returns `next` unchanged when `cfg == nil`, so the middleware is a pass-through there and the request does reach `handlePostCertificateRenewal`.

Verified rather than argued — I disabled both `errors.Is(err, ca.ErrCertRevoked)` branches and re-ran: both specs fail. They are exercising the handler.

That said, the test you asked for is a better test, so I added it: `internal/api/auth_test.go` now drives a renewal through a peer replica where `AuthConfig` *is* set, whose middleware admits the certificate on its stale cache, and asserts the handler's read-through check refuses it 403 on both branches. That's the arrangement the check actually exists for, and nothing covered it before.

## Behaviour changed, not just tests

- **`AutoRenew` now checks revocation before key strength**, matching `Renew`. Both refuse, so the order only decided the answer — and 422 "renew with a new CSR" was the wrong one, pointing a revoked identity at the re-key path without mentioning the revocation.
- **The renewal guarantee is no longer documented as absolute.** `docs/api.md` and `docs/configuration.md` now say the storage re-read is best-effort and falls back to the cached CRL on a failed read, and name `puppetca_crl_sync_failures_total` as the signal for it.

## Coverage and tooling

- `mage test:mixin` renders the mixin and runs `promtool check rules` + `promtool test rules`, with a CI job to match. `mixin/tests.yaml` covers both arms of `PuppetCACRLStale` including the scrape-success qualifier that stops a storage outage double-paging.
- `backgroundJobs` extracted from the serve command, so "CRL sync runs regardless of `disable_crl_refresh`" is a proposition a spec checks rather than a property of where a line sits.
- Import gate's malformed-PEM and unparseable-CRL branches; multi-block bundle preserved byte for byte; `crlSupersedes`' unnumbered ordering (white-box, because `x509.CreateRevocationList` refuses a nil `Number`, so it cannot be reached through storage); gather-failure metrics; `runCRLSync`'s absent startup pass.
- LOW: `--crl-chain` help text, `ImportCA` doc comment, sync-path assertion depth, `seedCA` hoisted, both stale `signing.go:NNN` references (replaced with names rather than new line numbers), `SECURITY:` label on `refuseIfRevoked`.

## Declined

**Blog issue** — no Docs repo is configured, and whether this warrants a post is a maintainer call rather than something to file speculatively.

## Two things the fixes themselves got wrong

Re-reviewing the above turned up two more, both now fixed in 924110b:

1. The new `mixin` CI job wasn't in the `ci` aggregate's `needs:`, so `alls-green` never saw it and branch protection would have gone green over a red mixin — a check that can't fail the build, while the README I added claimed the opposite.
2. Making `readStoredCRL` fail closed left reissue/refresh/cleanup logging without counting: only `revoke` incremented `crl_update_failures_total`, which `docs/metrics.md` says covers all four and the mixin alerts on. A replica whose stored CRL stopped verifying would have retried silently until its CRL lapsed. The counter now lives in `readStoredCRL`, and `revoke` drops its own so it isn't double-counted.

On the brittle-annotation point: I tried dropping `exp_annotations` from the promtool cases and the suite fails — promtool compares annotations whenever a case expects an alert, and omitting the field asserts the alert carries none. The coupling is promtool's rather than a choice, so it stays, with a comment in `mixin/tests.yaml` saying so.

CI is green (30/30).

Issue #197, opened 2026-08-07
OCSP response signing holds the CA's global write lock


bootc:
### Problem

`OCSPResponse` (`internal/ca/ocsp.go`) answers cache hits under `c.mu.RLock`, but its slow path — status lookup, `ocsp.CreateResponse`, cache write — runs under an exclusive `c.mu.Lock`. Signing an OCSP response is not cheap: it is an RSA/ECDSA signature, and under key isolation (`ExternalSigner`) it is a synchronous IPC round trip to the signer process. All of that happens while holding the CA's global write lock.

Two consequences:

- Every uncached lookup serialises the whole CA behind it: concurrent OCSP requests queue, and so do issuance paths and the `c.mu.RLock` readers on the hot authentication path (`IsRevokedSerial`).
- Nonced requests (RFC 8954 — the default in several client stacks) always bypass the cache, so a client sending nonces takes the exclusive path on every request. A modest stream of nonced OCSP traffic effectively serialises the CA, which sits awkwardly with OCSP being a nominally read-only endpoint; `docs/development/locking.md` records the design intent that read paths stay cheap and lock-free.

This is an efficiency/robustness observation, not a correctness bug: the produced responses are correct.

### Suggested direction

Restructure the slow path so the signature happens outside the lock:

1. Under `c.mu.RLock`, snapshot what the decision needs (serial-known?, revocation entry from `cachedCRL`, `CACert`/`CAKey` references — both effectively immutable after `Init`).
2. Release the lock, build and sign the response.
3. Re-acquire briefly (write) to populate `ocspCache`, tolerating the benign race where two goroutines both sign the same serial and one result wins.

The revocation status may be a few microseconds stale relative to a concurrent revoke, but that is already true the moment the response is signed (responses assert validity for `OCSPValidity` anyway), and the existing revoke path invalidates the cache entry afterwards.

### Cross-references

Found while auditing lock usage for `docs/development/locking.md`. Distinct from #183, which is about `serialIndex`/`ocspCache` never being resynchronised across replicas; this issue is purely about lock scope on the response-signing path.

Issue #196, opened 2026-08-07
DELETE /certificate_request deletes the CSR without taking the subject lock


bootc:
### Problem

`DELETE /certificate_request/{subject}` (`handleDeleteRequest` in `internal/api/handlers.go`) calls `Storage.DeleteCSR` directly, bypassing the CA layer and therefore the `subject:<name>` cluster lock. Every other mutation of per-subject state goes through a CA method that takes that lock.

### Failure modes

Low severity — the delete itself is a single atomic operation, so nothing corrupts — but the outcome can silently contradict the caller's intent:

- A concurrent `Sign`/autosign for the same subject reads the CSR under the subject lock; the unlocked delete lands after the read, the sign completes anyway, and a certificate is issued for a request the operator believes they rejected. The 204 response tells them the request is gone.
- The delete lands inside `SaveRequest`'s locked section, between `SaveCSR` and the autosign `sign` call: signing fails with "CSR not found" and the agent's submission turns into a 500 instead of either a clean save or a clean rejection.

### Suggested fix

Route the handler through a small CA method (e.g. `CA.DeleteRequest`) that wraps the delete in `Storage.WithLock(ctx, subjectLockName(subject), ...)`, preserving the current 404-on-absent behaviour. That makes request deletion linearise with signing: it either lands before the sign (which then fails cleanly with no CSR) or after it (the cert exists and deleting the CSR is the normal post-sign cleanup).

### Cross-references

Found while auditing lock usage for `docs/development/locking.md`. The lock-name table and ordering invariant the new method should follow are documented there.

Issue #195, opened 2026-08-07
Generate bypasses the per-subject cluster lock, allowing cross-replica double issuance


bootc:
### Problem

`CA.Generate` (`internal/ca/generate.go`, reached from `POST /generate/{subject}` in `internal/api/handlers.go`) is the only issuance path that does not take the cluster-wide per-subject lock. It serialises the evict + save-CSR + sign sequence on `c.mu` alone, whereas every sibling path — `SaveRequest`, `Sign`, `SignWithTTL`, `Renew`, `AutoRenew`, `Clean`, `ImportCertificate` — wraps the equivalent sequence in `Storage.WithLock(subjectLockName(subject))`.

`c.mu` is process-local, so on an HA backend (etcd, Redis, PostgreSQL, MySQL) nothing coordinates a `Generate` on one replica with a `Generate`, `Sign`, or `SaveRequest` for the same subject on another. Both replicas can pass `evictRevokedLocked` (no cert on disk yet), and both proceed to issue.

### Failure modes

- Two `Generate` calls race: both mint certificates (different serials, different private keys) and both append inventory rows — the unique-serial constraint does not help because the serials differ. The last `SaveCert` wins, so at least one caller walks away holding a private key that does not match the certificate now served for that subject.
- `Generate` races a `SaveRequest`/`Sign` from the agent flow: two valid certificates exist for one subject, and which one is on disk depends on write ordering.

Within a single process `c.mu` serialises everything, so filesystem/SQLite deployments are unaffected (modulo the separate same-host gap in #187). `Generate` also never applies `lockTimeout`, unlike the other paths.

### Suggested fix

Wrap the storage-touching tail of `Generate` (evict + `SaveCSR` + `sign` + `SavePrivateKey`) in `Storage.WithLock(ctx, subjectLockName(subject), ...)` with the usual `context.WithTimeout(ctx, lockTimeout)`, exactly as `SaveRequest` does. Key generation and CSR assembly already run before any lock is taken and should stay there. Wants a spec along the lines of `renewrace_test.go` on #186: a second `StorageService` on the same directory to bypass the process-local lock map, proving two concurrent `Generate` calls cannot both issue.

### Cross-references

[...]

Issue #187, opened 2026-08-02
Add same-host locking to the filesystem and SQLite backends


bootc:
`StorageService.WithLock` coordinates across processes only when the backend
implements `Locker`. `FilesystemBackend` implements no `AcquireLock` at all, and
`SQLBackend` returns `ErrDistributedLockingUnsupported` for SQLite
(`internal/storage/sql.go`, the `default:` arm — "SQLite is single-node"). Both
therefore fall through to `s.localNamedLock(name)`, a process-local mutex.

That is the right call for clustering, and this issue does not propose changing
it: `docs/storage-backends.md` scopes both backends to "single-node installs"
and its comparison table gives them `Cross-node consistency: single node`.
Shared storage across hosts stays unsupported and out of scope — `flock(2)` over
NFS would not be a sound basis for it anyway.

What is not covered is **two processes on the same host**, which is a real and
increasingly likely shape:

- `openvox-ca-ctl setup` / `import` / `migrate` already operate directly on
  storage, #160 adds `openvox-ca csr` / `import-ca-cert`, and #175 proposes
  `openvox-ca generate`, which *issues certificates*. An operator running that
  beside a live server on the default backend gets two writers with no
  coordination at all.
- The failure is neither theoretical nor self-limiting. Two issuers can each
  pass `evictRevokedLocked` and both write, producing two certificates for one
  subject. Worse, `AppendInventory`'s blob path recomputes a whole-blob HMAC
  from a pre-read snapshot, so a concurrent append leaves an HMAC covering a
  blob that never existed. `InitHMAC` then fails at startup and **the server
  will not restart**, with `RebuildInventoryHMAC` reachable only from
  `internal/storage/migrate.go`.

A same-host lock turns that from silent corruption into a clean "another process
holds the CA lock" error.

## Why `flock(2)` rather than a lockfile

It is released automatically when the descriptor closes or the process dies, so
there is no stale-lock cleanup and no PID-liveness guessing — the failure mode
that makes `O_EXCL` lockfiles a maintenance burden.

## Sketch — filesystem backend

`AcquireLock` opens (creating) a file under `<baseDir>/locks/`, takes `LOCK_EX`,
and returns an `Unlocker` that closes the descriptor. Three details that matter:

- **Encode the lock name.** The names in use are `bootstrap`, `crl` and
  `subject:<name>` (`internal/ca/init.go`). The colon wants encoding, and
  hashing the name avoids ever having to think about subject characters again.
[...]

PR #198, opened 2026-08-07
Document the locking model and rules for working within it

bootc:
Locking spans three tiers — cluster-wide named locks taken through `StorageService.WithLock`, the storage service's process-local RW mutexes, and the CA's in-memory `c.mu` — but until now the design lived only in scattered code comments, so every change touching shared state had to rediscover the lock names, the ordering invariant, and the read-path policy from first principles.

This adds `docs/development/locking.md` as the authoritative reference:

- the three locking tiers and what class of interleaving each protects against;
- the lock-name table (`bootstrap`, `crl`, `subject:<name>`), which operations hold each lock, and the per-backend `Locker` mechanics (etcd lease, Redis `SET NX PX` + heartbeat, Postgres advisory locks, MySQL `GET_LOCK`, process-local fallback);
- the global acquisition order (`subject:<name>` → `crl` → `c.mu`), including `Init`'s deliberate pre-serving inversion;
- the read-path policy: read-only operations never take a distributed lock — in-memory caches and read locks only — and the deliberate costs of that choice;
- nine rules for new or changed code (check-inside-the-lock, lock ordering, lease caveats, SQL pool sizing, lock names as cross-replica protocol);
- a known-gaps section cross-referencing the open locking issues (#171, #173, #183, #187, #195, #196, #197 and PRs #182/#186) so they are not re-reported.

`AGENTS.md` gains a three-line summary pointing at the document, the development README indexes it, and `storage-internals.md`'s three cross-node coordination sections now defer to the new document's lock table instead of carrying a stale copy (its etcd table predated `ReissueCRL`, `Clean`, `Renew` et al).

The audit that produced this document also filed #195 (`Generate` bypasses the per-subject lock), #196 (`DELETE /certificate_request` bypasses it too), and #197 (OCSP signs under the global write lock); those are tracked separately and recorded in the known-gaps section.

The document went through four rounds of review-council: eight documentation-accuracy findings were fixed across the first three rounds (notably scoping rule 6's `ErrDuplicateSerial` example to match the code's own cross-replica disclaimer), and the final round returned a unanimous APPROVE with zero findings.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -139,6 +139,20 @@ The build tags in use are `etcd_integration`, `redis_integration`,
 must be reachable from a `magefile.go` `Test.Backends*` target so it runs in CI;
 a build-tagged suite wired to no target is dead code.
 
+## Locking and concurrency
+
+Any change that reads or mutates shared CA state (certificates, CSRs, the CRL,
+the inventory, in-memory caches) must follow
+[docs/development/locking.md](docs/development/locking.md). The short form:
+
+- Mutations serialise on cluster-wide named locks via `StorageService.WithLock`
+  (`bootstrap`, `crl`, `subject:<name>`); the check that justifies a mutation
+  must run inside the same lock as the mutation.
+- Read-only paths must **not** take `WithLock` — they use in-memory caches and
+  read locks only.
+- Lock ordering is `subject:<name>` → `crl` → `c.mu`; lock names are a stable
+  cross-replica protocol — never invent or rename one casually.
+
 ## Compatibility contracts (do not rename)
 
 openvox-ca is a drop-in for the OpenVox/Puppet Server CA. The following
--- a/docs/development/README.md
+++ b/docs/development/README.md
@@ -15,5 +15,6 @@ style).
 | --- | --- |
 | [Testing](testing.md) | Unit, integration, compose, and load test suites and how to run them |
 | [Storage internals](storage-internals.md) | Backend key layouts, cross-node coordination, SQL schema, and how to add a backend |
+| [Locking and concurrency](locking.md) | The three locking tiers, lock names and ordering, read-path policy, and rules for new code |
 | [Inventory store](inventory-store.md) | Internal design of the structured (`InventoryStore`) inventory capability |
 | [Publishing container images](publishing-images.md) | The GHCR publishing workflow and one-time repository setup |
--- a/docs/development/locking.md
+++ b/docs/development/locking.md
@@ -32,7 +32,10 @@ backends (filesystem, SQLite).
 
 The lock names are part of the cross-replica protocol: every replica must agree
 on them, so they are **stable across releases**. They are defined in
-[init.go](../../internal/ca/init.go):
+[init.go](../../internal/ca/init.go) — with one mirror: `migrateLockName` in
+[migrate.go](../../internal/storage/migrate.go) redefines `"bootstrap"`
+independently (the `internal/storage` package cannot import `internal/ca`), so
+the two are coupled only by the string literal and must be renamed together.
 
 | Lock name | Serialises | Taken by |
 | --- | --- | --- |
@@ -40,16 +43,31 @@ on them, so they are **stable across releases**. They are defined in
 | `crl` | Every CRL read-modify-write (read entries → re-sign → write) | `Revoke`, `ReissueCRL`, `RefreshCRLIfDue`, `CleanupExpiredCerts`, and the revoke step inside `Clean`, `Renew`, `AutoRenew` |
 | `subject:<name>` | The whole lifecycle of one subject: evict/save CSR/sign/import/clean | `SaveRequest`, `Sign`, `SignWithTTL`, `Renew`, `AutoRenew`, `Clean`, `ImportCertificate` — but currently **not** `Generate` (see known gaps below) |
 
-How each backend provides the distributed lock:
+How each backend provides the distributed lock (a summary — the full per-backend
+mechanism, key layouts and transaction/retry detail lives in
+[storage internals](storage-internals.md), which owns it):
 
 | Backend | Mechanism | Crash recovery |
 | --- | --- | --- |
-| etcd | `concurrency.Mutex` under `<prefix>/locks/<name>` on a lease-backed session (30 s TTL) | Lease expires, lock releases |
-| Redis/Valkey | `SET NX PX` with a per-acquisition random token; a heartbeat goroutine extends the TTL while held; unlock is a Lua compare-token-and-delete | TTL elapses, lock releases |
-| PostgreSQL | `pg_advisory_lock` (session-level) on a dedicated pooled connection | Session ends, lock releases |
-| MySQL/MariaDB | `GET_LOCK` on a dedicated connection, polled with a 1 s server-side wait so context cancellation is honoured | Session ends, lock releases |
+| etcd | `concurrency.Mutex` under `<prefix>/locks/<name>` on a lease-backed session (30 s TTL) | Lease expires within the TTL, lock releases |
+| Redis/Valkey | `SET NX PX` with a per-acquisition random token; a heartbeat goroutine extends the TTL while held; unlock is a Lua compare-token-and-delete | Key expires within the TTL, lock releases |
+| PostgreSQL | `pg_advisory_lock` (session-level) on a dedicated pooled connection | Only when the server reaps the session — no TTL (see below) |
+| MySQL/MariaDB | `GET_LOCK` on a dedicated connection, polled with a 1 s server-side wait so context cancellation is honoured | Only when the server reaps the session — no TTL (see below) |
 | SQLite, filesystem | None — `ErrDistributedLockingUnsupported` / no `Locker`; `WithLock` falls back to a process-local mutex | n/a (single process assumed; see [#187](https://github.com/voxpupuli/openvox-ca/issues/187)) |
 | Overlay | Delegates to the base backend's `Locker`; reports unsupported when the base has none | as base |
+
+Crash recovery is not uniform across those backends. etcd and Redis self-heal
+within the lock TTL (30 s): a crashed holder's lease or key expires and the lock
+frees itself, which is what the 60 s `lockTimeout` below is sized to ride out.
+The SQL advisory locks have **no TTL** — `pg_advisory_lock` and `GET_LOCK`
+persist until the database tears the holder's session down, and after a hard
+host loss or network partition that is governed by the server's own keepalive
+settings (PostgreSQL `tcp_keepalives_idle`, MySQL `wait_timeout`), whose
+defaults are measured in hours. A crashed replica can therefore hold `crl` or
+`subject:<name>` far beyond `lockTimeout` while every surviving replica's
+revoke/sign fails at that timeout; the recovery action is to terminate the
+orphaned backend session (`pg_terminate_backend` / `KILL`) or to lower those
+server-side keepalives for HA deployments.
 
 Every distributed implementation first takes a **per-name process-local mutex**
 before touching the network. This is load-bearing, not an optimisation: etcd's
@@ -57,16 +75,18 @@ before touching the network. This is load-bearing, not an optimisation: etcd's
 backends it stops N in-process callers each pinning a pooled connection just to
 queue on the same lock.
 
-Callers bound lock acquisition *and* the critical section together with
-`lockTimeout` (60 s, [init.go](../../internal/ca/init.go)) via
+CA request paths bound lock acquisition *and* the critical section together
+with `lockTimeout` (60 s, [init.go](../../internal/ca/init.go)) via
 `context.WithTimeout` — long enough to ride out a brief leader election, short
-enough that a crashed replica's stale lease doesn't hang requests forever.
-Caveat: that timeout bounds only the *distributed* half. The process-local
-mutexes (both the fallback path and the per-name gate in front of every
-distributed implementation) are plain `sync.Mutex` and do not honour context
-cancellation, so same-process waiters queue unboundedly. The in-flight
-[PR #186](https://github.com/voxpupuli/openvox-ca/pull/186) adds this caveat
-to `StorageService.WithLock`'s godoc.
+enough that a crashed replica's stale lease doesn't hang requests forever. Two
+things that timeout does *not* cover. First, it bounds only the *distributed*
+half: the process-local mutexes (both the fallback path and the per-name gate
+in front of every distributed implementation) are plain `sync.Mutex` and do not
+honour context cancellation, so same-process waiters queue unboundedly. Second,
+the offline `openvox-ca-ctl migrate` path applies no timeout at all —
+`MigrateService` inherits the caller's context, which is signal-cancellable but
+carries no deadline — so a migration waits indefinitely on a contended
+`bootstrap` lock; interrupt it and stop the server rather than waiting.
 
 ### Tier 2: storage-service mutexes
 
@@ -82,6 +102,12 @@ has no read-only fast path:
 | `crlMu` | `crl` | Plain read/write pairs |
 | `fileMu` | `ca_cert`, `ca_pubkey`, `ca_key`, `csr/<subject>`, `cert/<subject>`, per-subject private keys | One mutex spans all subjects; simple and sufficient at current scale |
 | `serialMu` | `serial` | Plain read/write pairs |
+
+One shared-state key is deliberately absent from this table: `hmac_key`. Its
+initialisation (`EnsureHMACKey`, reached via `InitHMAC` *before* any lock is
+taken) is an unlocked read-modify-write, guarded by neither a mutex nor a
+cluster lock today — a genuine gap, tracked as
+[#202](https://github.com/voxpupuli/openvox-ca/issues/202) (see known gaps).
 
 These are **internal to `StorageService`** — callers never touch them, and no
 `StorageService` method calls another locked method while holding one (they are
@@ -101,6 +127,20 @@ without a storage round-trip).
 Mutating operations hold `c.mu` (write) across the storage mutation *and* the
 cache update, so within a process the caches can never be observed out of step
 with what the same process just wrote. Read paths take `c.mu.RLock` only.
+
+`c.mu` is also held across the signing call itself. Every issuance path (`Sign`,
+`SignWithTTL`, `SaveRequest`'s autosign, `Renew`, `AutoRenew`,
+`ImportCertificate`, `Generate`) calls `issueLeafLocked` with `c.mu` held, and
+`x509.CreateCertificate` runs inside it — so with an external key provider
+(`ca_key_provider: openbao`, or the isolated signer) `c.mu`, not the per-subject
+cluster lock, is the process-wide issuance serialiser, and it spans a
+network/IPC round trip. Issuance therefore proceeds at roughly one signing
+round trip at a time within a process, and a stalled signer backend pins the
+mutex and stalls all issuance — see the "Performance and outage behaviour"
+section of [the OpenBao Transit guide](../openbao-transit.md). This is the one
+deliberate exception to rule 3 (keep expensive work outside the lock): the
+signature is inside the lock because the cache update it guards must be atomic
+with the issuance.
 
 `c.mu` is non-reentrant. The same `...Locked` suffix convention applies: e.g.
 `revokeLocked` requires the cluster `crl` lock **and** `c.mu`; each `...Locked`
@@ -125,9 +165,12 @@ subject:<name>  →  crl  →  c.mu  →  (StorageService internal mutexes)
 - Two *different* subject locks are never held at once (bulk operations like
   `SignMultiple` loop, taking one at a time).
 - `CA.Init` inverts the order (it holds `c.mu` while acquiring `bootstrap`).
-  That is safe only because `Init` runs to completion before the server starts
-  serving, so nothing else can be holding a distributed lock while waiting on
-  `c.mu`. Do not copy this pattern into anything that runs while serving.
+  The inversion itself is safe only because `Init` runs to completion before
+  the server starts serving, so nothing else can be holding a distributed lock
+  while waiting on `c.mu`; do not copy this pattern into anything that runs
+  while serving. Init also has a *separate*, unfixed hazard on the same lock —
+  its slow path can re-enter `bootstrap` and deadlock startup
+  ([#201](https://github.com/voxpupuli/openvox-ca/issues/201)); see known gaps.
 - `MigrateService` holds two `bootstrap` locks (source backend, then
   destination). Pointing both at the same distributed backend would deadlock;
   migrating a store onto itself is unsupported.
@@ -180,7 +223,9 @@ path — still provides no cross-replica guarantee anyway.
 3. **Keep expensive, shared-state-free work outside the lock.** Key
    generation and CSR assembly in `Generate` run before any lock is taken;
    parsing and validation in `Renew`/`SaveRequest` likewise. Only the
-   storage-touching tail belongs inside.
+   storage-touching tail belongs inside. The deliberate exception is the CA
+   signature itself: `x509.CreateCertificate` runs under `c.mu` (see Tier 3),
+   because the cache update it guards must be atomic with the issuance.
 4. **Respect the ordering** (`subject` → `crl` → `c.mu`), never acquire the
    same lock reentrantly, and release `c.mu` before entering another
    `WithLock`. Use the closure-with-defer shape from `Renew`/`AutoRenew` so a
@@ -203,23 +248,41 @@ path — still provides no cross-replica guarantee anyway.
    runs under the process-local `inventoryMu` only, and etcd's CAS-guarded
    append protects the blob against lost updates, not serial uniqueness (the
    doc comment on `ErrDuplicateSerial` in
-   [storage.go](../../internal/storage/storage.go) spells this out).
+   [storage.go](../../internal/storage/storage.go) spells this out, and the
+   blob-backend gap is tracked as
+   [#204](https://github.com/voxpupuli/openvox-ca/issues/204)).
 7. **New lock names are protocol.** Add them to the constants in
-   [init.go](../../internal/ca/init.go), keep them stable across releases, and
-   document them in the table above. All callers using a name contend on one
-   lock, so never derive a name from unvalidated input (subject names pass
-   `ValidateSubject` first).
+   [init.go](../../internal/ca/init.go) (and keep `migrateLockName` in
+   [migrate.go](../../internal/storage/migrate.go) in sync — it redefines
+   `"bootstrap"` independently), keep them stable across releases, and document
+   them in the table above. All callers using a name contend on one lock, so
+   never derive a name from unvalidated input (subject names pass
+   `ValidateSubject` first). `ValidateSubject` is necessary but not sufficient
+   on the SQL backends: there the lock identity is a 64-bit FNV-1a hash of the
+   name (`advisoryLockKey`), so distinct names can alias and a crafted subject
+   could collide with `crl`/`bootstrap`
+   ([#203](https://github.com/voxpupuli/openvox-ca/issues/203)).
 8. **SQL pool sizing:** on PostgreSQL/MySQL every *held* distributed lock pins
-   one pooled connection. Paths that nest `subject:<name>` → `crl` need two
-   simultaneously, plus connections for the queries inside; a `MaxOpenConns`
-   set too low turns that into acquisition timeouts under load.
+   one pooled connection. A single in-flight `Clean`/`Renew`/`AutoRenew` needs
+   at least three connections at once — one for the `subject:<name>` lock, a
+   second for the nested `crl` lock, and a third for the reads/writes inside
+   the revocation step — so `sql_max_open_conns` must be at least 3 per
+   concurrently mutating request. Set below that and a single request
+   hard-stalls (not only under load), bounded only by the 60 s `lockTimeout`.
+   See the `sql_max_open_conns` knob in
+   [storage backends](../storage-backends.md).
 9. **Offline `openvox-ca-ctl` commands** (import, migrate) assume the server
    is stopped. `MigrateService` holds `bootstrap` on both stores, which
    excludes a booting server but deliberately not per-subject signing — and on
    filesystem/SQLite the fallback mutex has no cross-process effect at all
-   ([#187](https://github.com/voxpupuli/openvox-ca/issues/187)).
+   ([#187](https://github.com/voxpupuli/openvox-ca/issues/187)). It also
+   inherits the caller's context with no `lockTimeout`, so it waits
+   indefinitely on a contended `bootstrap` lock (see Tier 1).
 
-## Known gaps (mostly tracked — check before re-reporting)
+## Known gaps
+
+Concurrency limitations that are understood and tracked. This list reflects the
+state when the document was last updated and is not guaranteed exhaustive.
 
 - [#195](https://github.com/voxpupuli/openvox-ca/issues/195) — `CA.Generate`
   (the `POST /generate/{subject}` endpoint) is the one issuance path that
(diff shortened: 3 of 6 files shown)


trevor-vaughan-ai:
> **Obsolete.** Superseded by the [current Review Council verdict](https://github.com/voxpupuli/openvox-ca/pull/198#issuecomment-5255250138) for commit `3495cc8`. 

## 🟢 Review Council: APPROVE

>
> Models used:
> - claude-opus-5
> - claude-opus-5[1m]

_Reviewed at commit [`87fd343`](https://github.com/voxpupuli/openvox-ca/commit/87fd343475475f01cf7a8f153d488a2f71153117) (standard effort)._

**TL;DR:** Approved by all five reviewers: the locking docs are accurate but incomplete, with non-blocking gaps and nothing that blocks merge.

**Findings:** 🔴 0 Critical, 🟠 0 High, 🟡 10 Medium, 🔵 6 Low

|---|---|---|
| 🛡️ Adversary (code) | ✅ Approve | 2 MEDIUM, 1 LOW |
| 📚 Curator (code) | ✅ Approve | 1 MEDIUM, 1 LOW |
| 🧭 Guard (code) | ✅ Approve | 2 MEDIUM, 1 LOW |
| ⚙️ Operator (code) | ✅ Approve | 3 MEDIUM, 1 LOW |
| 🧪 Tester (code) | ✅ Approve | 2 MEDIUM, 2 LOW |

<details><summary>🟡 MEDIUM (10)</summary>

- 🛡️ **Doc calls CA.Init's lock inversion safe, but Init can re-enter the bootstrap lock and deadlock startup** ([`docs/development/locking.md:127`](https://github.com/voxpupuli/openvox-ca/blob/87fd343475475f01cf7a8f153d488a2f71153117/docs/development/locking.md#L127))

  ```
  - `CA.Init` inverts the order (it holds `c.mu` while acquiring `bootstrap`).
    That is safe only because `Init` runs to completion before the server starts
    serving, so nothing else can be holding a distributed lock while waiting on
    `c.mu`. Do not copy this pattern into anything that runs while serving.
  ```

  💡 **Recommendation:** Either add this to 'Known gaps' with an issue reference (Init -> finishLoadExisting -> seedSupportingState re-enters `bootstrap`, and the process-local gate ignores ctx so the hang is unbounded), or amend the `CA.Init` bullet to state the second condition its safety depends on. The underlying code fix is to give `seedSupportingState` a `...Locked` variant called from inside Init's existing bootstrap closure.
  **Constraint:** Documented locking guarantees must not overstate the safety actually implemented (reviewer-protocol evidence discipline; CWE-833 Deadlock)

  This bullet gives one reason Init's ordering inversion is safe and stops there, but Init has a second, unmentioned hazard: it can acquire the `bootstrap` lock twice on one goroutine. In internal/ca/init.go the slow path calls `c.Storage.WithLock(ctx, lockNameBootstrap, ...)` (line 98) and, inside that closure, `return c.finishLoadExisting(ctx)` (line 101). `finishLoadExisting` calls `c.seedSupportingState(ctx)` (line 155) whenever `loadCRLCache` fails with `fs.ErrNotExist`, and `seedSupportingState` itself calls `c.Storage.WithLock(ctx, lockNameBootstrap, ...)` (line 168) - a reentrant acquisition of the same name. That path is reachable when the fast-path `loadCA` fails but the in-lock `loadCA` succeeds while the CRL is absent, e.g. after a replica crashed between the cert write and the CRL write in `bootstrapCA`, or after a transient KeyProvider failure. Every implementation deadlocks there and none of them honour the 60 s ctx: `StorageService.WithLock` falls back to `s.localNamedLock(name)` (`sync.Mutex`, storage.go:88-90) for filesystem/SQLite, and `EtcdBackend.AcquireLock` (etcd.go:391-392), `RedisBackend.AcquireLock` (redis.go:478-479) and `SQLBackend.acquirePostgresLock`/`acquireMySQLLock` (sql.go:582-583, 647-648) each take a plain per-name `sync.Mutex` before any network call. Result: CA startup hangs forever rather than returning a lock-acquisition error. I grepped `grep -rn "seedSupportingState\|finishLoadExisting" --include=*.go` (only internal/ca/init.go lines 82, 101, 138, 155, 165) and `grep -n "seedSupportingState\|reentrant\|nest" docs/development/locking.md` - the doc names `CA.seedSupportingState` only as a `bootstrap` holder (line 39) and states the no-reentrancy rule as forward-looking guidance (line 185), never noting that Init already violates it. Because the doc is presented as the authority for this subsystem and its 'Known gaps' list omits this, a future contributor is steered away from finding a real startup deadlock.
  </details>

- 🛡️ **Rule 7 presents ValidateSubject as sufficient for lock-name derivation, omitting the SQL backends' 64-bit hash aliasing** ([`docs/development/locking.md:207`](https://github.com/voxpupuli/openvox-ca/blob/87fd343475475f01cf7a8f153d488a2f71153117/docs/development/locking.md#L207))

  ```
  7. **New lock names are protocol.** Add them to the constants in
     [init.go](../../internal/ca/init.go), keep them stable across releases, and
     document them in the table above. All callers using a name contend on one
     lock, so never derive a name from unvalidated input (subject names pass
     `ValidateSubject` first).
  ```

  💡 **Recommendation:** State in rule 7 (and/or the backend table) that on PostgreSQL/MySQL the lock name is reduced to a 64-bit FNV-1a key, so distinct names can alias and `ValidateSubject` does not prevent an attacker-chosen subject from colliding with `crl`/`bootstrap`. Note the mitigation for the code side: namespace the derived key (e.g. two-argument `pg_advisory_lock(classid, objid)` keyed on lock class) or hash with a keyed/wide digest so a subject lock can never land in the CRL/bootstrap key space.
  **Constraint:** Security-relevant caveats about lock-name derivation from input must be stated accurately (CWE-694: Use of Multiple Resources with Duplicate Identifier)

  The rule tells a contributor that validating input is what makes deriving a lock name from it safe. That is true for etcd and Redis, which use the literal name (`b.prefix+"/locks/"+name`, etcd.go:399; `b.prefix + ":locks:" + name`, redis.go:481), but not for the SQL backends: `advisoryLockKey` (sql.go:611-618) collapses the name to a 64-bit FNV-1a digest used directly as the `pg_advisory_lock` key (sql.go:591-592), and `mysqlLockName` (sql.go:687-689) formats that same digest as the `GET_LOCK` name. FNV-1a is trivially invertible, so a subject name that passes `ValidateSubject`'s `^[a-z0-9_.][a-z0-9._-]*$` charset can be constructed (meet-in-the-middle over a 64-bit space) to hash to the same key as `crl` or `bootstrap`. On PostgreSQL/MySQL that subject's lock then *is* the CRL lock: an unauthenticated CSR submission for that name serialises against every `Revoke`/`ReissueCRL`/`RefreshCRLIfDue` cluster-wide, and any path that nests `subject:<name>` -> `crl` (Clean/Renew/AutoRenew, signing.go:494, 759, 862) blocks against itself on a second pooled connection until the 60 s `lockTimeout` fires - a repeatable denial of revocation, not merely extra contention. I grepped `grep -n "hash\|FNV\|fnv\|advisory\|GET_LOCK\|collide\|collision\|alias" docs/development/locking.md`: the only hits are the backend-table rows at lines 49-50, which name `pg_advisory_lock` and `GET_LOCK` but never mention that the lock identity on those backends is a 64-bit digest of the name rather than the name itself.
  </details>

- 📚 **Tier 3 omits that c.mu is held across the CA signing round trip** ([`docs/development/locking.md:101`](https://github.com/voxpupuli/openvox-ca/blob/87fd343475475f01cf7a8f153d488a2f71153117/docs/development/locking.md#L101))

  ```
  Mutating operations hold `c.mu` (write) across the storage mutation *and* the
  cache update, so within a process the caches can never be observed out of step
  with what the same process just wrote. Read paths take `c.mu.RLock` only.
  ```

  💡 **Recommendation:** Add a sentence or two to the Tier 3 section stating that mutating paths hold c.mu across the CA signature itself (x509.CreateCertificate via issueLeafLocked), that with an external key provider this spans a network/IPC round trip, and that issuance therefore serialises process-wide at roughly one signing round trip at a time - cross-referencing docs/openbao-transit.md's "Performance and outage behaviour" section. Then reword the #197 bullet so it reads as an extension of that general property rather than an OCSP-only quirk, and note in rule 3 that the signing call is the deliberate exception to keeping expensive work outside the lock.
  **Constraint:** AGENTS.md:143 designates this document authoritative for any change that reads or mutates shared CA state; a material concurrency property present in the source must therefore appear here.

  The Tier 3 section scopes c.mu's write side to "the storage mutation and the cache update", but every issuance path holds c.mu across x509.CreateCertificate itself: Sign (signing.go:169), SignWithTTL (192), SaveRequest (647), Renew (739), AutoRenew (835), ImportCertificate (importcert.go:142) and Generate (generate.go:109) all lock c.mu inside their closure and call issueLeafLocked, whose godoc at signing.go:317 states "c.mu must be held by the caller". Under ca_key_provider: openbao or the isolated signer, the CA-key signature inside that call is a network or IPC round trip, so c.mu - not the per-subject cluster lock - is the process-wide issuance serialiser. The repository documents this twice elsewhere: internal/signer/openbao/signer.go ("the CA holds its process-wide mutex across x509.CreateCertificate - and therefore across this network call - so an unbounded Sign against a stalled Transit backend would pin that mutex and stall all issuance") and docs/openbao-transit.md:268 ("Issuance holds the CA's process-wide lock across the signing call"). Searching this document with grep -n -i "signer|openbao|transit|CreateCertificate|process-wide|network call|round trip|throughput|issuance" docs/development/locking.md returns only lines 225 and 234, both inside the Known gaps list; line 234 attributes the pattern to OCSP's slow path alone, which reads as an OCSP-specific anomaly rather than the shared property of every issuance path. Rule 3 at line 180 ("Keep expensive, shared-state-free work outside the lock") compounds the omission, since the most expensive operation in the system runs inside the lock with no explanation.
  </details>

- 🧭 **Doc names init.go as the only definition site, but `bootstrap` is also defined in migrate.go** ([`docs/development/locking.md:33`](https://github.com/voxpupuli/openvox-ca/blob/87fd343475475f01cf7a8f153d488a2f71153117/docs/development/locking.md#L33))

  ```
  The lock names are part of the cross-replica protocol: every replica must agree
  on them, so they are **stable across releases**. They are defined in
  [init.go](../../internal/ca/init.go):
  ```

  💡 **Recommendation:** Add the second definition site to the sentence and to rule 7, e.g. "defined in init.go, and mirrored as `migrateLockName` in internal/storage/migrate.go - rename both together".
  **Constraint:** Cross-Component

  The document, and rule 7 at line 207 ("Add them to the constants in [init.go], keep them stable across releases"), present internal/ca/init.go as the single definition site for lock names. `grep -rn '\"bootstrap\"' --include=*.go .` returns two independent definitions: internal/ca/init.go:41 `lockNameBootstrap = "bootstrap"` and internal/storage/migrate.go:87 `const migrateLockName = "bootstrap"`. internal/storage cannot import internal/ca (the dependency runs the other way), so the two constants are coupled only by the string literal. A contributor following rule 7 would rename the init.go constant and silently break the mutual exclusion between a migration and a bootstrapping server that the doc's own `bootstrap` row advertises. The doc already flags the reuse as deliberate, so it is aware of the coupling but not of the second definition site.
  </details>

- 🧭 **Per-backend lock mechanism is now documented in two places with no stated owner** ([`docs/development/storage-internals.md:84`](https://github.com/voxpupuli/openvox-ca/blob/87fd343475475f01cf7a8f153d488a2f71153117/docs/development/storage-internals.md#L84))

  ```
  grabs per-name mutexes under `<prefix>/locks/<name>`. The lock names, every
  operation that holds each one, and the ordering invariant are documented in
  [locking and concurrency](locking.md) — that table is authoritative, so it is
  not duplicated here.
  ```

  💡 **Recommendation:** Declare one owner for mechanism as the PR already did for names: either have storage-internals.md defer to locking.md's backend table, or have locking.md cite storage-internals.md for mechanism and keep only the crash-recovery column.

  The PR de-duplicates lock names, holders and ordering into locking.md and says so explicitly here. It simultaneously introduces a new duplication for the other half: locking.md's "How each backend provides the distributed lock" table (lines 45-52) restates facts this document still carries in its three Cross-node coordination sections - etcd concurrency.Mutex under `<prefix>/locks/<name>` on a 30 s lease (locking.md:47 vs storage-internals.md:82-90); Redis SET NX PX with a random token, heartbeat TTL extension and Lua compare-and-delete (locking.md:48 vs storage-internals.md:121-127); pg_advisory_lock on a dedicated connection (locking.md:49 vs storage-internals.md:194-199); GET_LOCK with a 1 s polled wait (locking.md:50 vs storage-internals.md:200-206); SQLite/filesystem falling back to a process-local mutex (locking.md:51 vs storage-internals.md:91-92 and 190-193). storage-internals.md:187-188 frames its SQL section as "only the per-dialect mechanism differs", implying mechanism is its responsibility, while locking.md supplies a competing cross-backend view. Two owners for the same facts is the drift this PR set out to remove.
  </details>

- ⚙️ **Crash-recovery table treats SQL session teardown as equivalent to a lease TTL** ([`docs/development/locking.md:49`](https://github.com/voxpupuli/openvox-ca/blob/87fd343475475f01cf7a8f153d488a2f71153117/docs/development/locking.md#L49))

  ```
  | PostgreSQL | `pg_advisory_lock` (session-level) on a dedicated pooled connection | Session ends, lock releases |
  ```

  💡 **Recommendation:** Split the recovery-time bound out of the mechanism: state that etcd and Redis self-heal within the 30 s TTL, while PostgreSQL/MySQL locks release only when the database detects the dead session (tcp_keepalives_idle / wait_timeout, hours by default). Point operators at pg_terminate_backend / KILL and at tuning those server settings for HA deployments.
  **Constraint:** Operator-facing crash-recovery behaviour must state the recovery time bound, not just the mechanism

  The table puts PostgreSQL and MySQL alongside etcd's 30 s lease and Redis's TTL under one 'Crash recovery' column, and the surrounding prose justifies the 60 s lockTimeout as 'short enough that a crashed replica's stale lease doesn't hang requests forever' (line 62-63). Neither SQL lock has a TTL: pg_advisory_lock and GET_LOCK persist until the database server tears the session down. I grepped internal/storage/sql.go and internal/storage/spec.go for ConnMaxLifetime, ConnMaxIdleTime, KeepAlive and keepalive and found none - the backend sets only SetMaxOpenConns/SetMaxIdleConns (sql.go:159-166). After a hard host loss or partition, session reaping is governed by PostgreSQL tcp_keepalives_idle and MySQL wait_timeout, whose defaults are hours. A crashed replica can therefore hold `crl` or `subject:<name>` far beyond 60 s while every surviving replica's revoke/sign fails at lockTimeout, and no doc names the recovery action (terminate the orphaned backend session, or lower the server-side keepalive).
  </details>

- ⚙️ **hmac_key read-modify-write is covered by no tier and is missing from the known gaps** ([`docs/development/locking.md:83`](https://github.com/voxpupuli/openvox-ca/blob/87fd343475475f01cf7a8f153d488a2f71153117/docs/development/locking.md#L83))

  ```
  | `fileMu` | `ca_cert`, `ca_pubkey`, `ca_key`, `csr/<subject>`, `cert/<subject>`, per-subject private keys | One mutex spans all subjects; simple and sufficient at current scale |
  | `serialMu` | `serial` | Plain read/write pairs |
  ```

  💡 **Recommendation:** Add hmac_key to the tier-2 table (noting it is currently guarded by neither a mutex nor a cluster lock) and add a 'Known gaps' bullet for the unlocked EnsureHMACKey read-modify-write, with the concurrent-cold-start consequence spelled out so a contributor can file or fix it rather than assume it is covered.
  **Constraint:** A document AGENTS.md makes mandatory for all shared-state changes must enumerate every shared-state read-modify-write or list it as a known gap

  The tier-2 section claims StorageService 'guards each family of logical keys with its own process-local mutex' and then lists four mutexes. The logical key hmac_key (KeyHMACKey, internal/storage/backend.go:63) is absent from every row, and it is a genuine read-modify-write: EnsureHMACKey does backend.Get(ctx, KeyHMACKey) -> generate -> backend.Put(ctx, KeyHMACKey, ...) at internal/storage/storage.go:721-743 with no mutex taken, and its only caller InitHMAC (storage.go:164-171) runs at internal/ca/init.go:74, before the bootstrap WithLock acquired at init.go:96-98. I ran `grep -n "hmac_key\|HMACKey\|EnsureHMACKey\|InitHMAC" docs/development/locking.md` and it returns no matches, so the page neither covers it in the tier-2 table nor lists it under 'Known gaps' next to #195/#196/#187. Failure scenario: two replicas cold-starting against a fresh shared etcd/Redis/SQL backend both find hmac_key absent, both generate, and last write wins; the loser keeps its own key in s.hmacKey and writes inventory_hmac under it, after which a replica loading the stored key fails VerifyInventoryHMAC (storage.go:925-935) and CA.Init aborts with 'inventory integrity check failed' (init.go:74-76).
  </details>

- ⚙️ **SQL pool-sizing rule gives no minimum and understates the stall as load-dependent** ([`docs/development/locking.md:214`](https://github.com/voxpupuli/openvox-ca/blob/87fd343475475f01cf7a8f153d488a2f71153117/docs/development/locking.md#L214))

  ```
     simultaneously, plus connections for the queries inside; a `MaxOpenConns`
     set too low turns that into acquisition timeouts under load.
  ```

  💡 **Recommendation:** State the floor (at least 3 connections per concurrently mutating request, since subject -> crl nesting plus inner queries needs three) and drop the 'under load' qualifier, since a single request stalls. Cross-reference the rule from the sql_max_open_conns entry in docs/storage-backends.md so operators tuning the knob see the constraint.
  **Constraint:** Operational sizing guidance must be quantified when the corresponding knob is operator-configurable

  Rule 8 correctly identifies that each held distributed lock pins a pooled connection but never states the floor. A single in-flight Clean/Renew/AutoRenew needs three connections concurrently: WithLock(subject:<name>) pins one via b.db.Conn(ctx) (internal/storage/sql.go:585 for Postgres, :650 for MySQL), the nested WithLock("crl") at internal/ca/signing.go:494/759/862 pins a second, and the storage reads/writes inside the revocation step draw a third from the same pool. With sql_max_open_conns set to 2 that is a hard stall for one request with no concurrency at all, bounded only by the 60 s lockTimeout, so the 'under load' qualifier misdescribes the failure. The knob is operator-facing and unqualified: docs/storage-backends.md:310 shows `sql_max_open_conns: 0  # 0 = database/sql default` with no minimum, and `grep -rn "max_open_conns\|pool siz" docs/` finds only that line, the env-var row at docs/storage-backends.md:374, and this new rule.
  </details>

- 🧪 **Nested lock ordering invariant has no concurrent regression test** ([`docs/development/locking.md:117`](https://github.com/voxpupuli/openvox-ca/blob/87fd343475475f01cf7a8f153d488a2f71153117/docs/development/locking.md#L117))

  ```
  - `Clean`, `Renew`, and `AutoRenew` are the paths that take all three: the
    subject lock around the whole operation, then the `crl` lock + `c.mu` for the
    revocation step.
  ```

  💡 **Recommendation:** Add a concurrency test in internal/ca that drives Clean/Renew (or Revoke) and a competing signing path against the same subject from multiple goroutines with a bounded context, asserting completion rather than timeout. Model it on the existing autosign race at internal/ca/ca_test.go:1597 and reference it from the Tests section so the ordering invariant has a named anchor.
  **Constraint:** Tester: regression anchors - invariants a contributor is told to preserve need automated protection

  This section and rule 4 ("Respect the ordering (`subject` -> `crl` -> `c.mu`)") present the ordering as an invariant future changes must preserve, and the Known gaps section warns that PR #186 will move `Revoke` under `subject:<name>` -> `crl`. No test guards it. `grep -rn "go func()" --include=*_test.go internal/ca internal/api` returns exactly one hit, internal/ca/ca_test.go:1608, inside the "allows exactly one autosign when concurrent requests race for the same subject" case, which exercises only the subject lock via SaveRequest. `grep -rn "deadlock\|lock order\|ordering" --include=*_test.go internal/` returns nothing. Clean, Renew, AutoRenew and Revoke are exercised only sequentially, and a sequential test cannot deadlock on an inverted acquisition order. Failure scenario: a contributor implementing the PR #186 reordering acquires `crl` before `subject:<name>` on one path while Clean keeps the documented order; both suites stay green because no test drives the two paths concurrently, and under production load two replicas (or two in-process goroutines) deadlock until lockTimeout expires, turning revocation and cleanup into 60 s timeouts.
  </details>

- 🧪 **Unit suite runs concurrency tests without the race detector** ([`magefile.go:303`](https://github.com/voxpupuli/openvox-ca/blob/87fd343475475f01cf7a8f153d488a2f71153117/magefile.go#L303))

  ```
  	testArgs := append([]string{"test", "-json", "-cover", "-coverprofile=coverage.out"}, pkgs...)
  ```

  💡 **Recommendation:** Add `-race` to Test.Unit's testArgs (and ideally to the Test.Backends* invocations), or add a separate race-enabled target covering internal/storage and internal/ca, then mention it in docs/development/testing.md and this document's Tests section.
  **Constraint:** lang-go testing_conventions: race detection (`go test -race`) - flag missing race detection in concurrent code tests

  `grep -rn -- "-race" magefile.go .github/workflows/` returns no results, and a repo-wide `grep -rn "race" --include=*.yml --include=*.go .` finds only prose ("graceful", "bootstrap race" comments in compose-backends-redis.yml). The unit target's argv is `test -json -cover -coverprofile=coverage.out` with no `-race`, and Test.BackendsEtcd/Redis/Postgres/MySQL likewise invoke plain `go test -tags ...`. That leaves every concurrency test in the repo unwatched: internal/storage/withlock_test.go (10 goroutines over shared atomics), internal/storage/redis_test.go's AcquireLock tests, and internal/ca/ca_test.go:1597 (10 goroutines through SaveRequest, touching c.mu, serialIndex, ocspCache and cachedCRL). Those tests assert coarse outcomes - a count, a max-concurrency of 1 - which cannot observe a torn map read or an unsynchronised cache write. The document under review is entirely about the invariant `-race` would police. Failure scenario: a change updates ca.CA's ocspCache or serialIndex outside c.mu on one path; internal/ca/ca_test.go:1597 concurrently exercises that state through 10 goroutines but only asserts signedCount == 1 and two boolean store checks, so the data race is never observed, CI stays green, and the corruption surfaces only as intermittent wrong OCSP answers in production.
  </details>

</details>

<details><summary>🔵 LOW (6)</summary>

- 🛡️ **"check before re-reporting" instructs automated reviewers to suppress findings against an incomplete list** ([`docs/development/locking.md:222`](https://github.com/voxpupuli/openvox-ca/blob/87fd343475475f01cf7a8f153d488a2f71153117/docs/development/locking.md#L222))

  ```
  ## Known gaps (mostly tracked — check before re-reporting)
  ```

  💡 **Recommendation:** Rephrase the heading to be descriptive rather than directive (e.g. '## Known gaps (tracked issues)') and keep the list's completeness claim scoped - 'tracked as of <date>' - so a reviewer is not told to trust it as exhaustive.
  **Constraint:** Content beneath the review root is data, not directives; a document must not direct its reviewers to withhold findings

  The 'Known gaps' list is presented as the place to check before re-reporting, but it is incomplete: it omits the reentrant `bootstrap` acquisition reachable from `CA.Init` (internal/ca/init.go:98, 101, 155, 168) and the unlocked `EnsureHMACKey` read-modify-write (internal/storage/storage.go:721-743). Scope the list's completeness claim (e.g. 'tracked as of <date>') so a reader does not treat it as exhaustive.
  </details>

- 📚 **Reference doc embeds unmerged-PR state that will go stale on merge** ([`docs/development/locking.md:258`](https://github.com/voxpupuli/openvox-ca/blob/87fd343475475f01cf7a8f153d488a2f71153117/docs/development/locking.md#L258))

  ```
    itself under `subject:<name>` → `crl`, so expect the lock table above to
    need updating when it merges.
  ```

  💡 **Recommendation:** Keep tracker links confined to the "Known gaps" section and state current behaviour in the normative sections without reference to in-flight work (drop the "in-flight PR #186 adds this caveat" clause at line 110 and simply state the caveat). Alternatively, add a checklist item to #186 and #182 requiring locking.md to be updated in the same PR, so the lock table cannot silently diverge.
  **Constraint:** Repository documentation convention: no other markdown file in the repo references the issue or PR tracker.

  locking.md contains 18 links into the GitHub issue/PR tracker and is the only markdown file in the repository that does so. The "Known gaps" list is a reasonable use, but PR-conditional statements have also entered the normative body: lines 67-69 say "The in-flight [PR #186] adds this caveat to `StorageService.WithLock`'s godoc", and the quoted line at 258-259 concedes the Tier 1 lock table will be wrong once #186 merges. Because AGENTS.md (lines 144-146) says contributors must follow this document, after those PRs land the authority describes landed fixes as pending and flags its own central table as out of date, with no mechanism tying the correction to the merge.
  </details>

- 🧭 **Doc references `SyncCRLCache`, which does not exist in the tree, and declares its own table pending update** ([`docs/development/locking.md:247`](https://github.com/voxpupuli/openvox-ca/blob/87fd343475475f01cf7a8f153d488a2f71153117/docs/development/locking.md#L247))

  ```
    [PR #182](https://github.com/voxpupuli/openvox-ca/pull/182) fixes it with a
    background poll (`SyncCRLCache`, monotonic in the CRL number, deliberately
    lock-free).
  ```

  💡 **Recommendation:** Keep the forward references to #182/#186 but drop the self-addressed "expect the table above to need updating" instruction - let the merging PR carry its own doc update - and either file or omit the split-out blob-backend issue.
  **Constraint:** Zero-Waste

  `grep -rn "SyncCRLCache" .` over the entire review root returns exactly one hit - this doc line; the symbol is not defined in any Go file. `git log --oneline --all | grep -iE "#18[26]|#173|#171"` returns nothing, so PRs #182 and #186 are genuinely unmerged and the "in-flight" framing is honest today. The cost is baked-in staleness in a document AGENTS.md now makes mandatory reading, most visibly at line 258 ("so expect the lock table above to need updating when it merges"), which is an in-document TODO against the table the doc calls authoritative. The blob-backend AppendInventory gap has the same shape at lines 242-243: described as "explicitly split out of #187 for an issue of its own" with no issue number, so it has no tracked handle.
  </details>

- ⚙️ **lockTimeout is described as universal but the migrate path passes an untimed context** ([`docs/development/locking.md:60`](https://github.com/voxpupuli/openvox-ca/blob/87fd343475475f01cf7a8f153d488a2f71153117/docs/development/locking.md#L60))

  ```
  Callers bound lock acquisition *and* the critical section together with
  `lockTimeout` (60 s, [init.go](../../internal/ca/init.go)) via
  ```

  💡 **Recommendation:** Qualify the sentence to 'CA request paths bound ...' and note in rule 9 that MigrateService inherits the caller's context, so `openvox-ca-ctl migrate` waits indefinitely on a contended bootstrap lock; suggest interrupting it and stopping the server rather than waiting.
  **Constraint:** Blanket statements about request bounding must hold for every documented caller, including offline CLI paths

  The claim holds for every CA-side call site (I confirmed context.WithTimeout(ctx, lockTimeout) at internal/ca/init.go:96,166; crl.go:97,141; cleanup.go:52; revoke.go:40; importcert.go:137; signing.go:165,188,472,643,718,809). It does not hold for the migrate path the same page documents in rule 9 and in the `bootstrap` lock row: storage.MigrateService (internal/storage/migrate.go:99-108) adds no timeout, and its only production caller passes cobra's cmd.Context() straight through (cmd/openvox-ca-ctl/migrate.go:124), which is signal-cancellable but has no deadline. On etcd/Redis/PostgreSQL/MySQL, `openvox-ca-ctl migrate` therefore blocks indefinitely when a replica holds `bootstrap`, with no timeout and no progress output (MigrateOptions.Logf only prints under --verbose). Rule 9 at lines 216-220 covers the migrate lock's scope but not this difference.
  </details>

- 🧪 **Tests section overstates coverage: the Overlay lock path is untested** ([`docs/development/locking.md:267`](https://github.com/voxpupuli/openvox-ca/blob/87fd343475475f01cf7a8f153d488a2f71153117/docs/development/locking.md#L267))

  ```
  `WithLock`'s fallback semantics are covered in
  [withlock_test.go](../../internal/storage/withlock_test.go); each distributed
  implementation's mutual exclusion is exercised in its backend integration
  suite (build-tagged; see [testing](testing.md)).
  ```

  💡 **Recommendation:** Add an overlay lock test in internal/storage/overlay_test.go: assert NewOverlayBackend over a filesystem base returns ErrDistributedLockingUnsupported from AcquireLock, and that an overlay over a Locker base (the existing stubLocker works) forwards the call and returns the base's Unlocker. Either that, or narrow the Tests section to say the Overlay row is covered only indirectly through WithLock's stub and fix the misleading comment at withlock_test.go:115-118.
  **Constraint:** Tester: coverage completeness - documented coverage claim must match the tests that exist

  `OverlayBackend.AcquireLock` (internal/storage/overlay.go:213-217) has no test: overlay_test.go covers only Put/Get/Delete, AppendLine, Path and constructor validation, and the only other test constructing an overlay (internal/storage/sql_inventory_test.go:233) exercises InitHMAC/ReadInventory/LatestSerialForSubject, never WithLock, so both branches are uncovered. The comment at withlock_test.go:115-118 claims to cover "the OverlayBackend-over-filesystem case" while the test at line 122 actually drives a stubLocker, so the real type is never instantiated. (The Tests section's own claim is accurate: Overlay is a delegating wrapper, not a distributed implementation, and all four distributed implementations do have the build-tagged mutual-exclusion suites.) Add an overlay lock test and fix the comment.
  </details>

- 🧪 **stubUnlocker error field is dead scaffolding for an untested unlock-failure path** ([`internal/storage/withlock_test.go:199`](https://github.com/voxpupuli/openvox-ca/blob/87fd343475475f01cf7a8f153d488a2f71153117/internal/storage/withlock_test.go#L199))

  ```
  type stubLocker struct {
  	Backend
  	err        error
  	unlockErr  error
  	unlockedCh chan struct{}
  }
  ```

  💡 **Recommendation:** Add one case setting stubLocker.unlockErr and assert WithLock still returns fn's result (nil, or the sentinel fn returned) rather than the unlock error - or delete the unlockErr field if that path is deliberately out of scope.
  **Constraint:** Tester: no dead scaffolding; error paths that are deliberately swallowed need a test

  `grep -rn "unlockErr" internal/` returns only the declaration on line 202 and the pass-through on line 210; no test ever sets it, so stubUnlocker always returns a nil error from Unlock. Meanwhile StorageService.WithLock (internal/storage/storage.go:75-77) deliberately swallows a failed release into slog.Warn and still returns fn()'s result - a contractual decision (release failure must not mask fn's outcome) with no test. The scaffolding needed to write that test already exists and is unused. Failure scenario: a refactor makes WithLock return the Unlock error instead of warning, so a transient etcd/Redis release failure turns a successful CRL re-sign into a reported error and the caller retries a mutation that already committed - no test fails, because the unlock-error path is never driven.
  </details>

</details>

---

bootc:
Thanks for the thorough council pass. I've worked through all 16 findings; disposition below. Pushed as 3495cc8.

**Documentation corrections applied (locking.md unless noted):**

- **Crash-recovery vs TTL** — the backend table now separates etcd/Redis (self-heal within the 30s TTL) from the SQL advisory locks, which have no TTL and release only when the DB reaps the session (`tcp_keepalives_idle` / `wait_timeout`, hours by default); the recovery action (`pg_terminate_backend` / `KILL`, or lowering server keepalives) is spelled out.
- **c.mu across the signing round trip** — Tier 3 now states that `c.mu` is held across `x509.CreateCertificate`, so under an external signer issuance serialises process-wide across the round trip; rule 3 names this as the deliberate exception, and the #197 gap is reworded as the same property surfacing on OCSP. Cross-linked to the OpenBao guide's outage section.
- **`bootstrap` second definition site** — the "defined in init.go" sentence and rule 7 now record `migrateLockName` in migrate.go and that the two must be renamed together.
- **SQL hash aliasing** — rule 7 now notes `ValidateSubject` is necessary but not sufficient on the SQL backends, where the lock identity is a 64-bit FNV hash of the name.
- **Pool-sizing floor** — rule 8 now gives the ≥3-connections-per-mutating-request floor, drops the "under load" qualifier (a single request stalls), and cross-refs `sql_max_open_conns`.
- **lockTimeout scope** — qualified to CA request paths; notes the migrate path inherits an untimed context and waits indefinitely.
- **Known-gaps heading** — changed from the directive "check before re-reporting" to a descriptive heading with a scoped, non-exhaustive completeness note.
- **Unmerged-PR state** — dropped the "PR #186 adds this to the godoc" clause and the self-addressed "expect the table to need updating" TODO from the normative body; the not-yet-existing `SyncCRLCache` symbol name is gone. Forward references now live only in Known gaps.
- **storage-internals.md** — declared the owner of per-backend lock *mechanism*, with locking.md's backend table framed as a summary, resolving the two-owners duplication.

**Code-level findings filed as issues and referenced from Known gaps:**

- #201 — `CA.Init` slow path re-enters `bootstrap` (`finishLoadExisting` → `seedSupportingState`) and deadlocks startup. Confirmed reachable; the doc no longer calls Init's inversion unconditionally safe.
- #202 — `hmac_key` init is an unlocked read-modify-write running before the bootstrap lock; can fork on concurrent cold start. Added to Known gaps and noted as absent from the Tier 2 table for that reason.
- #203 — SQL advisory-lock name aliasing (the code side of the rule 7 caveat above).
- #204 — the blob-backend `AppendInventory` gap that was "split out of #187" now has a real number.
- #205 — `-race` on the unit target plus a concurrent lock-ordering regression test. Kept out of this PR deliberately: `-race` can surface pre-existing races and belongs in its own PR where CI can be watched, and the ordering test wants care to avoid flakiness.

**Test findings fixed in this PR:**

- Corrected the misleading `withlock_test.go` comment (it claimed OverlayBackend coverage while driving a `stubLocker`).
- Added a real `OverlayBackend.AcquireLock` test (unsupported-base, forwarding, and error-propagation).
- Added a test that `WithLock` swallows an `Unlock` error without masking fn's result (closing the dead `unlockErr` scaffolding).

The two larger test items (#205) are the only findings not addressed in-branch, by design.

trevor-vaughan-ai:
## 🟢 Review Council: APPROVE

>
> Models used:
> - claude-sonnet-5

_Reviewed at commit [`3495cc8`](https://github.com/voxpupuli/openvox-ca/commit/3495cc87389f18a59bf9199efdb257575332c970) (standard effort)._

**TL;DR:** All five reviewer personas approved this PR with zero new findings; prior issues were already resolved.

**Findings:** 🔴 0 Critical, 🟠 0 High, 🟡 0 Medium, 🔵 0 Low

|---|---|---|

---

